PR #2, opened 2015-12-03
Fix gcm.conf.sample file in plist

jlduran:
@rbgarga: I think it's better to submit this downstream. There's too much red tape upstream!


rbgarga:
Good catch, thanks. I've also fixed it on FreeBSD ports tree.

jlduran:
:+1: Thanks!

I'm also trying to include the `chapoly` plugin, I want to test with a bunch of "always on" Android devices. But it's going to need a little tweaking in strogSwan (`le32toh`).

PR #7, opened 2015-12-05
avahi - remove PBI crap, add missing includes, fix services handling

doktornotor:


doktornotor:
@jim-p Please, check it now before I touch the rest of the PRs. Nuked lib_depends as well, so there' s just build-options left.

jim-p:
I believe lib_depends and port_category were only used by the sync script so now that it is out of use they can go, but @rbgarga will have to confirm.

jim-p:
OK, after talking to @rbgarga and looking over the code a bit, lib_depends and port_category can go now -- it's safe to remove them and nothing referenced them outside of the script which won't be run any longer.
It's also going to be safe to remove config_file once the commit I just made reaches snapshots.

doktornotor:
@jim-p - OK, all info.xml cruft gone from the ports PRs I created.

jim-p:
Looks good, these should all be merged now, thanks!

PR #24, opened 2015-12-10
pfBlockerNG Bootstrap conversion

BBcan177:
There are still some issues with pkg.php and pkg_edit.php that need to be resolved. It will not affect functionality, as they are mostly cosmetic issues.

Redmine Issue:
https://redmine.pfsense.org/issues/5618


cbuechler:
merged, thanks!

PR #34, opened 2015-12-22
First version to add AD/Kerberos authentication to squid to pfSense 2.3

gitdevmod:
This is my first PR to pfSense FreeBSD-ports

kerberos (kinit) from FreeBSD base
pkg needed:
	- openldap-sasl-client (replace openldap-client)
	- cyrus-sasl
	- cyrus-sasl-gssapi
	- msktutil

TODO:
	- cron auto-update computer is missing
	- test and test

### DIFF
--- a/www/pfSense-pkg-squid/files/usr/local/pkg/squid.inc
+++ b/www/pfSense-pkg-squid/files/usr/local/pkg/squid.inc
@@ -1948,7 +1948,7 @@ function squid_resync_msnt() {
 }
 
 /* adk remove keytab/krb5secret if exists */
-squid_adk_remove_files() {
+function squid_adk_remove_files() {
 	unlink_if_exists(SQUID_CONFBASE . "/PROXY.keytab");
 	unlink_if_exists("/etc/krb5secret");
 }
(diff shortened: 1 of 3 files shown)


anahimself:
I am really interested to test this because I was trying on pfsense 2.2 and no success (missing file libgssapi_spnego.so.10 while msktutil, ...)

Can i patch the 3 files used on the stable 2.3 and debug with you if any trouble (for the others)?
(are you sure the msktutil uses /etc/krb5.conf on 2.3, because it apparently doesnt on 2.3 => creating a fake krb file for no reason)

anahimself:
Problem:
if (($auth_method != 'cp') or ($auth_method != 'adk')) {

This statement is always true, you should replace the method ;-)

Atm all is working but i miss a group member check too, will add this myself in the squid.conf

anahimself:
Another comment, concerning the upn:
--upn HTTP/' . $hostname  .
should be changed to
--upn HTTP/' . strtoupper($config['system']['hostname'])

all the rest seems very good so far

gitdevmod:
Hi @anahimself, thanks for your comments, I've not tried this patch since my last commit, not sure it will apply cleanly

anahimself:
Good job anyway for all that

gitdevmod:
I still think it's a nice feature to have in pfsense, maybe you can fork and update the patch :)

pfSense-git-updates:

gitdevmod:
Hi, I don't think it can apply correctly, if anyone is interested you can reuse it.

anahimself:
Got it working in production, very useful

rbgarga:
CLA was not signed, closing.

leeramsay:
This would be a crazy useful feature, is there no way this can be advanced without the original author signing CLA? It looks like it worked out of the box, and a few people have tested it.

anahimself:
I agree.
It is an awesome feature, rarely implemented but a lot more efficient than NTLM.
I can easily give the modifications I made on the 3 files if any (wrong auth_method test + upn fix).
It is in production since 5 months ;-)

doktornotor:
If someone wants  to do a new PR to add this, of course feel free. Before doing that, however, you'd better discuss the pitfalls here. Such as, depending on packages that conflict with base pfSense dependencies is a no go from the very beginning. Best to discuss first in the forums before implementing something.

PR #47, opened 2016-01-18
Logic simplifications

stilez:
1) Nested IF statements can be combined
2) Add/remove shellcmd contains logic that only needs to run if the early shellcmd array does/doesn't exist
3) Logs should be non-vague ("...added/removed a shellcmd" is better if it states the command added/removed, and for removal, clarify it removed an existing command)

Resubmit, as originally against old pfsense-packages, see https://github.com/pfsense/pfsense-packages/pull/1229

### DIFF
--- a/sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
+++ b/sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
@@ -214,15 +214,15 @@ function patch_remove_shellcmd() {
 	global $config;
 	$a_earlyshellcmd = &$config['system']['earlyshellcmd'];
 	if (is_array($a_earlyshellcmd)) {
-		$removed = false;
+		$removed = 0;
 		foreach ($a_earlyshellcmd as $idx => $cmd) {
 			if (stristr($cmd, "apply_patches.php")) {
 				unset($a_earlyshellcmd[$idx]);
-				$removed = true;
+				$removed++;
 			}
 		}
-		if ($removed) {
-			write_config("System Patches package removed existing early shellcmd: apply patches");
+		if ($removed > 0) {
+			write_config("System Patches package removed {$removed} existing early shellcmd(s): apply patches");
 		}
 	}
 }
--- a/sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
+++ b/sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
@@ -194,6 +194,10 @@ function bootup_apply_patches() {
 
 function patch_add_shellcmd() {
 	global $config;
+
+	if (!isset($config['system']['earlyshellcmd'])) {
+	    $config['system']['earlyshellcmd'] = array();
+	}
 	$a_earlyshellcmd = &$config['system']['earlyshellcmd'];
 	$found = false;
 	if (is_array($a_earlyshellcmd)) {
@@ -212,6 +216,9 @@ function patch_add_shellcmd() {
 
 function patch_remove_shellcmd() {
 	global $config;
+	if (!isset($config['system']['earlyshellcmd'])) {
+	    $config['system']['earlyshellcmd'] = array();
+	}
 	$a_earlyshellcmd = &$config['system']['earlyshellcmd'];
 	if (is_array($a_earlyshellcmd)) {
 		$removed = 0;
--- a/sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
+++ b/sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
@@ -183,16 +183,11 @@ function bootup_apply_patches() {
 	$a_patches = &$config['installedpackages']['patches']['item'];
 
 	foreach ($a_patches as $patch) {
-		/* Skip the patch if it should not be automatically applied. */
-		if (!isset($patch['autoapply'])) {
-			continue;
-		}
-		/* If the patch can be reverted it is already applied, so skip it. */
-		if (!patch_test_revert($patch)) {
-			/* Only attempt to apply if it can be applied. */
-			if (patch_test_apply($patch)) {
-				patch_apply($patch);
-			}
+		/* Skip if it should not be automatically applied;
+		   only attempt to apply if it can be applied;
+		   and	if it can be reverted it is presumably already applied, so skip it. */
+		if (isset($patch['autoapply']) && patch_test_apply($patch) && !patch_test_revert($patch)) {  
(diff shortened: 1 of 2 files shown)


jim-p:
Looks good, thanks! One last thing:
Needs the version bumped so it will get rebuilt and show as an updated version: Update Makefile, files/usr/local/share/pfSense-pkg-System_Patches/info.xml and files/usr/local/pkg/systempatches.xml

stilez:
Every single PR on the packages, however small, will need those 3 files updated? 

Please excuse any ignorance displayed by this question, but - can that be either automated,  or only updated when there's something substantial and new?  Otherwise it sounds like it could become a source of chaos for versioning, avoidable tripling of work for small PRs, or constant "new versions" for end users... :)

rbgarga:
It was like this in the past because we were migrating from pfSense-packages repo using a script to avoid maintain 2 different packages base. Now that we stopped sync'ing both repos it can be done using ports mechanisms. I'll work on a batch conversion to make it possible to set portversion on xml files based on main Makefile

rbgarga:
Regarding the second point, if version doesn't change, a new package will never be built. Builder is automated and track package versions.

If you change package and doesn't change version you risky to have different software installed on customers.

We can always use PORTREVISION for small changes and PORTVERSION changes for big ones

stilez:
Then shall I leave this for now?  (It's uncharted waters for me, being unclear what line/s needs changing in Makefile, and sounds like script will take care of the other 2).

As an aside on this specific PR it's small enough that I wonder if it can be left until the next "new version" someone does, and included as part of that.  I am also a bit mindful as an end-user of the time + annoyance that could be caused if users arrived every day to find new reinstalls needed of their packages because some minor change took place in the source. Maybe for minor changes that can wait and be "rolled up" with others, it would be worth doing so? After all, it doesn't create a new firmware version for every 'core code' PR, either, even if PRs on the core router code are often quite significant. I've seen code with "every small update is a new version" and it can get pretty annoying if not done in a way that doesn't intrude on the end user/s, but presumably this would?

rbgarga:
If it's a non-functional change it can be merged without need to change version, then when a functional change is made, users will get a new version.

jim-p:
It's fairly minor so that's OK, though I wouldn't want to get in the habit of letting things like this sit over time. If there is a problem it wouldn't be found until much later when it gets deployed. 
@stilez The package repo changes are all documented on the wiki and the forum: https://forum.pfsense.org/index.php?topic=103481.0

rbgarga:
Yeah, @jim-p has a good point. If the non-functional change breaks something, we will just notice that in the future, and it's not a good practice. I'm going to commit a cleanup of info.xml and other internal xml files to make life easier so only PORTVERSION and/or PORTREVISION variables will need to be changed on main Makefile

rbgarga:
Done. Now only PORTREVISION or PORTVERSION need to be changed

stilez:
Much of this is a bit over my head :) But the fundamentals seem easy enough on a "monkey see, monkey do" basis. 

So where does this leave things for this PR, what do I need to know for future PRs to this repo, which are the fields (in which files) if any that need manually changing and the requirements for how and when they should change and the schema they will follow

stilez:
(okay the >0 is strictly redundant but makes clear the outcome)

sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
```diff
@@ -200,36 +195,35 @@ function bootup_apply_patches() {
 function patch_add_shellcmd() {
 	global $config;
 	$a_earlyshellcmd = &$config['system']['earlyshellcmd'];
```
rbgarga: It must check if isset() or is_array() on $config['system']['earlyshellcmd'] before set the reference here. If it's not an array then it must be initialized as
```
if (!isset($config['system']['earlyshellcmd'])) {
    $config['system']['earlyshellcmd'] = array();
}
$a_earlyshellcmd = &$config['system']['earlyshellcmd'];
```
After that , the if(is_array()) from line 199 can go away

sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
```diff
@@ -200,36 +195,35 @@ function bootup_apply_patches() {
...
+		// Implicitly creates array if needed
 		$a_earlyshellcmd[] = "/usr/local/bin/php-cgi -f /usr/local/bin/apply_patches.php";
-		write_config("System Patches package added a shellcmd");
+		write_config("System Patches package added an early shellcmd: apply patches");
 	}
 }
 
 function patch_remove_shellcmd() {
 	global $config;
 	$a_earlyshellcmd = &$config['system']['earlyshellcmd'];
```
rbgarga: Same check and initialization suggested above

stilez:
Done - but please see my Q's a couple or so posts up, thankyou!

sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
```diff
@@ -183,53 +183,54 @@ function bootup_apply_patches() {
...
+
+	if (!isset($config['system']['earlyshellcmd'])) {
+	    $config['system']['earlyshellcmd'] = array();
 	}
+	$a_earlyshellcmd = &$config['system']['earlyshellcmd'];
 	$found = false;
-	foreach ($a_earlyshellcmd as $idx => $cmd) {
-		if (stristr($cmd, "apply_patches.php")) {
-			$found = true;
+	if (is_array($a_earlyshellcmd)) {
```
rbgarga: This check is not necessary anymore, you are always setting it above

sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
```diff
@@ -183,53 +183,54 @@ function bootup_apply_patches() {
...
+	if (!isset($config['system']['earlyshellcmd'])) {
+	    $config['system']['earlyshellcmd'] = array();
 	}
-	$removed = false;
-	foreach ($a_earlyshellcmd as $idx => $cmd) {
-		if (stristr($cmd, "apply_patches.php")) {
-			unset($a_earlyshellcmd[$idx]);
-			$removed = true;
+	$a_earlyshellcmd = &$config['system']['earlyshellcmd'];
+	if (is_array($a_earlyshellcmd)) {
```
rbgarga: This check is not necessary

rbgarga:
@stilez also, please bump PORTVERSION on port Makefile to make sure a new package is built

stilez:
Resubmitted as PR #199 to avoid conflicts

PR #54, opened 2016-01-25
fix vars translation and STDIN LOOP

marcelloc:


rbgarga:
Please bump PORTVERSION or PORTREVISION on port Makefile otherwise a new package will never be built

rbgarga:
nvm, I noted you submitted another PR doing it. I'll just simplify part of this code and them merge both

PR #74, opened 2016-03-08
Status / Monitoring - Pollyfill - Includes Method

NOYB:
Polyfill for older browsers that yet don't have the newer "includes()" method implemented.


jdillard:
I haven't checked this one yet, but since it could apply to all pages, it might be better to keep it in this file: https://github.com/pfsense/pfsense/blob/master/src/usr/local/www/jquery/pfSense.js

NOYB:
Okay thanks.  I'll move it over there.

jdillard:
Closing this PR since it moved to here: https://github.com/pfsense/pfsense/pull/2718

PR #77, opened 2016-03-10
Status Monitoring - Auto Updating

NOYB:
Auto update graph 4 times per the selected resolution period.
Auto update graph on settings change.

### DIFF
--- a/sysutils/pfSense-Status_Monitoring/files/usr/local/www/status_monitoring.php
+++ b/sysutils/pfSense-Status_Monitoring/files/usr/local/www/status_monitoring.php
@@ -378,17 +378,32 @@ include("head.inc");
 
 					<span class="help-block">Inverse</span>
 				</div>
+				<div class="col-sm-2">
+					<select class="form-control" id="auto-update" name="auto-update">
+						<option value="0" selected>Off</option>
+						<option value="15">15 Seconds</option>
+						<option value="60">1 Minute</option>
+						<option value="300">5 Minutes</option>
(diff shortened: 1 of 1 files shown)


jdillard:
I like the idea of the auto update, but I think it should be a setting that is off by default and has the options of updating every 1, 5, or 10 minutes. There is no point in updating less than once a minute since the RRD data is polled once a minute.

NOYB:
Agree with it being user enabled option.

Rational for updating 4 times per selected resolution is that the graph updating and RRD updating are not synchronized.  Thus the graph updating is kept from lagging the RRD updates by no more than 1/4th the selected resolution.

jdillard:
I wouldn't over engineer it, RRD isn't really meant for real time stats. I am tempted to change the name back to RRD Graphs from Monitoring as it might be a little misleading on functionality.

NOYB:
If a decision is made not to use “Monitoring” how about one of these?
Status / System Charts
Status / System Graphs
Status / System Health
With or without “System”.
I like Graphs or System Graphs best.  Never did like “RRD Graphs”.  RRD is techie term only known by those who know what it is.

jdillard:
I think just Graphs might be best. RRD does sounds technical. System is one of the categories of graphs, so System Graphs doesn't sound all encompassing, and Status > Status Graphs sounds repetitive.

Ever Knowing (but not understanding) Graphs :)

NOYB:
Replaced by PR #83

PR #79, opened 2016-03-10
Status Monitoring - Quality Graph Tooltip Values

NOYB:
Don't know what may be planned for making the values more humanly friendly.  So sharing this that I like.
We mere humans just don't relate well to long strings of digits.  Displaying the quality graph delay in milliseconds is much more human friendly.

Probably a more elegant means but this will suffice for now.


jdillard:
I think the best implementation would be a lookup table of sorts, I'd have to play with the best way to populate it though. Possibly generate an array in local storage that can be referenced.

jdillard:
Here it is implemented with local storage: 04d3576ec04ae95fdeed77fcee7d0a3223e5250d

Looked good in the testing I did.

PR #82, opened 2016-03-12
Can't load a default resolution for a value that is disabled.

NOYB:


jdillard:
I ended up doing it different (and I think simpler) way before seeing this in commit 181e6d7f75832447c4aa2c19314c25fefe8e29d5. Test it out and see if it works for you.

PR #105, opened 2016-04-13
Automatically update apinger to dpinger

phil-davis:
in the Service Watchdog watch list.
Forum: https://forum.pfsense.org/index.php?topic=109770.msg611158#msg611158

Doing it here means that people will get the automatic change as soon as they upgrade to 2.3 and the upgrade installs the Service Watchdog code from here. The other place it could be done is in upgrade_config.inc - but then that will not be distributed to them until pfSense 2.3.1 happens - rather too late.

### DIFF
--- a/sysutils/pfSense-pkg-Service_Watchdog/files/usr/local/pkg/servicewatchdog.inc
+++ b/sysutils/pfSense-pkg-Service_Watchdog/files/usr/local/pkg/servicewatchdog.inc
@@ -106,7 +106,12 @@ function servicewatchdog_check_services() {
 	}
 	$a_pwservices = &$config['installedpackages']['servicewatchdog']['item'];
 
-	foreach ($a_pwservices as $svc) {
+	foreach ($a_pwservices as $idx => $svc) {
+		// apinger became dpinger in pfSense 2.3
+		if ($svc['name'] == 'apinger') {
+			$a_pwservices[$idx]['name'] = $svc['name'] = 'dpinger';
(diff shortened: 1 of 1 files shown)


sysutils/pfSense-pkg-Service_Watchdog/files/usr/local/pkg/servicewatchdog.inc
```diff
@@ -106,7 +106,12 @@ function servicewatchdog_check_services() {
 	}
 	$a_pwservices = &$config['installedpackages']['servicewatchdog']['item'];
 
-	foreach ($a_pwservices as $svc) {
+	foreach ($a_pwservices as $idx => $svc) {
+		// apinger became dpinger in pfSense 2.3
+		if ($svc['name'] == 'apinger') {
+			$a_pwservices[$idx]['name'] = $svc['name'] = 'dpinger';
+			write_config(gettext("Service Watchdog updated service apinger to dpinger"));
```
jim-p: I like the idea of this fix, but I don't like doing a write_config here. Even if it only triggers once, it's not an ideal place.

It could assume by apinger they meant dpinger without the save, perhaps fixing it permanently next time they make a save in the GUI.

phil-davis: I made PR #106 which has the simplified version without write_config() - take your pick.

phil-davis:
Closing in favor of #106

PR #113, opened 2016-04-19
Fixed password length for squid local user Bug #5940

SilvioGiunge:
The basic_ncsa_auth authentication helper with DES only accecpt passwords with 8 characters or less.
By default the crypt function of PHP uses DES, it will change the squid_resync_users function to save the password with MD5-based hashing


rbgarga:
Please bump PORTREVISION

jim-p:
I committed a better change to handle this using SHA512.

PR #122, opened 2016-04-27
For Snort sfPortScan, add ignore_scanned option to ignore destination…

zxvv:
For snort sfPortScan, this adds a 'ignore_scanned' option, described in snort documentation as:

`ignore_scanned { <ip1 ip2/cidr[ [port1 port2-port3]]> }`
Ignores the destination of scan alerts. The parameter is the same format as that of watch_ip.
[https://www.snort.org/faq/readme-sfportscan](url)

The syntax is identical to the existing ignore_scanned option, so the php validation and selection of aliases is treated identically.

This was proposed in the forums:
[https://forum.pfsense.org/index.php?topic=110657.0](url)


zxvv:
The last two commits resolved issues with emitting aliases and CIDR values.  This is based on some testing of configuration of ignore_scanners and ignore_scanned using aliases and using CIDR addresses, with and without ports.  

I've tried to stick to consistent style, but I may not be familiar with conventions for this project, so if you wish, please feel free to modify this however needed, or tell me where it's not conforming.

One motivation was to allow one to suppress certain false positives in order to be able to raise the portscan sensitivity level.  For example, inbound ssh and openvpn can trigger a false positive, and blacklist the client address, and if so, adding these addr/ports to ignore_scanned may diminish the undesired blacklisting.

zxvv:
I've been able to keep the scan sensitivity at 'high' without locking myself out, which is an improvement for me.

Andrew17856:
I'm also awaiting this functionality - thanks for adding it.

rbgarga:
Can you please check this out @bmeeks8 ?

bmeeks8:
I am good with this addition.  Please merge the change.

Thanks,
Biill

PR #131, opened 2016-05-23
Status / Monitoring - Custom Time Period - Date/Time Defaults

NOYB:
Status / Monitoring - Custom Time Period - Date/Time Defaults

Default to current date and/or time when the fields are left empty.

This allows creating some useful default graphs such as Month to Date (MTD).
    Start Date: mm/01/yyyy
    Start Time: 0
    End Date: (left empty)
    End Time: (left empty)


gonzopancho:
ping @jdillard let's get to some resolution on this.

jdillard:
While this could be a cool feature if developed more, although I don't see enough value in this that outweighs the technical debt it accrues, as it makes the calculations and field validation increasingly more complex.

PR #133, opened 2016-05-24
bind package pfSense 2.3

svenauhagen:
Initial commit for the bind package pull request


cbuechler:
@svenauhagen could you please re-submit without the NRPE commits? Though it's effectively a no-op, all that would still show in the history unnecessarily if this were merged.

PR #134, opened 2016-05-24
bind package pfSense 2.3

svenauhagen:
Initial commit bind package

@cbuechler sure, no problem.
Here is a fresh commit.

Thanks


klepallec:
Hi, 

I'm really interessing about this package.
Do you know a realease date ?

thanks

cbuechler:
Merged, thanks @svenauhagen ! 

Should show up in the next hour or so.

rbgarga:
@svenauhagen I didn't understand why you needed to create that port bind-pfsense. What is the idea behind it?

svenauhagen:
Hi,

the problem is the pfSense package depends on bind-tools which is in conflict with bind.
bind-tools is just a meta package containing a subset of files from the bind package.

So we created a package bind-pfsense containing the files from bind excluding the files already in bind-tools so the entire bind package files are installed.

I hope that makes it more understandable.

Best
Sven

PR #135, opened 2016-05-25
Add stunnel package for pfSense 2.3

djmarcin:
Mostly straight copies of the files from the 2.2 package with the new 2.3 package metadata files.

No attempt has been made to clean up the existing code, notably in security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel.inc:
- There is code checking for pfsense versions < 2.3 which can probably be deleted
- The gui attempts to render colored text, but since it gets escaped simply renders text like &lt;font color="#AABBCC"&gt;blahblah&lt;/font&gt;.  I couldn't figure out how to get colored text in the status page tables.

I am willing to work on those issues, but since this doesn't seem to have any major issues perhaps it is OK as a first approximation to get the stunnel package working again.  If the pull can't be accepted as is, any pointers on how to fix the above issues would be greatly appreciated.  I couldn't find anything promising in the convert to bootstrap docs.

### DIFF
--- a/security/pfSense-pkg-stunnel/Makefile
+++ b/security/pfSense-pkg-stunnel/Makefile
@@ -1,0 +1,45 @@
+# $FreeBSD$
+
+PORTNAME=	pfSense-pkg-stunnel
+PORTVERSION=	5.37
+CATEGORIES=	security
+MASTER_SITES=	# empty
+DISTFILES=	# empty
+EXTRACT_ONLY=	# empty
+
+MAINTAINER=	coreteam@pfsense.org
+COMMENT=	pfSense package stunnel
+
+LICENSE=	APACHE20
+
+RUN_DEPENDS=	${LOCALBASE}/bin/stunnel:security/stunnel
+
+NO_BUILD=	yes
+NO_MTREE=	yes
+
+SUB_FILES=	pkg-install pkg-deinstall
+SUB_LIST=	PORTNAME=${PORTNAME}
+
+do-extract:
+	${MKDIR} ${WRKSRC}
+
+do-install:
+	${MKDIR} ${STAGEDIR}${PREFIX}/pkg
+	${MKDIR} ${STAGEDIR}/etc/inc/priv
+	${MKDIR} ${STAGEDIR}${PREFIX}/www/shortcuts
+	${MKDIR} ${STAGEDIR}${DATADIR}
+	${INSTALL_DATA} -m 0644 ${FILESDIR}${PREFIX}/pkg/stunnel.xml \
+		${STAGEDIR}${PREFIX}/pkg
+	${INSTALL_DATA} ${FILESDIR}${PREFIX}/pkg/stunnel.inc \
+		${STAGEDIR}${PREFIX}/pkg
+	${INSTALL_DATA} ${FILESDIR}/etc/inc/priv/stunnel.priv.inc \
+		${STAGEDIR}/etc/inc/priv
+	${INSTALL_DATA} ${FILESDIR}${PREFIX}/www/shortcuts/pkg_stunnel.inc \
+		${STAGEDIR}${PREFIX}/www/shortcuts
+	${INSTALL_DATA} ${FILESDIR}${DATADIR}/info.xml \
+		${STAGEDIR}${DATADIR}
+	@${REINPLACE_CMD} -i '' -e "s|%%PKGVERSION%%|${PKGVERSION}|" \
+		${STAGEDIR}${DATADIR}/info.xml \
+		${STAGEDIR}${PREFIX}/pkg/stunnel.xml
+
+.include <bsd.port.mk>
--- a/security/pfSense-pkg-stunnel/files/etc/inc/priv/stunnel.priv.inc
+++ b/security/pfSense-pkg-stunnel/files/etc/inc/priv/stunnel.priv.inc
@@ -1,0 +1,31 @@
+<?php
+/*
+ * stunnel.priv.inc
+ *
+ * part of pfSense (https://www.pfsense.org)
+ * Copyright (c) 2015-2016 Rubicon Communications, LLC (Netgate)
+ * All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+global $priv_list;
+
+$priv_list['page-system-stunnel'] = array();
+$priv_list['page-system-stunnel']['name'] = "WebCfg - System: stunnel package";
+$priv_list['page-system-stunnel']['descr'] = "Allow access to stunnel package GUI";
+$priv_list['page-system-stunnel']['match'] = array();
+$priv_list['page-system-stunnel']['match'][] = "pkg_edit.php?xml=stunnel.xml*";
+$priv_list['page-services-stunnel']['match'][] = "shortcuts/pkg_stunnel.inc*";
+
+?>
--- a/security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel.inc
+++ b/security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel.inc
@@ -1,0 +1,130 @@
+<?php
+/*
+ * stunnel.inc
+ *
+ * part of pfSense (https://www.pfsense.org)
+ * Copyright (c) 2015-2016 Rubicon Communications, LLC (Netgate)
+ * All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+require_once("certs.inc");
+require_once("config.inc");
+require_once("pfsense-utils.inc");
+require_once("util.inc");
+require_once('services.inc');
+require_once('service-utils.inc');
+
+define('STUNNEL_LOCALBASE', '/usr/local');
+define('STUNNEL_ETCDIR', STUNNEL_LOCALBASE . "/etc/stunnel");
+
+function stunnel_save() {
+	global $config;
+	conf_mount_rw();
+	safe_mkdir(STUNNEL_ETCDIR, 0755);
+	$fout = fopen(STUNNEL_ETCDIR . "/stunnel.conf", "w");
+	fwrite($fout, "cert = " . STUNNEL_ETCDIR . "/stunnel.pem \n");
+	fwrite($fout, "chroot = /var/tmp/stunnel \n");
+	fwrite($fout, "setuid = stunnel \n");
+	fwrite($fout, "setgid = stunnel \n");
+	if (!is_array($config['installedpackages']['stunnel']['config'])) {
+		$config['installedpackages']['stunnel']['config'] = array();
+	}
+
+	// Keep array of in-use certificates so we can clean up after ourselves.
+	$in_use_certs = array();
+	$in_use_certs[] = 'stunnel.pem';
+
+	foreach ($config['installedpackages']['stunnel']['config'] as $pkgconfig) {
+		fwrite($fout, "\n[" . $pkgconfig['description'] . "]\n");
+		if ($pkgconfig['client']) {
+			fwrite($fout, "client = yes" . "\n");
+		}
+		if ($pkgconfig['certificate'] && $pkgconfig['certificate'] != 'default') {
+			$cert = lookup_cert($pkgconfig['certificate']);
+			if ($cert != false) {
+				file_put_contents(STUNNEL_ETCDIR . "/{$pkgconfig['certificate']}.pem", 
+					base64_decode($cert['prv']) . base64_decode($cert['crt']) . ca_chain($cert));
+				fwrite($fout, "cert = " . STUNNEL_ETCDIR . "/{$pkgconfig['certificate']}.pem\n");
+				$in_use_certs[] = "{$pkgconfig['certificate']}.pem";
+			}
+		}
+		if ($pkgconfig['sourceip']) {
+			fwrite($fout, "local = " . $pkgconfig['sourceip'] . "\n");
+		}
+		fwrite($fout, "accept = " . ($pkgconfig['localip'] ? $pkgconfig['localip'] . ":" : "")  . $pkgconfig['localport'] . "\n");
+		fwrite($fout, "connect = " . $pkgconfig['redirectip'] . ":" . $pkgconfig['redirectport'] . "\n");
+		fwrite($fout, "TIMEOUTclose = 0\n\n");
+	}
+	fclose($fout);
+
+	// Clean up certs that are no longer in use.
+	foreach (glob(STUNNEL_ETCDIR . "/*.pem") as $file) {
+		if (!in_array(basename($file), $in_use_certs)) {
+			unlink($file);
+		}
+	}
+
+	conf_mount_ro();
+
+	restart_service("stunnel");
+}
+
+function stunnel_install() {
+	global $config;
+	safe_mkdir(STUNNEL_ETCDIR);
+
+	// Generate a self-signed default certificate if one does not already exist.
+	$stunnel_pem_filename = STUNNEL_ETCDIR . "/stunnel.pem";
+	if (!file_exists($stunnel_pem_filename)) {
+		$cert = array();
+		$cert['refid'] = uniqid();
+		$cert['descr'] = sprintf(gettext("stunnel default (%s)"), $cert['refid']);
+
+		$dn = array(
+			'countryName' => "US",
+			'stateOrProvinceName' => "State",
+			'localityName' => "Locality",
+			'organizationName' => "{$g['product_name']} stunnel Self-Signed Certificate",
+			'emailAddress' => "admin@{$config['system']['hostname']}.{$config['system']['domain']}",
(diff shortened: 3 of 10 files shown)


security/pfSense-pkg-stunnel/Makefile
```diff
@@ -0,0 +1,47 @@
...
+PORTREVISION=	1
+CATEGORIES=	security
+MASTER_SITES=	# empty
+DISTFILES=	# empty
+EXTRACT_ONLY=	# empty
+
+MAINTAINER=	coreteam@pfsense.org
+COMMENT=	pfSense package stunnel
+
+LICENSE=	ESF
```
rbgarga: License changed to APACHE20

security/pfSense-pkg-stunnel/files/etc/inc/priv/stunnel.priv.inc
```diff
@@ -0,0 +1,39 @@
...
+	THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+	INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+	AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+	AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+	OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+	SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+	INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+	CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+	ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+	POSSIBILITY OF SUCH DAMAGE.
```
rbgarga: License has changed to APACHE20, replace text accordingly. You can see an example at:
https://github.com/pfsense/FreeBSD-ports/blob/devel/security/pfSense-pkg-sudo/files/etc/inc/priv/sudo.priv.inc#L2

security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel.inc
```diff
@@ -0,0 +1,250 @@
...
+	THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+	INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+	AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+	AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+	OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+	SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+	INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+	CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+	ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+	POSSIBILITY OF SUCH DAMAGE.
```
rbgarga: Fix license text

security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel.inc
```diff
@@ -0,0 +1,250 @@
...
+*/
+global $config;
+$pf_version = substr(trim(file_get_contents("/etc/version")), 0, 3);
+if ($pf_version == "2.1" || $pf_version == "2.2") {
+	define('STUNNEL_LOCALBASE', '/usr/pbi/stunnel-' . php_uname("m"));
+} else {
+	define('STUNNEL_LOCALBASE', '/usr/local');
+}
```
rbgarga: We don't need to care about pfSense 2.1 or 2.2 here, this conditional can be removed

security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel.inc
```diff
@@ -0,0 +1,250 @@
...
+define('STUNNEL_ETCDIR', STUNNEL_LOCALBASE . "/etc/stunnel");
+if (!isset($_GET['id']) and !isset($_POST['id'])) {
+	if ($GLOBALS['config']['installedpackages']['stunnelcerts']['savemsg']) {
```
rbgarga: Remove all use of $GLOBALS. The standard way in pfSense is to declare 'global $config' and access $config direct

security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel.xml
```diff
@@ -0,0 +1,174 @@
...
+	THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+	INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+	AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+	AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+	OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+	SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+	INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+	CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+	ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+	POSSIBILITY OF SUCH DAMAGE.
```
rbgarga: Replace license text

security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel_certs.xml
```diff
@@ -0,0 +1,143 @@
...
+	THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+	INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+	AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+	AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+	OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+	SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+	INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+	CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+	ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+	POSSIBILITY OF SUCH DAMAGE.
```
rbgarga: Replace license text

security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel_certs.xml
```diff
@@ -0,0 +1,143 @@
...
+*/
+/* ====================================================================================== */
+	]]>
+	</copyright>
+	<name>stunnelcerts</name>
+	<version>5.20.2</version>
```
rbgarga: Replace 5.20.2 by macro %%PKGVERSION%%

security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel.xml
```diff
@@ -0,0 +1,174 @@
...
+*/
+/* ====================================================================================== */
+	]]>
+	</copyright>
+	<name>stunnel</name>
+	<version>5.20.3</version>
```
rbgarga: Replace 5.20.3 by %%PKGVERSION%%

security/pfSense-pkg-stunnel/Makefile
```diff
@@ -0,0 +1,47 @@
...
+	${INSTALL_DATA} -m 0644 ${FILESDIR}${PREFIX}/pkg/stunnel_certs.xml \
+		${STAGEDIR}${PREFIX}/pkg
+	${INSTALL_DATA} ${FILESDIR}/etc/inc/priv/stunnel.priv.inc \
+		${STAGEDIR}/etc/inc/priv
+	${INSTALL_DATA} ${FILESDIR}${PREFIX}/www/shortcuts/pkg_stunnel.inc \
+                ${STAGEDIR}${PREFIX}/www/shortcuts
+	${INSTALL_DATA} ${FILESDIR}${DATADIR}/info.xml \
+		${STAGEDIR}${DATADIR}
+	@${REINPLACE_CMD} -i '' -e "s|%%PKGVERSION%%|${PKGVERSION}|" \
+		${STAGEDIR}${DATADIR}/info.xml
```
rbgarga: Add XML files that contain macro %%PKGVERSION%% to this REINPLACE_CMD statement

security/pfSense-pkg-stunnel/Makefile
```diff
@@ -0,0 +1,47 @@
+# $FreeBSD$
+
+PORTNAME=	pfSense-pkg-stunnel
+PORTVERSION=	5.30
+PORTREVISION=	1
```
rbgarga: New packages shouldn't contain PORTREVISION line, please remove

doktornotor:
I fixed most of the review comments (plus additional things unnoticed by @rbgarga) in https://github.com/djmarcin/FreeBSD-ports/pull/1

I have not touched any of the $GLOBALS stuff since I think the page should use certs.inc and store certs for the package in System - Cert Manager, rather than reimplementing it.

djmarcin:
I merged in the changes from @doktornotor

rbgarga:
As suggested by @doktornotor, change it to use certs.inc instead of reimplementing it

djmarcin:
I'll probably be able to look at this over the weekend, but to be honest
I'm not terribly familiar with the certs part of stunnel, I'm using it
without certs.

doktornotor:
@djmarcin - Need some help here?

djmarcin:
This fell off my radar, but I looked at it last night. To clarify the goal
here, we want to eliminate the certs page in the stunnel config UI and just
allow the user to choose from the system cert manager storage?

doktornotor:
Yeah, exactly, the XML can just let users pick up certificates for each tunnel from those set up in the Cert Manager, see e.g. https://github.com/pfsense/FreeBSD-ports/blob/devel/www/pfSense-pkg-squid/files/usr/local/pkg/squid_reverse_general.xml

Plus if you need code to do something with the certificates, use certs.inc and the functions in there if possible.

djmarcin:
Ok -- I've removed the custom certs implementation and replaced it with certs.inc functions.  As far as I can tell, stunnel needs access to .pem files in order to function, and I couldn't figure out if the system cert manager provided access to such files -- I think not? So I pull the referenced certs on save and write out .pem files in the stunnel directory and delete them when they're no longer in use by stunnel.

As to whether or not stunnel actually works correctly with these certs, I *think* so but I haven't tested it rigorously.  Will give it some more testing tomorrow.

doktornotor:
Looks much better. (And yeah, you need to write out the certificates somewhere to filesystem, nothing wrong with that.)

djmarcin:
Finally got a chance to do that testing.  Verified that stunnel using a Let's Encrypt cert imported to cert manager successfully acted in server mode and provided SSL access to pfsense being served over the HTTP port, so it looks like this is working properly.

Forgive my ignorance, do I need to do anything to mark the requested changes done?

security/pfSense-pkg-stunnel/Makefile
```diff
@@ -0,0 +1,45 @@
...
+	${MKDIR} ${STAGEDIR}${PREFIX}/www/shortcuts
+	${MKDIR} ${STAGEDIR}${DATADIR}
+	${INSTALL_DATA} -m 0644 ${FILESDIR}${PREFIX}/pkg/stunnel.xml \
+		${STAGEDIR}${PREFIX}/pkg
+	${INSTALL_DATA} ${FILESDIR}${PREFIX}/pkg/stunnel.inc \
+		${STAGEDIR}${PREFIX}/pkg
+	${INSTALL_DATA} ${FILESDIR}/etc/inc/priv/stunnel.priv.inc \
+		${STAGEDIR}/etc/inc/priv
+	${INSTALL_DATA} ${FILESDIR}${PREFIX}/www/shortcuts/pkg_stunnel.inc \
+                ${STAGEDIR}${PREFIX}/www/shortcuts
```
rbgarga: Use TAB instead of spaces here, otherwise portlint will complain

security/pfSense-pkg-stunnel/pkg-plist
```diff
@@ -0,0 +1,6 @@
+pkg/stunnel.xml
+pkg/stunnel.inc
+www/shortcuts/pkg_stunnel.inc
+/etc/inc/priv/stunnel.priv.inc
+%%DATADIR%%/info.xml
+@dir /etc/inc/priv
```
rbgarga: It's missing '@dir /etc/inc' item here

djmarcin:
Done.

I noticed something while doing this.  The config file is being clobbered on installation by a version without tunnel definitions.  I attempted to fix this by replacing the deleted lines in c9d58d9 with stunnel_save() but it still didn't work.  The pfSense UI still shows the tunnels, and after saving any tunnel, all the tunnels get written out again, however it doesn't seem like the script is aware of this config during stunnel_install().

Is there another function I can implement that would be called after install but have access to the config stored by pfSense?

doktornotor:
@djmarcin You mean like putting

```
<custom_php_resync_config_command>
	stunnel_save();
</custom_php_resync_config_command>
```

to the XML, or what is exactly is the issue?

security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel.inc
```diff
@@ -0,0 +1,133 @@
...
+$tunnels = $config['installedpackages']['stunnel']['config'];
+is_array($tunnels) ? $num_tunnels = count($tunnels) : $num_tunnels = 0;
+if (!isset($_GET['id']) and $num_tunnels) {
+	for ($i = 0; $i < $num_tunnels; $i++) {
+		$tunnel = $tunnels[$i];
+	}
+}
+
+function stunnel_save($config) {
+	conf_mount_rw();
```
jim-p: `conf_mount_rw()` and `conf_mount_ro()` calls can be removed. It is a no-op on 2.3 and irrelevant on 2.4

security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel.inc
```diff
@@ -0,0 +1,133 @@
...
+
+	conf_mount_ro();
+	stop_service("stunnel");
+	start_service("stunnel");
+}
+
+function stunnel_install() {
+	safe_mkdir(STUNNEL_ETCDIR);
+
+	// Generate a self-signed default certificate.
```
jim-p: You should check here to make sure such a cert does not already exist, otherwise any time the package is upgraded or reinstalled, it will make another similar certificate.

security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel.inc
```diff
@@ -0,0 +1,133 @@
...
+	fclose($fout);
+
+	// Clean up certs that are no longer in use.
+	foreach (glob(STUNNEL_ETCDIR . "/*.pem") as $file) {
+		if (!in_array(basename($file), $in_use_certs)) {
+			unlink($file);
+		}
+	}
+
+	conf_mount_ro();
```
jim-p: `conf_mount_rw()` and `conf_mount_ro()` calls can be removed. It is a no-op on 2.3 and irrelevant on 2.4

rbgarga: It's a noop on 2.4 but still used on 2.3 since 2.3 still supports nanobsd. I suggest to keep them for now so this package can be cherry-picked to 2.3

djmarcin: kept for now

security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel.inc
djmarcin: done

djmarcin:
There's just one problem with this which is that when the package is upgraded (well, with pkg install) it doesn't seem to start the service.  It will only start the service after some edit is made.

djmarcin:
It appears there is a bug with either pfSense or stunnel where start_service does not work in the context of 'pkg install' or 'pkg upgrade'.  The bug causes the service not to daemonize and become adopted by init -- instead the processes is adopted by the pkg command and then killed when that exits.  See my post in https://forum.pfsense.org/index.php?topic=111563.0 for some details.

The service is properly started on boot or on config changes though.

Added two minor cleanup edits to the branch.  Let me know if you want me to squash these or anything.

security/pfSense-pkg-stunnel/Makefile
djmarcin: Done (responding so it gets marked complete)

security/pfSense-pkg-stunnel/pkg-plist
djmarcin: Done (responding so it gets marked complete)

rbgarga:
@djmarcin package looks OK now but I believe squashing all these commits is a good idea and it'll make it easy to add it to 2.3.3 when it's properly tested for some time. Can you do it please?

djmarcin:
Sorry for the delay, commits have been squashed to one commit, attribution for the changes by doktornotor via "Co-authored-by" in this commit.

PR #139, opened 2016-05-31
Status / Monitoring - Last Sample Zero

NOYB:
Monitoring graphs last sample being zero

If there is no RRD data sample for the current minute then the graph and last and minimum data value will be zero in certain circumstances.  Such as resolution 1 minute and graph loaded prior to the current minute RRD sample being made.  Similar for the first minute of other resolutions.

This PR corrects this situation by never requesting RRD data for the uncompleted current minute to ensure that there is a data point for the last/end time.


jdillard:
According to this page: http://linux.die.net/man/1/rrdfetch

You can do something like:

> `end-3weeks` or `e-3w` -- 3 weeks before end time (may be used as start time specification).

which seems cleaner to me.

jdillard:
I ended up doing it a little different in a93f57d657e2fec4420a993e5a778adeec6fc931. Notice how there was still a hack in place to remove the last value, which is no longer needed. 

I just noticed #145, I'll take a look at that too.

PR #141, opened 2016-05-31
Fix null unit on status monitoring graph

strigona-worksight:
When localStorage is cleared and you add a right axis field that has units other than the main left axis, the units will be "null" due to the tempKey stripping out the "(right axis)" from the key and thus not finding a match in localStorage.
2016-05-19_09-16-37_firefox


pfSense-git-updates:

strigona-worksight:
Done

jdillard:
Thanks! I'll test this out tomorrow, but at a glance it looks like a great catch.

PR #146, opened 2016-06-08
Status / Monitoring - InfoBox - s/axes/axis/

NOYB:


jdillard:
axes is the plural from of axis.

PR #148, opened 2016-06-10
add GeoIP as a dependency of ntopng

zxvv:
There were reports that ntopng was failing to restart, discussed here:
https://redmine.pfsense.org/issues/6443

There is a missing dependency on GeoIP
I tried to restart the service, which failed.
Then logged in and tried to start it from the command line, as follows.
/usr/local/etc/rc.d/ntopng.sh start
[2.3.2-DEVELOPMENT][root@pf232test.lan]/root: Shared object "libGeoIP.so.1" not found, required by "ntopng"

find / -name '*GeoIP*' showed nothing.
It was resolved by:
pkg install GeoIP-1.6.9
/usr/local/bin/geoipupdate.sh
/usr/local/etc/rc.d/ntopng.sh start
which successfully started ntopng.


rbgarga:
Please bump PORTREVISION or PORTVERSION otherwise a new package will never show up

zxvv:
Thank you for the pointer.  Just let me know if there's anything else I can do.

lunetics:
+1. can that be published?

cbuechler:
I committed a slightly modified version of this, bumping the REVISION as well in one commit. 

Thanks @zxvv

PR #152, opened 2016-07-01
Bump Snort binary to v2.9.8.3 to sync with upstream.

bmeeks8:
Update Snort binary package to version 2.9.8.3 to match upstream and since version 2.9.8.0 is EOL for  rules updates.


cbuechler:
Merged, thanks @bmeeks8 ! Will follow up on your email momentarily.

PR #154, opened 2016-07-16
Small cosmetic patch to System Patches list view

lukehamburg:
When viewing installed patch list on 2.3+ the GitHub commit URLs are quite long and cause the main table to extend beyond the screen boundary. This looks weird and forces you to scroll to access the action buttons.

Seeing the commit URL on the list doesn't add much as far as usability (and it can be seen anyway if needed on the edit page). This patch makes the page display properly.


cbuechler:
Yeah that's cleaner. Merged, thanks @lukehamburg

PR #157, opened 2016-07-18
pfBlockerNG - Add dnsbl_tld file

BBcan177:
* Add dnsbl_tld file, which is used by the TLD process to determine if all Sub-Domains should be blocked for each listed Domain.


rbgarga:
Merged, thanks!

PR #158, opened 2016-07-19
Fix reload openbgpd config

brunostein:
Changed the 'openbgpd_restart' function to reload the new config after save.


rbgarga:
Manually merged since your branch was deleted

PR #159, opened 2016-07-26
Missing files from prior commit

dennypage:
These files are missing from the prior commit. Apologies Chris!


sysutils/pfSense-pkg-nut/files/etc/inc/priv/nut.priv.inc
```diff
@@ -15,7 +17,7 @@
 	   notice, this list of conditions and the following disclaimer in the
 	   documentation and/or other materials provided with the distribution.
 
-	THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+	THIS SOFTWARE IS PROVIDED "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
```
rbgarga: Please do not change license text. I'll import it to unbreak the ports build and will revert it after. Thanks

dennypage: It wasn't a change per se, it's actually a completely new file with the same name. All the files in the new package have the same copyright using double quotes.

Why is there a preference for  ticks vs double quotes?

rbgarga: It's not a preference, just avoiding to have different versions of license text around

dennypage: Okay. I have created a PR to change them all. Sorry for the difficulty.

PR #160, opened 2016-07-26
Change copyright notice

dennypage:
Change copyright notice from double quote to ticks to match copyright notices elsewhere in pfSense per Renato.


rbgarga:
Thank you very much!

PR #161, opened 2016-07-27
Error installing AutoConfigBackup

leozoccoli:
When trying to install the process is looped . Please find attached the print with error.

Please help me , I bought the gold package pfsense only by this package .
erro gold


jim-p:
Please contact us via support and we can help you get the package installed.
This PR is attempting to change many unrelated files and this is not a means of requesting assistance.

PR #166, opened 2016-08-07
[RELENG_2_3_2] Unbreak info.xml in pfSense-pkg-nut

davidjwood:
Remove the erroneous second `<pkginfolink>...</pkginfolink>` that has crept into info.xml on the RELENG_2_3_2 branch.


davidjwood:
These changes have been made in 678436e0f0d4030026caa1a188b00c4c474fc340 , so this PR is now redundant.

rbgarga:
Yeah, I didn't notice the PR before I push the fix. Thanks!!

PR #171, opened 2016-08-16
List remote clients, connected to UPS

SunStroke74:
In UPS Status widget shows remote clients, currently connected (if NUT port is 3493 by default).
$raw_data needs to pass `<br />` to output to print clients line by line.

### DIFF
--- a/sysutils/pfSense-pkg-nut/files/usr/local/www/widgets/widgets/nut_status.widget.php
+++ b/sysutils/pfSense-pkg-nut/files/usr/local/www/widgets/widgets/nut_status.widget.php
@@ -40,14 +40,14 @@ if ($_REQUEST && $_REQUEST['ajax']) {
 }
 
 
-function print_row($desc, $value) {
+function print_row($desc, $value, $raw_data = false) {
 	print '<tr>';
 	print '  <td style="width:35%"><b>' . $desc . ':</b></td>';
-	print '  <td>' . htmlspecialchars($value) . '</td>';
+	print '  <td>' . ($raw_data ? $value : htmlspecialchars($value)) . '</td>';
 	print '</tr>';
 }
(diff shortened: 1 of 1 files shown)


dennypage:
Raw data really should not be used in the widget. Either "remote users" would a first class citizen and in a separate section, or should not be in the widget at all.

Honestly, I don't see listing remote clients as being appropriate for the NUT widget. The widget is focused on run-time for pfSense itself. And given that remote client connectivity itself is not a first class citizen of the NUT configuration, it really doesn't make sense to put it in the widget.

SunStroke74:
Yes, it's not base functionality of widget. But it would be helpful to see is other servers in the rack (if any) sucsessfully connected to same UPS and controlled by. Of couse we can see that in the log, but use widget is more comfortable.

Whatever it was, you decide is we need it or not. Simply close request if not.

rbgarga:
@SunStroke74 upsc binary has a parameter to list connected clients, upsc -c, isn't it enough to get data instead of all these hacks with pfctl states?

dennypage:
Yes, upsc has support for listing remote connections. However, the NUT package does not directly support remote connections. It really doesn't make sense to put something in the widget that isn't available in the settings or status page.

While I document how to support remote connections in the info thread, I recommend against the practice unless absolutely necessary. If you need multi-host, it's better from a security pov to connect the UPS to upsd on another host, and run NUT on the firewall as a remote client to that. Hosting on the firewall is a last resort approach.

SunStroke74:
Yes, `upsc -c {$status['_name']} | grep -v '::'` is better than pfctl. I do not know this command.
And I close request as your arguments is weighty. Those who want can add this functionality through custom patch. Mayby on forum we can publish this patch link.

PR #172, opened 2016-08-17
Support https

dennypage:
Ntopng now uses the same http(s) configuration as the webgui. Necessary following the introduction of HSTS.


dennypage:
Thank you Renato!

PR #173, opened 2016-08-17
Fix path for fetch patches (System Patches package)

SunStroke74:
Version 2.3 github repository is on pfsense/FreeBSD-ports, not pfsense/pfsense (was in version 2.2 and early).


jim-p:
The main pfSense code repo is still pfsense/pfsense. Packages moved from pfsense/pfsense-packages to pfsense/FreeBSD-Ports, but that is not where this feature ever pulled from.

PR #182, opened 2016-09-04
Fix Bug #6099 (igmpproxy does not recognize upstream interface)

NewEraCracker:
Here goes nothing :smiley: 
https://redmine.pfsense.org/issues/6099

Hope this fills the gaps. I've tested it on 2.3.3 and recognized all intended interfaces.
Please backport it there, thanks.

PS: This repository is huge, cloning with small depth is still an herculean task, same with git commands that take about 60 seconds to do anything...


NewEraCracker:
I've also forgot to update PORTREVISION so I'd appreciate if that is done. Thanks.

NewEraCracker:
Patch has been reworked with a new commit. Keeping previous commit for historic purposes on the evolution of the code. I've also bumped PORTREVISION.

NewEraCracker:
Reworking this to fix another bug regarding subnet handling on config.c

On ifvc.c subnets are handled as:
```subnet = addr & mask;
```allowednet->subnet_addr = subnet```
while on config.c:
```tmpSubnet->subnet_addr = addr;```
adding a patch to change that to:
```tmpSubnet->subnet_addr = (addr & mask);```

PR #187, opened 2016-09-13
Added PKCS#11 support.

skrude61:


jim-p:
I'm not a fan of adding stored settings to the package. It's a slippery slope that we don't yet have a clean solution for. It would have to support per-server settings, not just globals, and though this seems to do the trick it's not quite complete. The PKCS#11 changes seem OK, but not the stored settings part.

skrude61:
OK, good point. New PR submitted.

skrude61:
Sorry, wrong PR.

PR #188, opened 2016-09-15
Suricata, a broken download should not wait forever.

PiBa-NL:


rbgarga:
@bmeeks8 can you please review?

PR #191, opened 2016-09-18
Bind9 Error on ipv4 and ipv6 enabled restart the running daemon is not recognized

svenauhagen:
Hi,

this commit fixes the bug that the daemon is not recognized if IPv4 and IPv6 are both enabled.

Best
Sven


svenauhagen:
Hi,

sorry I forgot to bump up the version.
I pushed the fix.

Best
Sven

PR #192, opened 2016-09-19
Update Suricata binary to 3.1.1 to match upstream.

bmeeks8:
## Suricata 3.1.1##
This updates the Suricata binary to version 3.1.1 to match upstream FreeBSD ports.  This version contains several fixes for Netmap and other issues identified and patched by upstream.


rbgarga:
I've cherry-picked FreeBSD commits from master branch instead of manually patch the port. I also re-generated patch-alert-pf.diff to make it apply cleanly on 3.1.1.

New binaries are available for 2.4 and 2.3.3 snapshots. As soon as GUI package is declared stable I can take care of merging everything to 2.3.2

PR #194, opened 2016-09-22
Do not list client exports for revoked certificates.

skrude61:
Exporting clients for revoked certificates makes no sense.


jim-p:
It's a good start but it needs to be a stricter check. The fact that a cert is revoked doesn't mean it can't be used at all. Someone could have the same CA on two servers, with different CRLs chosen on each, but only have a specific cert revoked in one of them. So in that example, the certificate in question should be listed on one server and not on the other, depending on the chosen CRL.

So what this should do is only check the cert to see if it's revoked against the CRL chosen for this specific OpenVPN instance. If the OpenVPN instance has no CRL selected, then it should not check at all.

Thanks!

skrude61:
OK, good point. New PR submitted.

PR #197, opened 2016-09-25
pkg-bind: Slaves must be able to NOTIFY other Secondaries

hb9cwp:
The default behaviour of BIND is that Slaves will send NOTIFY to other Secondaries. In pfSense, `notify no` in all Slave zones prevents this default behaviour, which is unexpected and looks like a bug in BIND until inspecting named.conf generated by pfSense for Slaves.

There is no way in the WebUI of pfSense to overcome this `notify no`, even when putting `notify explicit {}` into Custom Options of surrounding views for example. Even worse, BIND will throw an error when putting another `notify yes|explicit {}` into Custom Options of zones, and named will subsequently fail to start.

Admins who do want to prevent their Slaves from sending NOTIFY should put `notify no` explicitly into the Custom Options.

In my use case however, I need my Slaves to send NOTIFY to I/AXFR servers of the global anycast Secondaries at DNSMadeEasy.com using in the Custom Options of the view 
`notify explicit;
also-notify {
 208.94.147.135;
 208.94.150.198;
 63.219.151.12;
};`

This only works with the proposed PR.


pfSense-git-updates:

hb9cwp:
CLA signed & executed.

rbgarga:
Merged manually. Thanks!

PR #198, opened 2016-09-29
Added PKCS#11 support.

skrude61:
Add support for PKCS#11 certificate store (e. g. SafeNet eToken).
Settings can be saved as default (for each server), so only the individual PKCS#11 ID must be filled in before client export.
Only tested with Microsoft Windows client, not with Viscosity.

### DIFF
--- a/security/pfSense-pkg-openvpn-client-export/files/usr/local/www/vpn_openvpn_export.php
+++ b/security/pfSense-pkg-openvpn-client-export/files/usr/local/www/vpn_openvpn_export.php
@@ -108,6 +108,51 @@ foreach ($a_server as $sindex => $server) {
 	$ras_server[] = $ras_serverent;
 }
 
+$a_config =& $config['installedpackages']['openvpnclientexport'];
+$hashes = array_map(function($s){return('server_'.md5($s['name']));},$ras_server);
+foreach ($a_config as $shash => $cfg) {
+	$any_change = false;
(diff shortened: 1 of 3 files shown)


jim-p:
I still see some of the persistent settings code being added here. That all has to be removed before it can be accepted.

skrude61:
I have no clue what you meant with "I still see some of the persistent settings code being added here".

security/pfSense-pkg-openvpn-client-export/files/usr/local/www/vpn_openvpn_export.php
```diff
@@ -108,6 +108,51 @@
 	$ras_server[] = $ras_serverent;
 }
 
+$a_config =& $config['installedpackages']['openvpnclientexport'];
```
jim-p: This code is looking for stored settings, the page to store the settings was removed but not this code.

skrude61:
The settings are now stored on the page itself, separate for each OVPN server, so I do not see the problem.

jim-p:
I had asked for the stored settings code to be removed before it could be accepted, however. That is a feature that needs a lot more thought/design work before implementation. The two features (Stored settings and PKCS#11) are distinct and shouldn't be made in the same pull request.

skrude61:
Jim, you accepted PR #209. Did you change your mind regarding stored settings?

PR #199, opened 2016-10-03
System patches logic simplifications (resubmit of PR #47)

stilez:
1) Nested IF statements can be combined
 2) Add/remove shellcmd contains logic that only needs to run if the early shellcmd array does/doesn't exist
 3) Logs should be non-vague ("...added/removed a shellcmd" is better if it states the command added/removed, and for removal, clarify it removed an existing command)

Re-resubmit, as originally against old pfsense-packages, see pfsense/pfsense-packages#1229, then conflicts prevented merge (PR#47 here)

### DIFF
--- a/sysutils/pfSense-pkg-System_Patches/Makefile
+++ b/sysutils/pfSense-pkg-System_Patches/Makefile
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-System_Patches
-PORTREVISION=	1
+PORTVERSION=	1.1.5
 CATEGORIES=	sysutils
 MASTER_SITES=	# empty
 DISTFILES=	# empty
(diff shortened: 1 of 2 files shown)


sysutils/pfSense-pkg-System_Patches/Makefile
```diff
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-System_Patches
-PORTVERSION=	1.1.4
+PORTVERSION=	1.1.5
 PORTREVISION=	1
```
rbgarga: Please remove PORTREVISION line when PORTVERSION is bumped

stilez: done

stilez: done

stilez: done

stilez:
@rbgarba - PORTREVISION removed

PR #200, opened 2016-10-09
suricata, make xmlrpc 2.4 compatible

PiBa-NL:


rbgarga:
@bmeeks8 could you please check this out?

@PiBa-NL please do not mix functional and non-functional changes (style fixes) in the same commit, it's hard to review and if, for some reason, functional change must be reverted we lose all style fixes together

bmeeks8:
These changes look OK to me.  I am good with accepting the PR.  I do agree with @rbgarga that style fixes and functional logic changes should be posted in separate pull requests, though.

PiBa-NL:
Is it OK like this? I split up the changes into two separate commits.

bmeeks8:
Yes, I that is what @rbgarga and I were saying.  It makes it easier to revert if necessary.  Thank you for your contribution to the Suricata package.

PR #208, opened 2016-10-19
openvpn-client-export, added a 'save defaults' button and search option

PiBa-NL:
openvpn-client-export, added a 'save defaults' button and search option in user/certificate list, also moved the export buttons out of the list.

### DIFF
--- a/security/pfSense-pkg-openvpn-client-export/Makefile
+++ b/security/pfSense-pkg-openvpn-client-export/Makefile
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-openvpn-client-export
-PORTVERSION=	1.3.11
+PORTVERSION=	1.3.12
 CATEGORIES=	security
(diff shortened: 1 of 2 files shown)


jim-p:
I tried this out and it seems promising but didn't work for me aside from saving the settings.

If we're going to store settings, the "default" should be saved per-server rather than for all. Different servers may have different advanced options or host name options, for example.

The biggest problem is that the exporting didn't work at all. I tried three basic scenarios: SSL only, User Auth only, and SSL+User auth. In all cases, I could select a client but the export buttons did nothing at all. I tried in FF and Chrome, both up-to-date.

I'd prefer to see the export buttons remain by the clients, moving them outside of the list makes it much more awkward and non-intuitive to use. It wasn't obvious to the user how they were supposed to proceed.

So I do like the way the settings are handled, and the searching, but it needs some work yet.

PiBa-NL:
New commit, that only provides the search option at this moment. The other parts will follow later with separate requests.

PR #209, opened 2016-10-21
openvpn-client-export, added a 'save defaults' button

PiBa-NL:
openvpn-client-export, added a 'save defaults' button

Defaults are saved 'per server'.

Not bumping version in the Makefile in this pullrequest as it would then certainly conflict with the other pullrequest. #208


jim-p:
Looks like if there are no settings, the user is not shown any certs to export: https://forum.pfsense.org/index.php?topic=120099.0

PR #212, opened 2016-10-30
haproxy-devel, request syslogsocket in chroot environment, fix 'transparent-client-ip' function using ipfw on 2.4

PiBa-NL:
haproxy-devel, request syslogsocket in chroot environment, fix 'transparent-client-ip' function using ipfw on 2.4


rbgarga:
Files from this repo will never be used for pfSense < 2.3. So for next upgrades you can consider to remove all these old code.

PR #215, opened 2016-11-21
pkg-bind: Correct filenames so DNSSEC signs Reverse Zones

hb9cwp:
Without this change, inline signing fails to sign Reverse Zones because BIND expects to find the key files which were generated by ```dnssec-keygen``` with filenames ```/cf/named/etc/namedb/keys/K171.168.192.in-addr.arpa.+00...```.

Still, this patch does not fix the problem where keys are restored from DNSSEC backup file for a zone that was previously created, then deleted, and finally re-created. Currently, when re-creating the signed Reverse Zone, it will restore the keys from the XML config using the (wrong) old filename. DNSSEC inline signing will fail to sign the newly created Reverse Zone.

Perhaps, it will be cleaner/simpler to refactor the code for Reverse Zones so to require the user to enter the full name of Reverse Zones, e.g. with trailing .in-addr.arpa. or .ip6.arpa. in $zonename.

This would also facilitate first time use of Reverse Zones in pfSense as the BIND package will not auto-magically append this stuff to Reverse Zones, and behave as one would expect?


rbgarga:
Looks like changes for #215 and #216 got mixed. Could you please re-submit? You could submit both changes, with 2 different commits, on the same Pull Request

rbgarga:
Manually merged, thanks!

PR #216, opened 2016-11-21
pkg-bind: dnssec-keygen to use ECDSA instead of default RSA algo/keys

hb9cwp:
dnssec-keygen without parameters for algorithm and key sizes uses per default RSA SHA1  (alg=5) with 2048 bit KSK resp. 1024 bit ZSK.

Add parameter ```-a ECDSAP256SHA256``` to use ECDSA P245 SHA256 (alg=13) instead of defaults, following the trend and recommendations presented in http://schd.ws/hosted_files/icann562016/f5/Wander-ICANN56-DNSSEC-Adoption-v2.pdf
and
https://www.cloudflare.com/dns/dnssec/ecdsa-and-dnssec/


rbgarga:
Looks like changes for #215 and #216 got mixed. Could you please re-submit? You could submit both changes, with 2 different commits, on the same Pull Request

rbgarga:
Manually merged, thanks!

PR #218, opened 2016-11-24
syslog-ng - multiple fixes

doktornotor:
- Prevent opening edit mode on the _DEFAULT entry when clicking the "Advanced" tab; makes it impossible to do anything without modifying the URL manually in the browser.
- Allow overriding default values (Feature #4548)
- Better defaults/options for log viewer (Bug #6547)
- Fix log rotation with bzip2

### DIFF
--- a/sysutils/pfSense-pkg-syslog-ng/Makefile
+++ b/sysutils/pfSense-pkg-syslog-ng/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-syslog-ng
 PORTVERSION=	1.1.2
-PORTREVISION=	4
+PORTREVISION=	5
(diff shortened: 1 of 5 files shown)


doktornotor:
Not sure what's the business with the branches ATM, please make sure this goes back to RELENG_2_3 as well.

doktornotor:
@wbedard - well, I just installed this package for some messing with haproxy logs offloading to ELK, and got annoyed by exactly the same issue.

rbgarga:
I've reverted the zzzDEFAULT related changes and pushed a different fix

https://github.com/pfsense/FreeBSD-ports/commit/838ef792201a41c193753e5e5af8ab997d21fef8

PR #224, opened 2016-12-09
No release option '-n' added. Removed dhcp6_ctl

marjohn56:
Addition of '-n' option prevents dhcp6c sending a release signal to the ISP. By default dhcp6c sends a release signal on exit, this is not a good idea for many ISP's as the allocation will change on next connection.

Also removed dhcp6_ctl log errors and internal functions in dhcp6c as it is not used in pfSense. It may be restored by simply compiling with the #define __PFSENSE__ being removed or set to 0.

More informative logging has been added.


rbgarga:
It would be nice if we have this parameter added upstream, on FreeBSD ports tree.

rbgarga:
Closed in favor of #229

PR #229, opened 2016-12-15
Created no-release '-n' option and extra logging - 2nd commit only

marjohn56:
no-release option added to prevent a release signal being sent to the ISP. Some ISP's will release the given allocation and issue a new allocation on next connect, regardless of the DUID. As this is a default state for dhcp6c whenever it terminates it has to be prevented. The -n option prevents the signal from ever being sent.

Some LOG_DEBUG info has been changed to LOG_INFO and some extra logging added also.

The original sync of these files included the last commit, I patched this in before commencing my changes. 

Everything appears to work fine but I am unsure as to whether I should or should not have patched them in first.

I will PR the changes required to interfaces.php and interfaces.inc with a link to this PR

### DIFF
--- a/net/dhcp6/files/patch-common.h
+++ b/net/dhcp6/files/patch-common.h
@@ -1,10 +1,20 @@
 --- common.h.orig	2007-03-21 09:52:57 UTC
 +++ common.h
-@@ -179,6 +179,7 @@ extern void duidfree __P((struct duid *)
+@@ -155,7 +155,7 @@ extern char *in6addr2str __P((struct in6
+ extern int in6_addrscopebyif __P((struct in6_addr *, char *));
+ extern int in6_scope __P((struct in6_addr *));
+ extern void setloglevel __P((int));
+-extern void dprintf __P((int, const char *, const char *, ...));
++extern void d_printf __P((int, const char *, const char *, ...));
+ extern int get_duid __P((char *, struct duid *));
+ extern void dhcp6_init_options __P((struct dhcp6_optinfo *));
+ extern void dhcp6_clear_options __P((struct dhcp6_optinfo *));
+@@ -179,7 +179,7 @@ extern void duidfree __P((struct duid *)
  extern int ifaddrconf __P((ifaddrconf_cmd_t, char *, struct sockaddr_in6 *,
  			   int, int, int));
  extern int safefile __P((const char *));
+-
 +extern int opt_norelease;
- 
  /* missing */
  #ifndef HAVE_STRLCAT
+ extern size_t strlcat __P((char *, const char *, size_t));
--- a/net/dhcp6/files/patch-common.h
+++ b/net/dhcp6/files/patch-common.h
@@ -1,11 +1,10 @@
 --- common.h.orig	2007-03-21 09:52:57 UTC
 +++ common.h
-@@ -155,7 +155,7 @@ extern char *in6addr2str __P((struct in6
- extern int in6_addrscopebyif __P((struct in6_addr *, char *));
- extern int in6_scope __P((struct in6_addr *));
- extern void setloglevel __P((int));
--extern void dprintf __P((int, const char *, const char *, ...));
-+extern void d_printf __P((int, const char *, const char *, ...));
- extern int get_duid __P((char *, struct duid *));
- extern void dhcp6_init_options __P((struct dhcp6_optinfo *));
- extern void dhcp6_clear_options __P((struct dhcp6_optinfo *));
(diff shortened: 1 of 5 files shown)


net/dhcp6/files/patch-common.h
```diff
@@ -1,11 +1,10 @@
 --- common.h.orig	2007-03-21 09:52:57 UTC
 +++ common.h
-@@ -155,7 +155,7 @@ extern char *in6addr2str __P((struct in6
- extern int in6_addrscopebyif __P((struct in6_addr *, char *));
- extern int in6_scope __P((struct in6_addr *));
- extern void setloglevel __P((int));
--extern void dprintf __P((int, const char *, const char *, ...));
-+extern void d_printf __P((int, const char *, const char *, ...));
```
rbgarga: Why are you removing this change? d_printf is being used on FreeBSD ports for a reason... Are you applying your changes on top of upstream code? Or on top of FreeBSD ports code with patches applied?

marjohn56:
See my comment. I grabbed the ports as shown here:

 



Did a make on that then went from there.

rbgarga:
Run a make before start changing is not a good idea since in some cases there are post-patch: targets that does changes on files using REINPLACE_CMD. The correct workflow to produce new patches for ports is:
```
# make clean
# make extract do-patch
Change all files you want/need
# make makepatch
# make clean
```
After that you will have all patches created inside files/ subdir

rbgarga:
if you will touch files that are already patched, maybe you can use a different strategy:
```
# make clean
# make extract do-patch
# cp -r work/wide-dhcpv6-20080615 work/wide-dhcpv6-20080615.orig
Change all files inside work/wide-dhcpv6-20080615/
# cd work/wide-dhcpv6-20080615
# diff -ruN ../wide-dhcpv6-20080615.orig . > ../../files/patch-human-readable-name
# cd ../..
# make clean
```

marjohn56:
Thanks for that. I have found out the hard way :)

It's a learning curve.

marjohn56:
OK, if I do a make clean, then make extract do-patch the common.h file in work/wide... does not contain the extern d_printf; if I do a make first then it does. The existing contents of the files folder appears to put that in place. The patch is marked as 'Prevent collision with dprintf(3)'.

rbgarga:
> On 15 Dec 2016, at 13:06, marjohn56 <notifications@github.com> wrote:
> 
> OK, if I do a make clean, then make extract do-patch the common.h file in work/wide... does not contain the extern d_printf; if I do a make first then it does. The existing contents of the files folder appears to put that in place. The patch is marked as 'Prevent collision with dprintf(3)'.
> 

It’s odd, I’ve tried it here and patch is applied as you can see
```
garga@tp ~/f/p/h/n/dhcp6 ❯❯❯ make extract do-patch
===>  License BSD3CLAUSE accepted by the user
===>   dhcp6-20080615_7 depends on file: /usr/local/sbin/pkg - found
===> Fetching all distfiles required by dhcp6-20080615_7 for building
===>  Extracting for dhcp6-20080615_7
=> SHA256 Checksum OK for wide-dhcpv6-20080615.tar.gz.
===>  Applying FreeBSD patches for dhcp6-20080615_7

garga@tp ~/f/p/h/n/dhcp6 ❯❯❯ find . -name common.h
./work/wide-dhcpv6-20080615/common.h

garga@tp ~/f/p/h/n/dhcp6 ❯❯❯ grep d_printf ./work/wide-dhcpv6-20080615/common.h
extern void d_printf __P((int, const char *, const char *, ...));
```

marjohn56:
Yep, I was devel, should have been on master!

rbgarga:
> On 15 Dec 2016, at 13:52, marjohn56 <notifications@github.com> wrote:
> 
> Yep, I was devel, should have been on master!

No, master is a read-only copy of FreeBSD ports tree, devel is the correct branch you should use since it contains all pfSense changes

marjohn56:
Sorted...

The fundamental mistake I was making was deleting the *.orig file, copying and renaming the new source to *.orig and then editing the *.c or *.h file. All I needed to do was edit the *.c or *.h and then makepatch.

As I said earlier, a learning curve. I've issued a new commit. Hopefully it's now all correct!

Thank you for your help and guidance, extremely useful.

rbgarga:
@marjohn56 cool! Will you open a new PR?

marjohn56:
#230 

Do you want me to issue a new PR for the *.inc and *.php stuff too?

I did the changes online for those earlier.

PR #231, opened 2016-12-15
dhcp6c no release option & extra light info logging

marjohn56:
Add a no release option '-n'. This prevents a release signal from being sent to the ISP causing a new PD or address to be allocated.
Added some extra light INFO logging which is useful to see the communications between dhcp6c and the ISP

### DIFF
--- a/net/dhcp6/files/patch-common.h
+++ b/net/dhcp6/files/patch-common.h
@@ -9,3 +9,12 @@
  extern int get_duid __P((char *, struct duid *));
  extern void dhcp6_init_options __P((struct dhcp6_optinfo *));
  extern void dhcp6_clear_options __P((struct dhcp6_optinfo *));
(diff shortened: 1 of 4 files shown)


rbgarga:
It would be nice to hear @hrs-allbsd thoughts about this change. Maybe it could end up in FreeBSD ports tree instead of only pfSense. What do you think Hiroki?

hrs-allbsd:
Looks good to me.  I integrated the patch (+ manual page update) into my github repo and I will update net/dhcp6 in the FreeBSD ports tree.

rbgarga:
@hrs-allbsd just to plan what we should do here in pfSense. Do you plan to commit it soon to FreeBSD?

hrs-allbsd:
Yes.  Committed the change into net/dhcp6 at r428955 just now.

rbgarga:
I've cherry-picked FreeBSD/FreeBSD-ports commit that brought new version to our tree. Thanks!

PR #233, opened 2016-12-17
Allow quagga ospf stub, not so stub and totally stub areas (Feature #3303)

doktornotor:
Slightly modified patch from https://redmine.pfsense.org/issues/3303 to handle the checkbox-based configuration in old versions of the package. (I'd rather avoid renaming the badly chosen config variable name and upgrading the config on upgrade.)

While here, also tweaked the package description (conflicts should be handled by pkg now) and removed some copy-paste junk from privs.

### DIFF
--- a/net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.xml
+++ b/net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.xml
@@ -117,8 +117,15 @@
 		<field>
 			<fielddescr>Disable FIB updates (Routing table)</fielddescr>
 			<fieldname>updatefib</fieldname>
-			<description>Disables the updating of the host routing table(turns into stub router).</description>
-			<type>checkbox</type>
+			<description>Disables the updating of the host routing table (turns into stub router).</description>
+			<type>select</type>
+			<default_value>none</default_value>
+			<options>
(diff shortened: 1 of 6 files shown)


net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.xml
```diff
@@ -117,8 +117,15 @@
...
 			<fieldname>updatefib</fieldname>
-			<description>Disables the updating of the host routing table(turns into stub router).</description>
-			<type>checkbox</type>
+			<description>Disables the updating of the host routing table (turns into stub router).</description>
+			<type>select</type>
+			<default_value>none</default_value>
+			<options>
+				<option><name>None (FIB updates enabled) (default)</name><value>none</value></option>
+				<option><name>Stub Area (stub)</name><value>stub</value></option>
+				<option><name>Totally Stub (no-summary)</name><value>stub no-summary</value></option>
```
rbgarga: Shouldn't value be "nosum" here? Based on check done in https://github.com/pfsense/FreeBSD-ports/pull/233/files#diff-238d7552fa57444efc0f66cb1bd2b8d3R191

doktornotor: Thanks, good catch.

PR #234, opened 2016-12-17
Disable Auto Config Backup feature (Feature #6951), add some tweaks for savemsg (Bug #6950)

doktornotor:
Make it possible to disable the package without uninstalling (e.g. to prevent timeout delays when there's no WAN connection). Also, it is now possible to delete the stored credentials from config.xml. 

Some tweaks to avoid completely bogus "success" message for Bug #6950 while here. Now it should at least report failure when the write_config() failed due to the user having user-config-readonly privilege, instead of always claiming it worked.

### DIFF
--- a/sysutils/pfSense-pkg-AutoConfigBackup/files/usr/local/pkg/autoconfigbackup.inc
+++ b/sysutils/pfSense-pkg-AutoConfigBackup/files/usr/local/pkg/autoconfigbackup.inc
@@ -3,7 +3,7 @@
  * autoconfigbackup.inc
  *
  * part of pfSense (https://www.pfsense.org)
- * Copyright (c) 2008-2015 Rubicon Communications, LLC (Netgate)
+ * Copyright (c) 2008-2016 Rubicon Communications, LLC (Netgate)
  * All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -123,13 +123,7 @@ function test_connection($post) {
 	global $savemsg, $config, $g;
 
 	// Do nothing when booting or when not enabled
-	if (function_exists("platform_booting")) {
-		if (platform_booting()) {
-			return;
-		}
-	} elseif ($g['booting']) {
-		return;
-	} elseif (!acb_enabled()) {
+	if (platform_booting() || !acb_enabled()) {
 		return;
 	}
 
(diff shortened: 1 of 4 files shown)


sysutils/pfSense-pkg-AutoConfigBackup/files/usr/local/pkg/autoconfigbackup.inc
```diff
@@ -96,13 +122,15 @@ function configure_proxy() {
...
 
-	// Do nothing when booting
+	// Do nothing when booting or when not enabled
 	if (function_exists("platform_booting")) {
 		if (platform_booting()) {
 			return;
 		}
 	} elseif ($g['booting']) {
 		return;
+	} elseif (!acb_enabled()) {
```
jim-p: This could be combined with the conditional above it.

sysutils/pfSense-pkg-AutoConfigBackup/files/usr/local/pkg/autoconfigbackup.inc
```diff
@@ -150,13 +178,15 @@ function test_connection($post) {
...
```
jim-p: This could be combined with the conditional above it.

doktornotor: @jim-p - Before redoing it. Is the first check about function_exists() still needed here? Was a 2.2.x compatibility thing IIRC.

jim-p: That can go. That function was added on 2.2, and this code wouldn't ever touch anything that old so it can be cleaned up.

doktornotor:
Hmmm, well... Given my experience with completely retared isset() behaviour in PHP, I'd frankly leave the check well alone since there's nothing wrong with that and it completely avoids stupid issues such as this (as the very recent example): https://github.com/pfsense/FreeBSD-ports/pull/232/).  

Perhaps someone has a way to upgrade people to default it to on if defaulting it to on in the XML does not do that. PHP's isset() is broken by design.

jim-p:
Maybe instead of a checkbox, use a drop-down box with Enabled/Disabled and then the value could explicitly default to enabled _and_ the disabled value would be spelled out as well. Then unset/empty could be considered equal to enabled.

jim-p:
That works great! It assumed enabled and acted as expected when disabled/manually enabled. It's also more obvious to the user this way.

doktornotor:
@jim-p - thanks for testing/help.

Is there a ticket for:

- setting up defaults on package install
- setting up defaults on package upgrade
- renaming vars on upgrade?

Seriously needed. This hacking around sucks.

jim-p:
None that I'm aware of. We don't have any kind of special code path to handle upgrades for packages that I know of, they've been left to fend for themselves individually or maintain backward compatibility. It could use improvement, surely, but it might take some more in-depth design thought on how best to handle it all.

doktornotor:
Indeed, still I'd say it'd be worth it, looking at the amount of cruft that accumulated in packages like Squid is kinda sad.

PR #238, opened 2016-12-20
openvpn-client-export, fix, show local-users and their certificates also when a user above them has their certificate revoked

PiBa-NL:
openvpn-client-export, fix, show local-users and their certificates also when a user above them has their certificate revoked


rbgarga:
@PiBa-NL please rebase your fork, it's conflicting

sebastienbo:
I like the ovpn tool to export configuration but why isn't there the oppposite?

A way to import an ovpn file?

It would make ovpn much easier to use, export from one pfsense and import in the other one

jim-p:
That is a much more difficult tool to write, and it's irrelevant to this PR. If you wish to start a discussion, start a thread on the forum.

PR #240, opened 2016-12-22
Status / Monitoring - Graph Refresh Enhancement

NOYB:
Immediately apply refresh interval setting change.
Option for settings changes to dynamically refresh graphs.

### DIFF
--- a/sysutils/pfSense-Status_Monitoring/files/usr/local/www/status_monitoring.php
+++ b/sysutils/pfSense-Status_Monitoring/files/usr/local/www/status_monitoring.php
@@ -615,8 +615,9 @@ display_top_tabs($tab_array);
 					<span class="help-block">Inverse</span>
 				</div>
 				<div class="col-sm-2">
-					<select class="form-control" data-toggle="tooltip" data-trigger="hover" data-placement="top" title="You must save this view for the refresh interval to take effect." id="refresh-interval" name="refresh-interval">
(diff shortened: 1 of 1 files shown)


jdillard:
I don't think this is worth the complexity that it adds when having to support it during potential future changes. It also doesn't account for the case of the user wanting to refresh on settings change and also wants to refresh on an interval.

NOYB:
Simply refactors the existing little "scheduling" in-line code blurb into a function so it can be called by a handler.
The only additional code is the handler and it is not hardly complex.  Supporting it during potential future changes is a nit, as potential future changes should be including this capability anyway.  In which case it would not be additional complexity.

As for not accounting for the case of user wanting to refresh on settings change while an interval is selected.  The existing code doesn't account for refresh on settings change at all.  For any of the settings.  So not being part of the timed refresh interval settings really is moot.  Even so a simple expression in the if statement can easily provide that.

Not only that but the user can't even make a refresh interval change at all without saving a view and reloading the entire page.  Overly cumbersome just to turn on/off or even just change the refresh interval of a graph being viewed.

This provides 2 quite valuable features with nearly no new code and very infinitesimal complexity.
1) Ability to enable/disable/change refresh interval dynamically in real-time (on the fly).  Without having to save the graph as a view and reload the entire page.
2) Ability to dynamically change graph settings in real-time (on the fly).  Without having to reload the page.

These features are quite useful.  Especially dynamic graph update on settings change.  Convenient for looking though graphs with different time periods, resolutions, graph types, etc.  Others would probably find them useful too if available to them.

PR #241, opened 2016-12-22
squid - Remove NT Domain auth (Bug #7017), allow some configuration for DH and cipher suites (Feature #6593), fix EDH and EECDH cipher suites (Bug #6592), some cosmetics

doktornotor:
1. Remove NT Domain authentication. Dead deprecated junk. This has been broken ever since 2.3 release (https://redmine.pfsense.org/issues/7017) without anyone noticing. Just use LDAP for AD authentication.

2. Added Modern and Intermediate options for cipher suites 

- settings per http://wiki.squid-cache.org/ConfigExamples/Intercept/SslBumpExplicit#Modern_DH.2FEDH_ciphers_usage docs
- modern disables TLS v1.0 as well (matching the behavior of reverse proxy)
- intermediate enables the HIGH cipher suites for compatibility
- add DH key size selection (defaults to 2048 for security)

3. Fixed EDH and EECDH cipher suites usage (Bug #6592)

4. Cosmetics
- Fix "Update AV" button handling
- Remove confusing note about CP patch
- Handle IPv6 IPs in "User Defined Reverse Proxy IPs" properly - requested and tested by a user @ https://forum.pfsense.org/index.php?topic=123077.0 
- Nuked some comments cruft remaining from bootstrap conversion while here.

### DIFF
--- a/www/pfSense-pkg-squid/Makefile
+++ b/www/pfSense-pkg-squid/Makefile
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-squid
-PORTVERSION=	0.4.28
+PORTVERSION=	0.4.29
 CATEGORIES=	www
 MASTER_SITES=	# empty
 DISTFILES=	# empty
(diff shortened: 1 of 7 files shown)


doktornotor:
Should be complete now, other fixed need to wait for next round.

PR #242, opened 2016-12-28
Fix SslBump/MITM features for Squid 3.5.x (Bug #6527)

doktornotor:
(Also improved the compatibility mode description while here.)


rbgarga:
Please leave a message when it's ready for review

doktornotor:
@rbgarga - well I don't plan any changes there unless someone complains about something being more broken than it is already. Not using any of this evil MITM stuff myself so I thought someone who's making use of those would be able to spot issues more reliably/easily. 

So far the forum test request has collected ~60 views and no response. Funny, considering the flood of questions and complaints about MITM on the forum.

I'd leave this open until new year and then just go ahead.

doktornotor:
@rbgarga - Go ahead, please, 120 views and zero response on the forum, people apparently don't care.

(Also improved the compatibility mode description while here.)

rbgarga:
It looks OK to me, lets just wait for @jim-p review, just to have an extra pair of eyes

doktornotor:
Yeah, no hurry, been broken for quite some time (plus, the entire idea is sort of broken by design :D)

rbgarga:
Merging it to devel and RELENG_2_3 to give users a change to test and after that we can cherry-pick it to RELENG_2_3_2

rbgarga:
@doktornotor I was thinking about merge it to 2.3.2, did you hear back from users?

doktornotor:
@rbgarga Unfortunately not... It's be nice to get all the stuff merged including the latest GUI updates after it gets a couple of days of testing to prevent confusion with very different versions across branches.

tadaog:
doktornotor:

Thks for the patch. 
The current stage of squid.inc applies splice to whitelist, and I can confirm that any entry here doesn't interfere with Squidguard blocking by deny categories.

doktornotor:
@tadaog - Thanks for feedback and testing.

PR #243, opened 2016-12-29
haproxy, sync over all changes from haproxy-devel v0.52

PiBa-NL:


rbgarga:
Please rebase your fork to make it mergeable again

net/pfSense-pkg-haproxy/files/usr/local/pkg/haproxy/haproxy.inc
```diff
@@ -525,15 +532,16 @@ function haproxy_custom_php_deinstall_command() {
...
 
 function haproxy_custom_php_install_command() {
-	global $g, $config, $static_output;
+	global $g, $static_output;
 	$static_output .= "HAProxy, running haproxy_custom_php_install_command()\n";
 	update_output_window($static_output);
 	
 	$pf_version=substr(trim(file_get_contents("/etc/version")),0,3);
-	if ($pf_version == "2.1" || $pf_version == "2.2")
+	if ($pf_version == "2.1" || $pf_version == "2.2") {
```
rbgarga: This code will never reach 2.1 or 2.2 systems, so this check can be removed

net/pfSense-pkg-haproxy/files/usr/local/pkg/haproxy/haproxy.inc
```diff
@@ -2079,22 +2092,27 @@ function load_ipfw_rules() {
...
 	
-	if (haproxy_utils::$pf_version < 2.2) {
-		// pfSense 2.1 FreeBSD 8.3
-		mwexec("/usr/local/sbin/ipfw_context -a $ipfw_zone_haproxy", true);
-		
-		foreach($transparent_interfaces as $transparent_if => $value) {
-			mwexec("/usr/local/sbin/ipfw_context -a $ipfw_zone_haproxy -n $transparent_if", true);
+	if (haproxy_utils::$pf_version < 2.4) {
+		$rulenum = 64000; // why that high? captiveportal.inc also does it...
+		if (haproxy_utils::$pf_version < 2.2) {
```
rbgarga: This conditional for old version also can be removed

net/pfSense-pkg-haproxy/files/usr/local/pkg/haproxy/haproxy.inc
```diff
@@ -2079,22 +2092,27 @@ function load_ipfw_rules() {
...
+			}
 		}
 	} else {
-		// pfSense 2.2 FreeBSD 10
-		mwexec("/sbin/ipfw zone $ipfw_zone_haproxy create", true);
-		foreach($transparent_interfaces as $transparent_if => $value) {
-			mwexec("/sbin/ipfw zone $ipfw_zone_haproxy madd $transparent_if", true);
-		}
+		// pfSense 2.4, captive portal rules start at 1000
+		$rulenum = 10;
```
rbgarga: 1000 or 10?

net/pfSense-pkg-haproxy/files/usr/local/pkg/haproxy/haproxy.inc
```diff
@@ -2109,9 +2127,14 @@ function load_ipfw_rules() {
 	
 
 	file_put_contents("{$g['tmp_path']}/ipfw_{$ipfw_zone_haproxy}.haproxy.rules", $rules);
-	if (haproxy_utils::$pf_version < 2.2)
+	if (haproxy_utils::$pf_version < 2.2) {
```
rbgarga: Can be removed too

net/pfSense-pkg-haproxy/files/usr/local/pkg/haproxy/haproxy.inc
```diff
@@ -2157,18 +2180,36 @@ function haproxy_plugin_certificates($pluginparams) {
 				}
 			}
 		}
+		
+		$a_backends = $config['installedpackages']['haproxy']['ha_pools']['item'];
+		if (is_array($a_backends)) {
+			foreach ($a_backends as &$backend) {
```
rbgarga: Why use a reference here (&$backend) if this variable is not changed inside the loop?

net/pfSense-pkg-haproxy/files/usr/local/www/haproxy/haproxy_listeners.php
```diff
@@ -149,9 +149,10 @@ function array_moveitemsbefore(&$items, $before, $selected) {
 	}
 } else {
 	$result = haproxy_check_config($retval);
-	if ($result)
+	if ($result) {
 		$savemsg = gettext($result);
 }
```
rbgarga: Fix indent

PiBa-NL:
Most of above comments have been addressed in: https://github.com/pfsense/FreeBSD-ports/pull/245

net/pfSense-pkg-haproxy/files/usr/local/pkg/haproxy/haproxy.inc
PiBa-NL: 10 is right i think for haproxy, as 'zones' no longer separate stuff in ipfw contexts, and haproxy needs to catch reply traffic it shouldn't use the same rule numbers as CP which does afaik start with rule 1000, putting the rules with for haproxy with a lower number seems to work properly.

PR #247, opened 2017-01-03
Fixup RRD graphing enabled/disabled message

phil-davis:
enable_rrd_graphing() does not actually return a status. It "just works".
Better to just give a specific message (enabled or disabled) rather than make an irrelevant attempt to use get_std_save_message()


phil-davis:
Bumped PORTREVISION because this is just a fix to a bit of front-end UI feedback behavior - nothing that effects any sort of critical functionality.

PR #249, opened 2017-01-07
Squid - RELENG_2_3 sync up

doktornotor:
Fix description to match the one in devel branch.

### DIFF
--- a/www/pfSense-pkg-squid/files/usr/local/pkg/squid.xml
+++ b/www/pfSense-pkg-squid/files/usr/local/pkg/squid.xml
@@ -359,11 +359,12 @@
 			<description>
 				<![CDATA[
 				The compatibility mode determines which cipher suites and TLS versions are supported.
-				<span class="text-info">Default: Modern.</span> Intended for modern clients only (post FF 27, Chrome 22, IE 11 etc.)
+				<span class="text-info">Default: Modern.</span> Click Info for details.
 				<div class="infoblock">
 				Modern mode also disables HIGH ciphers and TLS v1.0.<br/>
-				<strong><span class="text-info">Hint:</span></strong> If you need to support older clients, use the Intermediate setting.<br/><br/>
-				<strong><span class="errmsg">Warning: </span>Clients like IE 6 and Java 6 are not supported anymore!</strong>
+				<strong><span class="text-info">Hint:</span></strong> If you get an error (such as "This proxy and the remote host failed to negotiate a mutually acceptable security settings for handling your request")<br/>
+				when accessing some websites, you will need to use the Intermediate setting.<br/><br/>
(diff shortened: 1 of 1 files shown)


doktornotor:
@rbgarga - This one got somehow out of sync since 0.4.30, no idea how.

doktornotor:
@rbgarga - I am getting completely confused about what's up with the sync lately? Looking at RELENG_2_3, ```squid_antivirus.inc``` has been last synced a year ago. Ditto with RELENG_2_3_2.  So, it's missing fixes like the ClamAV DB path on embedded (https://github.com/pfsense/FreeBSD-ports/commit/38872566f0001d804f33c3985ca28b199c49049c)

Overall, this is *extremely* confusing for users and pretty much everyone. Fixes doing to /dev/null, essentially. Noone's getting them.

I am mentioning this one specifically b/c of https://forum.pfsense.org/index.php?topic=123701.msg682969#msg682969 - but with state like this, it's basically impossible to keep track of things and provide support for users. There needs to be a way for fixes/package updates to pretty much automatically bubble through to the release versions of pfSense. Fixing packages for a bunch of devel snapshots users is a waste of time. :(

rbgarga:
I believe this is the only change missing on squid, few days ago I sync'd from RELENG_2_3 -> RELENG_2_3_2 but this specific change seems to be devel only. I would like to hear from @jim-p if it was something he planned to leave on 2.4 only or if we can sync it to 2.3.x

doktornotor:
@rbgarga - Yeah I noticed the RELENG_2_3_2 sync (thanks, BTW ;) and then got completely confused by this one. I commented on the Redmine ticket as well. Keeping track of these things must be a PITA for you, definitely a PITA for package maintainers, plus users cannot make sense of it, they get the tickets closed but basically don't see any fixed packages in most cases.

Getting back something along the line of the abandoned pfsense-packages repo, independent from the rest of ports included in FreeBSD, would be kinda neat. In general, most of the updates/fixes in pfsense-pkg-* do not require any updates for the underlying ports. What I figured out is that users could get the fixes much faster and more easily if there was a way to select either stable or development branch for packages, not just for pfSense itself. (Also, would significantly reduce the risk of breaking things badly when testing things.) 

Not sure how feasible this would be. Perhaps lets move this to the forums for some ideas?

doktornotor:
@jim-p Yeah, I'm pretty sure there was no hidden agenda behind this, lol. I raised this question b/c - as you've noticed with the OpenVPN 2.4 case - it's damn hard to get feedback from users. I'd hope that having a sort of "testing"/"development" branch for packages *only* with a toggle  could make the testing easier/better somehow. (I had no luck with getting people test the SSL/MITM stuff in Squid either. And OpenVPN/Squid are pretty widely used things. Seems like people definitely avoid using any sort of snapshots stuff I guess.)

PR #254, opened 2017-01-14
Squid - improve ClamAV DB updates cronjob handling

doktornotor:
Make the minutes random (avoids some mirrors overload issues) and the update frequency more obvious.


doktornotor:
@jim-p @rbgarga 

Please, revert this everywhere. It broke saving General tab. NFC why. https://forum.pfsense.org/index.php?topic=124218.0

PR #258, opened 2017-01-17
Added new flag '-x' prevents script execution

marjohn56:
In pfsense, dhcp6c will launch the wan script on receipt of a reply from the dhcp6 server. When in dhcp6withoutRA mode, that same script will also get called on receipt of the RA signal by rtsold. This is not an optimal situation with the same script running twice at the same time. The additional flag prevents dhcp6c from running the script only when it receives a REPLY signal from the server. As the REPLY signal is only sent when replying to a REQUEST, and as the REQUEST is only sent at the start of a session then any other trigger to run the script will execute it as normal.

### DIFF
--- a/net/dhcp6/files/patch-dhcp6c.c
+++ b/net/dhcp6/files/patch-dhcp6c.c
@@ -1,0 +1,97 @@
+--- dhcp6c.c.orig	2016-12-19 08:16:42 UTC
++++ dhcp6c.c
+@@ -109,6 +109,7 @@ static int ctldigestlen;
+ static int infreq_mode = 0;
+ 
+ int opt_norelease;
++int opt_noscript;
+ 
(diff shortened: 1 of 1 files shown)


rbgarga:
Hello @marjohn56. It would be nice if you submit this change upstream, what would make us to carry less patches around. Could you submit it to https://github.com/hrs-allbsd/wide-dhcpv6, on branch called 'freebsd'?

marjohn56:
Happy to do that, but as it’s pfSense specific would they have objections?

rbgarga:
I don't see it as pfSense specific, IMO it can be useful for other users

marjohn56:
OK, submitted.

rbgarga:
Waiting the discussion at https://github.com/hrs-allbsd/wide-dhcpv6/pull/2

marjohn56:
This is no longer required, alternative solution found.

PR #260, opened 2017-01-20
pfSense-pkg-suricata-3.1.2 0 -- Pass List implementation fix

bmeeks8:
This update to the Suricata GUI package corrects issues with implementation of default Pass Lists when running Suricata with inline IPS Mode enabled.  A couple of other minor bugs are also fixed.

**Bug Fixes**

1. Remove automatic inclusion of all locally-attached networks in the default Pass List when using IPS mode.  This had the unintended side-effect of essentially whitelisting all traffic to and from local hosts.

2. Added the capability to completely disable use of the default Pass List when running with inline IPS Mode.  Formerly, the default Pass List would be used when no custom list was specified.

3. Remove automatic inclusion of the WAN interface IP address in the default Pass List when using IPS mode.  This had the unintended side-effect of essentially whitelist all inbound NAT traffic because the destination IP would be the WAN IP.

4. Increase default value of Host Memcap on IP REPUTATION tab to 32 MB as most IP lists today are quite large.  This is effective only for newly created interfaces.

5. The checkbox for including/excluding the WAN IP from a custom Pass List was inadvertently removed during the Bootstrap conversion of the GUI code.  This checkbox is now restored.

### DIFF
--- a/security/pfSense-pkg-suricata/Makefile
+++ b/security/pfSense-pkg-suricata/Makefile
@@ -2,6 +2,7 @@
 
 PORTNAME=	pfSense-pkg-suricata
 PORTVERSION=	3.1.2
+PORTREVISION=	0
 CATEGORIES=	security
 MASTER_SITES=	# empty
 DISTFILES=	# empty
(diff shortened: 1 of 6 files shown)


bmeeks8:
The title of this pull request should be "pfSense-pkg-suricata-3.1.2_0", but the underscore to denote the Port Revision is missing and I can't figure out how to edit it.

Bill

PR #262, opened 2017-01-20
pfSense-pkg-tftpd package improvements

doktornotor:
- Add option to bind to a single IP
- Add option to force IPv4 only
- Add option to specify the maximum permitted block size
- Re-add backup/restore/upload/delete functionality
- Remove useless custom_add_php_command
- Remove unneeded includes
- Use a PID file for the service
- Remove useless filter_configure() call
- Nuke the shipped startup script on install
- Fix custom_php_{install,deinstall}_command

### DIFF
--- a/ftp/pfSense-pkg-tftpd/files/usr/local/pkg/tftpd.inc
+++ b/ftp/pfSense-pkg-tftpd/files/usr/local/pkg/tftpd.inc
@@ -3,6 +3,7 @@
  * tftpd.inc
  *
  * part of pfSense (https://www.pfsense.org)
+ * Copyright (c) 2015-2017 Rubicon Communications, LLC (Netgate)
  * Copyright (c) 2016 Stefan Seidel
  * All rights reserved.
  *
@@ -19,17 +20,14 @@
  * limitations under the License.
  */
 
-if (!function_exists("filter_configure")) {
-	require_once("filter.inc");
-}
 require_once("globals.inc");
-require_once("interfaces.inc");
 require_once("pfsense-utils.inc");
 require_once("service-utils.inc");
 require_once("util.inc");
 
 function install_package_tftpd() {
 	safe_mkdir("/tftpboot");
+	unlink_if_exists("/usr/local/etc/rc.d/tftpd");
 }
 
 function deinstall_package_tftpd() {
@@ -37,7 +35,7 @@ function deinstall_package_tftpd() {
 }
 
 function sync_package_tftpd() {
-	global $config;
+	global $g, $config;
 
 	conf_mount_rw();
 
@@ -56,17 +54,35 @@ function sync_package_tftpd() {
 		unlink_if_exists('/usr/local/etc/rc.d/tftpd.sh');
 		return;
 	}
-	
+
+	// Root directory
 	$datadir = $tftpd_conf['datadir'];
-	
-	if (!is_dir($datadir)) {
-		log_error("TFTP files directory {$datadir} does not exist.");
-		return;
-	} 
+
+	// TFTP Server Bind IP
+	if (!empty($tftpd_conf['tftpd_ip'])) {
+		$address = $tftpd_conf['tftpd_ip'];
+		if (is_ipaddrv6($address)) {
+			$address = "-a [{$address}]";
+		} else {
+			$address = "-a {$address}";
+		}
+	}
+
+	$pidfile = "{$g['varrun_path']}/tftpd-hpa.pid";
+
+	// IPv4 Only?
+	if ($tftpd_conf['tftpd_ipv4only'] == "on") {
+		$options = "-4";
+	}
+
+	// Max Block Size
+	if (!empty($tftpd_conf['tftpd_blocksize'])) {
+		$options .= " -B {$tftpd_conf['tftpd_blocksize']}";
+	}
 
 	write_rcfile(array(
 		"file" => "tftpd.sh",
-		"start" => "/usr/local/libexec/in.tftpd -l -s {$datadir}",
+		"start" => "/usr/local/libexec/in.tftpd -l -s {$datadir} {$address} -P {$pidfile} {$options}",
 		"stop" => "/usr/bin/killall in.tftpd"
 		)
 	);
@@ -82,13 +98,28 @@ function sync_package_tftpd() {
 	}
 
 	conf_mount_ro();
-
-	filter_configure();
 }
 
 function validate_form_tftpd($post, &$input_errors) {
 	if ($post['datadir'] && !is_dir($post['datadir'])) {
 		$input_errors[] = 'Directory for files does not exist!';
+	}
+
+	if ($post['tftpd_ip']) {
+		if ($post['tftpd_ipv4only'] && !is_ipaddrv4($post['tftpd_ip'])) {
+			$input_errors[] = 'TFTP Server Bind IP must be a valid IPv4 address!';
+		} elseif (!is_ipaddr($post['tftpd_ip'])) {
+			$input_errors[] = 'TFTP Server Bind IP must be a valid IP address!';
+		}
+		if (!is_ipaddr_configured($post['tftpd_ip'])) {
+			$input_errors[] = "{$post['tftpd_ip']} TFTP Server Bind IP must be a valid, locally configured IP address!";
+		}
+	}
+
+	if ($post['tftpd_blocksize']) {
+		if (!is_numericint($post['tftpd_blocksize']) || ($post['tftpd_blocksize'] < 512) || ($post['tftpd_blocksize'] > 65464)) {
+			$input_errors[] = 'Max Block Size must be an integer with a permitted range from 512 to 65464!';
+		}
 	}
 }
 
--- a/ftp/pfSense-pkg-tftpd/files/usr/local/pkg/tftpd.inc
+++ b/ftp/pfSense-pkg-tftpd/files/usr/local/pkg/tftpd.inc
@@ -21,28 +21,25 @@
  */
 
 require_once("globals.inc");
+require_once("notices.inc");
 require_once("pfsense-utils.inc");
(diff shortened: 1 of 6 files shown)


ftp/pfSense-pkg-tftpd/files/usr/local/www/tftp_files.php
```diff
@@ -0,0 +1,230 @@
...
+				<th>Last Modified</th>
+				<th>Size</th>
+				<th>Actions</th>
+			</thead>
+			<tbody>
+			<?php $tftpdfiles = new RecursiveDirectoryIterator("{$files_dir}"); ?>
+			<?php foreach (new RecursiveIteratorIterator($tftpdfiles) as $filename => $file): ?>
+			<?php if (is_file($file)): ?>
+				<tr>
+					<td><?=gettext($file); ?></td>
```
jim-p: Why `gettext()` on a filename? Shouldn't we display that exactly as it is?
Also, probably a good idea to use `htmlspecialchars()` before displaying that filename to the browser.

ftp/pfSense-pkg-tftpd/files/usr/local/www/tftp_files.php
```diff
@@ -0,0 +1,230 @@
...
+					<td><?=date('M-d Y g:i a', filemtime("{$file}")); ?></td>
+					<td><?=tftp_byte_convert(filesize("{$file}")); ?> </td>
+						<td>
+						<a name="tftpd_deleteX[]" id="tftpd_deleteX[]" type="button" title="<?=gettext('Delete this file');?>"
+							href='?type=tftp&amp;act=del&amp;filename=<?=$file;?>' style="cursor: pointer;" text="delete this">
```
jim-p: Should use `htmlspecialchars()` on the filename in the URL

ftp/pfSense-pkg-tftpd/files/usr/local/www/tftp_files.php
```diff
@@ -0,0 +1,230 @@
...
+							<i class="fa fa-trash" title="<?=gettext('Delete this file');?>"></i>
+						</a>
+						<a name="tftpd_dnloadX[]" id="tftpd_dnloadX[]" type="button" title="<?=gettext('Download this file');?>"
+							href='tftp_files.php?a=download&amp;filename=<?=$file;?>' style="cursor: pointer;">
```
jim-p: Should use `htmlspecialchars()` on the filename in the URL

ftp/pfSense-pkg-tftpd/files/usr/local/www/tftp_files.php
```diff
@@ -0,0 +1,230 @@
...
+		<button data-toggle="modal" data-target="#uploader" role="button" aria-expanded="false" type="button" name="tftpd_upload" id="tftpd_upload" class="btn btn-info btn-sm" title="<?=gettext('Upload files');?>">
+			<i class="fa fa-upload icon-embed-btn"></i>
+			<?=gettext("Upload")?>
+		</button>
+		<a name="tftpd_dnload_all" id="tftpd_dnload_all" type="button" class="btn btn-info btn-sm" title="<?=gettext('Download all files in a single gzip archive');?>"
+			href="tftp_files.php?a=download&amp;t=backup" text="download all files">
+			<i class="fa fa-download icon-embed-btn"></i>
+			<?=gettext('Download');?>
+		</a>
+		<a name="tftpd_backup" id="tftpd_backup" type="button" class="btn btn-success btn-sm" title="<?=sprintf(gettext('Backup all files to %s'), $backup_path);?>"
```
jim-p: Should use `htmlspecialchars()` on the variable here.

ftp/pfSense-pkg-tftpd/files/usr/local/www/tftp_files.php
```diff
@@ -0,0 +1,230 @@
...
+			href="tftp_files.php?a=other&amp;t=backup" text="backup files">
+			<i class="fa fa-save icon-embed-btn"></i>
+			<?=gettext('Backup');?>
+		</a>
+		<?php if (file_exists($backup_path)): ?>
+		<a name="tftpd_restore" id="tftpd_restore" type="button" class="btn btn-danger btn-sm" title="<?=sprintf(gettext('Restore all files from %s'), $backup_path);?>"
```
jim-p: Should use `htmlspecialchars()` on the variable here.

ftp/pfSense-pkg-tftpd/files/usr/local/pkg/tftpd.inc
```diff
@@ -19,25 +20,60 @@
...
 function deinstall_package_tftpd() {
-	@rmdir("/tftpboot");
+	if (is_array($config['installedpackages']['tftpd'])) {
+		$tftpd_conf = &$config['installedpackages']['tftpd']['config'][0];
+	} else {
+		$tftpd_conf = array();
+	}
+	$datadir = ($tftpd_conf['datadir'] ?: '/tftpboot');
+	// Will only get removed when empty
+	@rmdir("{$datadir}");
```
jim-p: This definitely needs a sanity check to make sure the user didn't do something dumb like set it to `/usr/local/www/` or `/etc/`. I see input validation prevents `/` but even so this may be too dangerous to `rmdir`. It's probably better to just leave the data in place.

doktornotor: As noted there, the only thing it removed are **empty** dirs. http://php.net/manual/en/function.rmdir.php

ftp/pfSense-pkg-tftpd/files/usr/local/pkg/tftpd.inc
```diff
@@ -19,25 +20,60 @@
...
-	require_once("filter.inc");
-}
 require_once("globals.inc");
-require_once("interfaces.inc");
 require_once("pfsense-utils.inc");
 require_once("service-utils.inc");
 require_once("util.inc");
 
+/* Helper function for files listing */
+function tftp_byte_convert($bytes) {
```
doktornotor: No idea, no clue that it existed.

doktornotor:
@jim-p - Most of the "variable" things you are requesting to escapeshellarg() are in fact hardcoded in the PHP. Most of the rest is treated by input validation. Whatever...

But before doing any other changes here - after looking at this again, I'd like to remove the datadir setting altogether and hardcoded /tftpboot there. I cannot see any benefit here leaving it configurable, just adds a code bloat and potential issues with idiotic settings.

ftp/pfSense-pkg-tftpd/files/usr/local/pkg/tftpd.inc
```diff
@@ -19,25 +20,60 @@
...
+	$e = floor(log($bytes, $convention));
+	return round($bytes/pow($convention, $e), 2) . ' ' . $s[$e];
+}
+
+/* Create backup of the TFTP server directory */
+function tftp_create_backup($trigger_download = false) {
+	global $backup_dir, $backup_path, $files_dir;
+
+	conf_mount_rw();
+	safe_mkdir("{$backup_dir}");
```
doktornotor: backup_dir is hardcoded to /root/backup in ```tftp_files.php```

ftp/pfSense-pkg-tftpd/files/usr/local/pkg/tftpd.inc
jim-p: True but if it's empty, why bother?

jim-p:
That may be, but it's the best practice to avoid potential future changes causing a problem. I'd rather do it correctly now and make sure there is never a problem than have one creep in later. And it's really simple to do the extra encoding now so there's little reason to skip it.

doktornotor:
OK. Any objections on hardcoding ```/tftpboot``` here?

ftp/pfSense-pkg-tftpd/files/usr/local/www/tftp_files.php
doktornotor: Nah. That code is a copy-paste from suricata package, pretty much. ;)

ftp/pfSense-pkg-tftpd/files/usr/local/pkg/tftpd.inc
doktornotor: Well to not leave lingering empty dir there (some people like to not leave cruft behind). Can remove it altogether of course, I have no preference here.

jim-p:
I can see how it's convenient to move it around, perhaps someone wants to put it under a specific user's home dir or whatnot, but those same issues could be solved via manual directory permissions and symlinks. I'd prefer to see it under `/usr/local` somewhere rather than in `/` but people are used to seeing it under `/tftpboot` so it may be too late to force that change if it can't be customized.

ftp/pfSense-pkg-tftpd/files/usr/local/www/tftp_files.php
jim-p: Ah, perhaps suricata had some files with names that could benefit from translation (e.g. rule categories) that wouldn't fit the same case as files uploaded by a user. That makes sense. Just swap `gettext` out for `htmlspecialchars` and you'll be good here then.

doktornotor:
Well, this has been hardcoded to ```/tftpboot``` all the way up to 2.3 when the package went MIA - https://github.com/pfsense/pfsense-packages/blob/master/config/tftp2/tftp.inc#L60 - plus been the default with the only 2.3+ package version released so far. 

Can do something like ```define('FILES_DIR', '/tftpboot');``` and use that across the rest of code, so if someone really needs something else badly, then can change a single line in the package.

jim-p:
That sounds good then.

doktornotor:
@jim-p - Revamped and cleaned up the code quite a bit again, please check again. Thanks for feedback.

PR #264, opened 2017-01-24
darkstat package improvements

doktornotor:
- Fix service handling
- Improve bootstrap conversion
- Remove useless custom_add_php_command
- Input validation clean up

### DIFF
--- a/net-mgmt/pfSense-pkg-darkstat/Makefile
+++ b/net-mgmt/pfSense-pkg-darkstat/Makefile
@@ -1,8 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-darkstat
-PORTVERSION=	3.1.2
-PORTREVISION=	1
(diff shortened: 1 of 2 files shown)


doktornotor:
BTW, looking at this, shouldn't the ```<custom_php_global_functions>``` have all the required includes for various functions used there? Doesn't seem to be an issue, so it somehow works in background, but still.

jim-p:
The `custom_php_global_functions` code is run before code like resync, deinstall, and so on. It's meant to be a place where functions can be defined for use in other parts of package code. It will inherit the includes of what called it (either `pkg.php`, `pkg_edit.php`, or `pkg-utils.inc`) at a minimum.

doktornotor:
OK, thanks... works for now. I'll split this to an ```.inc``` file when I have more time. (Also would like to get the port setting working, using some hack such as the redirect.php used in ntopng, but no time for that ATM.)

PR #267, opened 2017-01-25
freeradius service handling fixes (Bug #6404), fix chown handling and various bugs

doktornotor:
Service handling fixes:
- Remove unwanted freeradius restart on install/upgrade
This breaks things due to restarting the service multiple times - once here and more times when custom_php_resync_config_command is run on install, which in turn runs freeradius_settings_resync() which restarts the service, and then freeradius_users_resync() runs which restarts the service yet again.
- Add an argument to skip service restart to freeradius_settings_resync() function
- Do not restart service twice on resync
- Do not restart on boot
- Attempt to fix XMLRPC sync which was restarting the service four times for a good measure, WTF.

Other fixes:
- Remove an evil recursive chown call on /var/log screwing up logdir permissions for things like squid, clamav, redis... Same unwanted recursive chown would be run on entire /usr/local/lib if $frlib was empty for some reason. Create a function with sanity checks, use PHP's chown()/chgrp() functions instead of exec() and refuse to recursively chown invalid targets.
- Stop creating a symlink to itself instead of /usr/local/etc/raddb directory
- Use safe_mkdir() in buch of places and fix some wrong checks while here.

Note: Would be worth getting everywhere, incl. RELENG_2_3_2:
https://forum.pfsense.org/index.php?topic=119569.0
https://forum.pfsense.org/index.php?topic=87441.0

### DIFF
--- a/net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
+++ b/net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
@@ -163,10 +163,9 @@ SERVICENAME="radiusd"
 EOD;
 	$rcfile['stop'] = FREERADIUS_ETC . '/rc.d/radiusd onestop';
 	write_rcfile($rcfile);
-	start_service("radiusd");
 }
 
-function freeradius_settings_resync() {
+function freeradius_settings_resync($restart_svc = true) {
 	global $config;
 	$conf = '';
 
@@ -410,6 +409,9 @@ EOD;
 	exec("ldconfig -m /usr/local/lib/mysql");
 	// Change owner of freeradius created files
 	exec("chown -R root:wheel /var/log");
+	if ($restart_svc) {
+		restart_service("radiusd");
+	}
 }
 
 function freeradius_users_resync() {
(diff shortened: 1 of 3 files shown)


net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -409,11 +444,16 @@ EOD;
...
+		freeradius_chown_recursive("/var/log/radacct");
+	}
+
+	if ($restart_svc) {
+		restart_service("radiusd");
+	}
 }
 
-function freeradius_users_resync() {
+function freeradius_users_resync($via_rpc = "no") {
```
rbgarga: Please use boolean type for $via_rpc

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -634,11 +674,15 @@ EOD;
...
+	// Do not restart on boot
+	// Will get restarted later by freeradius_clients_resync() if called via XMLRPC sync
+	if ($via_rpc == "no" && !platform_booting()) {
+		restart_service('radiusd');
+	}
 }
 
 
-function freeradius_authorizedmacs_resync() {
+function freeradius_authorizedmacs_resync($via_rpc = "no") {
```
rbgarga: Please use boolean type for $via_rpc

doktornotor:
@rbgarga - Done. There should be a lot more done here, but I'd like to keep changes here to a minimum, considering this should target a stable release.

doktornotor:
Found yet more restart madness. Argh!!!

PR #268, opened 2017-01-26
freeradius service handling fixes (Bug #6404) - round #2

doktornotor:
- Honor $restart_svc in freeradius_clients_resync() - missed in previous round of fixes
- Add $restart_svc argument to freeradius_allcertcnf_resync() for consistency and in case it's needed in future (currently only used by freeradiuscerts.xml)
- Make the checks syntax consistent across functions for readability
- Improve some comments


doktornotor:
@rbgarga - Hopefully should be really complete now :) Sorry.

PR #269, opened 2017-01-27
Squid - implement SSL MITM mode options

doktornotor:
This implements a couple of options for configuring SSL MITM interception, as discussed @ https://forum.pfsense.org/index.php?topic=123461.0

- Splice Whitelist, Bump Otherwise
This is the current code, still used by default.
- Splice All
Will splice everything. No need to install CA certificate on clients, and lets SquidGuard do its job. No content filtering (AV) possible, obviously.
- Custom
Use advanced custom options. Tinker with it as you wish. Unsupported, if you break it, fix it yourself.

Tweak a couple of descriptions/comments and fix some tags while here.

### DIFF
--- a/www/pfSense-pkg-squid/files/usr/local/pkg/squid.xml
+++ b/www/pfSense-pkg-squid/files/usr/local/pkg/squid.xml
@@ -309,7 +309,45 @@
 			<fieldname>ssl_proxy</fieldname>
 			<description>Enable SSL filtering.</description>
 			<type>checkbox</type>
-			<enablefields>ssl_active_interface,dca,sslcrtd_children,ssl_proxy_port,interception_checks</enablefields>
+			<enablefields>ssl_proxy_port,sslcrtd_children</enablefields>
+		</field>
+		<field>
+			<fielddescr>SSL/MITM Mode</fielddescr>
+			<fieldname>sslproxy_mitm_mode</fieldname>
+			<description>
+				<![CDATA[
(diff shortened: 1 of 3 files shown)


www/pfSense-pkg-squid/files/usr/local/pkg/squid.xml
```diff
@@ -686,10 +723,58 @@
...
+		</field>
 	</fields>
 	<custom_php_validation_command>
 		squid_validate_general($_POST, $input_errors);
 	</custom_php_validation_command>
+	<!--
+	<custom_php_after_form_command>
+		squid_print_javascript_general2();
+	</custom_php_after_form_command>
+	-->
```
rbgarga: What is the idea about adding this commented out code here?

doktornotor: Well the idea is getting some JS magic there as in Antivirus/Authentication tab, but it's
- not ready yet
- mostly unrelated to this change
- not compatible with RELENG_2_3_2 in case someone decides to cherrypick this to that branch

doktornotor:
@jim-p Thanks for testing. Now, if only someone fixed/rewrote the SquidGuard thing. :/ I gave up everytime after 5 minutes.

jim-p:
I agree, re: SquidGuard. Every time I have to touch that code it makes my brain hurt, and that's just from the whitespace and formatting.

PR #270, opened 2017-01-27
Squid - dashboard widget bootstrap facelift + fixes

doktornotor:
- replace inline base64 images with bootstrap icons
- tidy up tables and remove old styling
- fix squiclamav version detection and use pkg_exec() to get packages' versions in general
- fix version and HTML on squid_clwarn.php page

### DIFF
--- a/www/pfSense-pkg-squid/files/usr/local/www/squid_clwarn.php
+++ b/www/pfSense-pkg-squid/files/usr/local/www/squid_clwarn.php
@@ -54,8 +54,11 @@ error_log(date("Y-m-d H:i:s") . " | VIRUS FOUND | "
 	. str_replace('|', '', $user) . "\n", 3, "/var/log/c-icap/virus.log");
 
 ?>
+<!DOCTYPE html>
+<html lang="en">
+<head>
 <style type="text/css">
-	.visu {
+.visu {
 	border:1px solid #C0C0C0;
 	color:#FFFFFF;
 	position: relative;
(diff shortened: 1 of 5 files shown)


www/pfSense-pkg-squid/files/usr/local/www/squid_clwarn.php
```diff
@@ -19,7 +19,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-$VERSION = '6.10';
+require_once("pkg-utils.inc");
+pkg_exec("query '%v' squidclamav", $version, $err);
```
rbgarga: It's missing to check pkg_exec return code to make sure it succeeded

doktornotor: Well it is working perfectly fine here? *confused*

rbgarga: pkg_exec returns 0 for success and != 0 when it fails. It would be good to check if return is 0 before assume $version really contains a valid value

doktornotor:
@rbgarga - No idea what scripts are you guys using for the cherrypick stuff, but they are broken. 

https://github.com/pfsense/FreeBSD-ports/blob/RELENG_2_3/www/pfSense-pkg-squid/files/etc/inc/priv/squid.priv.inc

Completely screwed people's access to WebGUI. https://forum.pfsense.org/index.php?topic=124814.msg689342#msg689342

Can we get rid of this licensing schizo, perhaps?

rbgarga:
@doktornotor fixed. Thanks!

PR #272, opened 2017-01-28
freeradius code style and other fixes

doktornotor:
Code style/other fixes
- Fix code style, whitespace, indentation
- Make things much more readable by leaving out the  middle part of the```?:``` ternary operator
- Replace exec() with mwexec() 
- Use native PHP functions where possible/suitable
- Replace horrible code using echo redirection from shell with file_put_contents()
- Use openvpn_create_dhparams() to create DH params file
- Do not use code like ```exec("cd $foo; bar")```
- Use full paths to executables in mwexec()
- Improve comments
- Etc. ...

Others:
- Add prominent deprecation warnings to 'Certificates' and 'EAP' tabs regarding the redundant FreeRADIUS Cert Manager feature. See https://redmine.pfsense.org/issues/7170.
- Fix Bug #6928 - "Access-Reject" not logged in MySQL radpostauth table

### DIFF
--- a/net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
+++ b/net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
@@ -26,6 +26,7 @@ require_once("filter.inc");
 require_once("functions.inc");
 require_once("globals.inc");
 require_once("notices.inc");
+require_once("openvpn.inc");
 require_once("pkg-utils.inc");
 require_once("services.inc");
 require_once("service-utils.inc");
@@ -48,7 +49,7 @@ define('FREERADIUS_ETC', FREERADIUS_BASE . '/etc');
  * freeradius_eapconf_resync($restart_svc = true)
  * freeradius_modulesldap_resync($restart_svc = true)
  * freeradius_allcertcnf_resync($restart_svc = true)
-*/
+ */
 
 // Check freeradius lib version
 	$frlib = "";
@@ -61,7 +62,7 @@ define('FREERADIUS_ETC', FREERADIUS_BASE . '/etc');
 		}
 	}
 	if ($frlib == "") {
-		log_error("freeRADIUS - No freeradius libs found on " . FREERADIUS_LIB);
+		log_error("freeRADIUS: No freeradius libs found on " . FREERADIUS_LIB);
 	}
 
 function freeradius_deinstall_command() {
@@ -79,7 +80,7 @@ function freeradius_deinstall_command() {
 
 function freeradius_chown_recursive($dir, $user = "root", $group = "wheel") {
 	if (empty($dir) || ($dir == '/') || ($dir == '/usr/local') || ($dir == '/usr/local/etc') || ($dir == '/usr/local/lib') || ($dir == '/var/log') || !is_dir($dir)) {
-		log_error(gettext("[freeradius] Attempted to recursively chown an invalid directory: '{$dir}'"));
+		log_error(gettext("freeRADIUS: Attempted to recursively chown an invalid directory: '{$dir}'"));
 		return;
 	}
 	chown($dir, $user);
@@ -96,7 +97,7 @@ function freeradius_chown_recursive($dir, $user = "root", $group = "wheel") {
 			}
 		}
 	} else {
-		log_error(gettext("[freedarius] freeradius_chown_recursive() call failed; permissions not set for directory: '{$dir}'"));
+		log_error(gettext("freeRADIUS: freeradius_chown_recursive() call failed; permissions not set for directory: '{$dir}'"));
 	}
 }
 
@@ -140,8 +141,12 @@ function freeradius_install_command() {
 	}
 
 	// Disable virtual-server we do not need by default
-	if (file_exists(FREERADIUS_ETC . "/raddb/sites-enabled/control-socket")) { unlink(FREERADIUS_ETC . "/raddb/sites-enabled/control-socket"); }
-	if (file_exists(FREERADIUS_ETC . "/raddb/sites-enabled/inner-tunnel")) { unlink(FREERADIUS_ETC . "/raddb/sites-enabled/inner-tunnel"); }
+	if (file_exists(FREERADIUS_ETC . "/raddb/sites-enabled/control-socket")) {
+		unlink(FREERADIUS_ETC . "/raddb/sites-enabled/control-socket");
+	}
+	if (file_exists(FREERADIUS_ETC . "/raddb/sites-enabled/inner-tunnel")) {
+		unlink(FREERADIUS_ETC . "/raddb/sites-enabled/inner-tunnel");
+	}
 
 	// We run this here just to suppress some warnings on syslog if file doesn't exist
 	freeradius_authorizedmacs_resync(false, false);
@@ -264,28 +269,27 @@ function freeradius_settings_resync($restart_svc = true) {
 	$varsettingsmaxqueuesize = ($varsettings['varsettingsmaxqueuesize'] ?: '65536');
 	$varsettingsmaxrequestsperserver = ($varsettings['varsettingsmaxrequestsperserver'] ?: '0');
 
-	// For more details look at "freeradius_sqlconf_resync"
+	// For more details look at freeradius_sqlconf_resync()
 	if (is_array($config['installedpackages']['freeradiussqlconf']['config'][0])) {
 		$sqlconf = $config['installedpackages']['freeradiussqlconf']['config'][0];
 	} else {
 		$sqlconf = array();
 	}
 
-	// Dis-/Enable SQL in "instatiate" section in "freeradius_settings_resync" and radiusd.conf SQL SERVER 2
+	// Dis-/Enable SQL in "instatiate" section in freeradius_settings_resync() and radiusd.conf SQL SERVER 2
 	if ($sqlconf['varsqlconf2includeenable'] == 'on') {
 		$varsqlconf2instantiate = 'sql2';
-	}
-	else {
+	} else {
 		$varsqlconf2instantiate = '### sql2 DISABLED ###';
 	}
 
 	$varsqlconf2failover = ($varsettings['varsqlconf2failover'] ?: 'redundant');
 
-	// Dis-/Enable SQL in "instatiate" section in "freeradius_settings_resync" and radiusd.conf SQL SERVER 1
+	// Dis-/Enable SQL in "instatiate" section in freeradius_settings_resync() and radiusd.conf SQL SERVER 1
 	if ($sqlconf['varsqlconfincludeenable'] == 'on') {
 		$varsqlconfinclude = '$INCLUDE sql.conf';
 		$varsqlconfincludecounter = '$INCLUDE sql/mysql/counter.conf';
-		$varsqlconfinstantiate = "$varsqlconf2failover {" . "\n\t\tsql" . "\n\t\t$varsqlconf2instantiate" . "\n\t}";
+		$varsqlconfinstantiate = "{$varsqlconf2failover} {" . "\n\t\tsql" . "\n\t\t{$varsqlconf2instantiate}" . "\n\t}";
 	} else {
 		$varsqlconfinclude = '#$INCLUDE sql.conf';
 		$varsqlconfincludecounter = '#$INCLUDE sql/mysql/counter.conf';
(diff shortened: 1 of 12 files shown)


net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -140,8 +141,12 @@ function freeradius_install_command() {
...
 
 	// Disable virtual-server we do not need by default
-	if (file_exists(FREERADIUS_ETC . "/raddb/sites-enabled/control-socket")) { unlink(FREERADIUS_ETC . "/raddb/sites-enabled/control-socket"); }
-	if (file_exists(FREERADIUS_ETC . "/raddb/sites-enabled/inner-tunnel")) { unlink(FREERADIUS_ETC . "/raddb/sites-enabled/inner-tunnel"); }
+	if (file_exists(FREERADIUS_ETC . "/raddb/sites-enabled/control-socket")) {
+		unlink(FREERADIUS_ETC . "/raddb/sites-enabled/control-socket");
+	}
+	if (file_exists(FREERADIUS_ETC . "/raddb/sites-enabled/inner-tunnel")) {
+		unlink(FREERADIUS_ETC . "/raddb/sites-enabled/inner-tunnel");
+	}
```
rbgarga: Replace these if statements by a simple call to unlink_if_exists() or @unlink()

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -326,62 +337,67 @@ extended_expressions = $varsettingsextendedexpressions
 
 EOD;
 
-// Deletes virtual-server coa by default. Will be re-enabled if there is an interface-type "coa"
-exec("rm -f " . FREERADIUS_ETC . "/raddb/sites-enabled/coa");
+	// Deletes virtual-server coa by default. Will be re-enabled if there is an interface-type "coa"
+	mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/sites-enabled/coa");
```
rbgarga: Use unlink_if_exists() instead of rm -f

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -447,15 +463,16 @@ EOD;
...
 	// to run freeradius_serverdefault_resync() and after that restart freeradius.
 	freeradius_plainmacauth_resync();
 	freeradius_motp_resync();
 	freeradius_serverdefault_resync();
-	
+
 	// This is to fix the mysqlclient.so which gets lost after reboot
-	exec("ldconfig -m /usr/local/lib/mysql");
+	// XXX: Probably some PBI leftover, remove?
+	mwexec("/sbin/ldconfig -m /usr/local/lib/mysql");
```
rbgarga: I strongly believe this could be removed. MySQL package registers itself to run when /etc/rc.d/ldconfig is called by rc script

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -467,225 +484,251 @@ EOD;
...
+	} else {
+		$arrusers = array();
+	}
 
-$conf = '';
+	if (!empty($arrusers)) {
+		foreach ($arrusers as $users) {
 
-$arrusers = $config['installedpackages']['freeradius']['config'];
+		// Variables for users file defined parameters
```
rbgarga: Looks like this block is not correctly indented

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -696,194 +739,222 @@ EOD;
...
-		foreach ($varmacstopadditionaloptions as $toptmp) {
-			$varmacsadditionaloptionstop .= $toptmp . "\n";
+		$arrmacs = array();
+	}
+
+	if (!empty($arrmacs)) {
+		foreach ($arrmacs as $macs) {
+
+		// Variables for authorized_macs file defined parameters
+		$varmacsaddress = $macs['varmacsaddress'];
```
rbgarga: Not correctly indented

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -2499,122 +2551,116 @@ EOD;
...
+		// If all certs should be deleted, we do not need to delete and recreate client-certs first.
+		if ($arrcerts['varcertsdeleteall'] == 'no') {
+			if ($arrcerts['varcertscreateclient'] == 'yes') {
+				// delete all old certificates and keys
+				log_error("freeRADIUS: deleting all client.csr .crt .key .pem .tar in " . FREERADIUS_ETC . "/raddb/certs");
+				mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/client.csr");
+				mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/client.crt");
+				mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/client.key");
+				mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/client.pem");
+				mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/client.tar");
```
rbgarga: Use unlink_if_exists()

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -2499,122 +2551,116 @@ EOD;
...
+				$varserial = preg_replace("/\s/","",file_get_contents(FREERADIUS_ETC . '/raddb/certs/serial.old'));
+					if (file_exists(FREERADIUS_ETC . "/raddb/certs/$varserial.pem")) {
+						rename(FREERADIUS_ETC . "/raddb/certs/$varserial.pem", FREERADIUS_ETC . "/raddb/certs/client.pem");
+					}
+
+				// tar client-cert files
+				mwexec("/usr/bin/tar -C " . FREERADIUS_ETC . "/raddb/certs -cf " . FREERADIUS_ETC . "/raddb/certs/client.tar client.crt client.csr client.key ca.der client.pem");
+
+				// Make all files in certs folder read/write only for root
+				mwexec("/bin/chmod -R 0600 " . FREERADIUS_ETC . "/raddb/certs/");
```
rbgarga: Use chmod()

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -2499,122 +2551,116 @@ EOD;
...
+				mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/ca.csr && /bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/server.csr && /bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/client.csr");
+				mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/ca.crt && /bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/server.crt && /bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/client.crt");
+				mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/ca.key && /bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/server.key && /bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/client.key");
+				mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/ca.p12 && /bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/server.p12 && /bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/client.p12");
+				mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/serial*");
+				mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/index*");
+				mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/dh");
+				mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/random");
+				mwexec("/bin/rm -f " . FREERADIUS_ETC . "/raddb/certs/client.tar");
```
rbgarga: Use unlink_if_exists()

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -2499,122 +2551,116 @@ EOD;
...
+
+				// run fuctions to create new .cnf files based on user input from freeradiuscert.xml
+				freeradius_cacertcnf_resync();
+				freeradius_servercertcnf_resync();
+				freeradius_clientcertcnf_resync();
+
+				// this command deletes the pfsense_cert_mgr checkfile so when we change back to pfsense cert manager a new DH + random file will be created
+				if (file_exists(FREERADIUS_ETC . "/raddb/certs/pfsense_cert_mgr")) {
+					unlink(FREERADIUS_ETC . "/raddb/certs/pfsense_cert_mgr");
+				}
```
rbgarga: Use unlink_if_exists()

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -2499,122 +2551,116 @@ EOD;
...
+				mwexec(FREERADIUS_ETC . "/raddb/certs/bootstrap");
+
+				// rename client generated 02.pem to client.pem
+				if (file_exists(FREERADIUS_ETC . "/raddb/certs/02.pem")) {
+					rename(FREERADIUS_ETC . "/raddb/certs/02.pem", FREERADIUS_ETC . "/raddb/certs/client.pem");
+				}
+
+				// tar client-cert files
+				mwexec("/usr/bin/tar -C " . FREERADIUS_ETC . "/raddb/certs -cf " . FREERADIUS_ETC . "/raddb/certs/client.tar client.crt client.csr client.key ca.der client.pem");
+				mwexec("/bin/chmod -R 0600 " . FREERADIUS_ETC . "/raddb/certs/");
```
rbgarga: Use chmod()

doktornotor:
As for the "use chmod()" - it needs to be recursive where used, not really an option without complicating the code even further. Roger otherwise. Thanks for taking time to read this monster :)

rbgarga:
@doktornotor it won't complicate too much because certs directory is supposed to contain only files and not subdirectories. You can do it like:
```
$certs = glob(FREERADIUS_ETC . "/raddb/certs/*");
array_walk($certs, function($cert_file) {
	chmod($cert_file, 0600);
});
```

doktornotor:
I'll see what I can do with that. Other than that, ```unlink_if_exists()``` does handle globbing correctly, right? (So that all the ```rm -f``` junk can be replaced.)

rbgarga:
Yes, it handles globbing. But anyway, -f is for force removal and not recursive option

doktornotor:
Well yeah, but the (unreadable) stuff is terminated with ```*``` in bunch of places, such as this:
```/var/log/radacct/datacounter/$varusersmaxtotaloctetstimerange/used-octets-$varusersusername*```
(extremely easy to miss)

Another WTF:
Do you think the recursive chown on ```$frlib``` (line 128) really serves any useful purpose now? NFC about how it got there in the first place and why.

doktornotor:
All done minus chmod(). The array_walk() thing does not work like this (```syntax error, unexpected '}'```) and I'd really rather leave that thing alone, just makes the code unreadable.

rbgarga:
@doktornotor I've tested array_walk before send it to you, it works. You probably had some copy/paste issue. It's a common PHP usage and also in pfSense you can see it being used at https://github.com/pfsense/pfsense/blob/master/src/usr/local/www/widgets/widgets/dyn_dns_status.widget.php#L59

doktornotor:
Done. (Note: Pasting from browser can screw newlines and the PHP error is not helpful at all.) So we now have 8 lines of PHP instead of two simple ones. Oh well...

rbgarga:
Thanks! It's better to have more PHP lines than spawning shell sessions to run system binaries

doktornotor:
Added a tiny patch for https://redmine.pfsense.org/issues/6928 (verified by a user) - and that's really all here unless some other changes are requested.

PR #273, opened 2017-01-29
pfSense-pkg-apcupsd - new option for set shutdown behavior

dbaio:
This quote from apccontrol explains:
```
[...]
# You can customize every single command creating an executable file (may be a
# script or a compiled program) and calling it the same as the $1 parameter
# passed by apcupsd to this script.
#
# After executing your script, apccontrol continues with the default action.
# If you do not want apccontrol to continue, exit your script with exit 
# code 99. E.g. "exit 99".
[...]
```

Suggested and tested by LucaTo from pfsense forum. Thank you.

### DIFF
--- a/sysutils/pfSense-pkg-apcupsd/files/usr/local/pkg/apcupsd.inc
+++ b/sysutils/pfSense-pkg-apcupsd/files/usr/local/pkg/apcupsd.inc
@@ -140,6 +140,7 @@ function sync_package_apcupsd() {
 			$annoy = $apcupsd_config['annoy'] ?: "300";
 			$annoydelay = $apcupsd_config['annoydelay'] ?: "60";
 			$killdelay = $apcupsd_config['killdelay'] ?: "0";
+			$shutdownbehavior = $apcupsd_config['shutdownbehavior'];
 			$netserver = $apcupsd_config['netserver'];
 			$nisip= $apcupsd_config['nisip'] ?: "0.0.0.0";
 			$nisport = $apcupsd_config['nisport'] ?: "3551";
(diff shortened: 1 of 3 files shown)


sysutils/pfSense-pkg-apcupsd/files/usr/local/pkg/apcupsd.inc
```diff
@@ -259,6 +260,23 @@ EOF;
 		file_put_contents(APCUPSD_BASE . "/etc/apcupsd/" . $apccontrol_script, $apccontrol_script_file, LOCK_EX);
 	}
 
+	if ($shutdownbehavior == "poweroff") {
+		$doshutdown_script_file =<<<EOF
+#!/bin/sh
+
+printf "Beginning Shutdown Sequence (shutdown -p)" | wall
+/sbin/shutdown -p now "apcupsd initiated shutdown"
```
rbgarga: Do not call /sbin/shutdown direct, you should use /etc/rc.halt instead

dennypage:
Calling /etc/rc.halt is probably not a good idea. If there are config changes happening, the shutdown could be denied which is not what you want. The pending loss of UPS power should be viewed as an overriding event, and higher priority than a pending config change.

dbaio:
Good point @dennypage.
By default apcupsd through apccontrol call this:

```
   doshutdown)
      printf "Beginning Shutdown Sequence" | wall
      ${SHUTDOWN} -h now "apcupsd initiated shutdown"
   ;;
```
Could I revert this last commit ?

dennypage:
I would think so, but you should wait for Renato’s agreement.

rbgarga:
The chances of this to happen are really small, it waits 30 seconds trying to acquire the lock, if it cannot after 30 seconds it means something is really wrong.

I'd say the best option would be to add a new parameter to rc.halt, -f, that would mean to force shutdown even if lock was not acquired.

dennypage:
pfBlockerNG?

Regardless, 30 seconds is really too long a potential delay in a critical power situation. Power failure is an overriding event. The value of trying to complete a config change a few seconds before power failure is pretty much nil. Definitely not worth the risk of corrupting the file system.

rbgarga:
Yeah, indeed, you are right. @dbaio can you please revert it?

dbaio:
Sure @rbgarga. Done.

PR #276, opened 2017-01-31
darkstat - make port configurable, split PHP from XML, add work around for Bug #7178

doktornotor:
- Split PHP code out of XML to darkstat.inc
- Make darkstat webgui port configurable
- Add a description tag to XML work around Bug #7178 - impossible to enable the service on pfSense themes otherwise; works properly with Compact-Red)

### DIFF
--- a/net-mgmt/pfSense-pkg-darkstat/files/usr/local/pkg/darkstat.inc
+++ b/net-mgmt/pfSense-pkg-darkstat/files/usr/local/pkg/darkstat.inc
@@ -1,0 +1,221 @@
+<?php
+/*
+ * darkstat.inc
+ *
+ * part of pfSense (https://www.pfsense.org)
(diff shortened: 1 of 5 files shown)


net-mgmt/pfSense-pkg-darkstat/files/usr/local/pkg/darkstat.inc
```diff
@@ -0,0 +1,221 @@
...
+		if ($if) {
+			$ifaces_final .= " -i {$if}";
+		}
+	}
+	$start = "/usr/local/sbin/darkstat {$ifaces_final}";
+
+	/* Configure bind interface(s) */
+	foreach (explode(",", $darkstat_config['bind_interfaces']) as $iface) {
+		$if = convert_friendly_interface_to_real_interface_name($iface);
+		$if = get_interface_ip("$iface");
```
rbgarga: $if is being attributed twice and first one is not used

PR #279, opened 2017-01-31
Squuid RELENG_2_3 - fix screwed cherrypick merge

doktornotor:
https://forum.pfsense.org/index.php?topic=124814.0


rbgarga:
Already fixed, thanks

PR #283, opened 2017-01-31
Squid - Re-add missing closing tag

doktornotor:
This went MIA with c995dda201ccf40b2c4a16ff65d90315360a6463 commit for god knows what reason.


doktornotor:
Guys really, what's eating these tags? On quick check I've found:
https://github.com/pfsense/FreeBSD-ports/blob/devel/sysutils/pfSense-pkg-AutoConfigBackup/files/usr/local/pkg/autoconfigbackup.inc#L305

https://github.com/pfsense/FreeBSD-ports/blob/devel/security/pfSense-pkg-openvpn-client-export/files/usr/local/pkg/openvpn-client-export.inc#L1077

It doesn't seem to really break anything, but still... *huh*

rbgarga:
Removing the closing tag is kind of recommended by PHP as you can see at http://php.net/basic-syntax.instruction-separation Note.

doktornotor:
Well thanks for link. I prefer my PHP closed, but the point was there seems to be some tag eating monster somewhere :rofl: Pretty sure those things used to have the closing tag, except for ACB. E.g., suricata is pretty much a copy of snort code, one has it, the other one doesn't.

PR #284, opened 2017-02-01
lightsquid - Hack around incredible lighttpd idiocy with HTTPS/IPv6

doktornotor:
See comments in the file for explanation. Lighttpd == POS.

### DIFF
--- a/www/pfSense-pkg-Lightsquid/files/usr/local/pkg/lightsquid.inc
+++ b/www/pfSense-pkg-Lightsquid/files/usr/local/pkg/lightsquid.inc
@@ -202,10 +202,10 @@ server.port			= "{$lightsquid_config['lighttpd_ls_port']}"
 
 EOF;
 	} else {
-	/* We ALWAYS need to put server.port there; otherwise
-	 * the retarded lighttpd thing binds to 80/HTTP no matter what
-	 */
-	$lighttpd_ls_conf .= <<<EOF
+		/* We ALWAYS need to put server.port there; otherwise
+	 	* the retarded lighttpd thing binds to 80/HTTP no matter what
+	 	*/
+		$lighttpd_ls_conf .= <<<EOF
 server.port			= "{$lightsquid_config['lighttpd_ls_port']}"
 
 EOF;
--- a/www/pfSense-pkg-Lightsquid/files/usr/local/pkg/lightsquid.inc
+++ b/www/pfSense-pkg-Lightsquid/files/usr/local/pkg/lightsquid.inc
@@ -201,8 +201,15 @@ server.port			= "{$lightsquid_config['lighttpd_ls_port']}"
 \$SERVER["socket"]		== "[::]:{$lightsquid_config['lighttpd_ls_port']}" {  }
 
 EOF;
-	}
+	} else {
+	/* We ALWAYS need to put server.port there; otherwise
+	 * the retarded lighttpd thing binds to 80/HTTP no matter what
+	 */
+	$lighttpd_ls_conf .= <<<EOF
+server.port			= "{$lightsquid_config['lighttpd_ls_port']}"
 
+EOF;
+	}
 	$lighttpd_ls_conf .= <<<EOF
 server.event-handler		= "freebsd-kqueue"
 server.network-backend		= "freebsd-sendfile"
@@ -328,12 +335,16 @@ ssl.honor-cipher-order	= "enable"
 ssl.cipher-list		= "AES128+EECDH:AES256+EECDH:AES128+EDH:AES256+EDH:AES128-SHA:AES256-SHA:!aNULL:!eNULL:!DSS"
 EOF;
 
-		// Configure IPv4 and IPv6 sockets
+		/*
+		 * IPv4 HTTPS bind must NOT be a socket; otherwise the retarded
+		 * lighttpd thing yet again binds to 80/HTTP no matter what.
+		 * You enable SSL-only IPv4 by not configuring bind for IPv4 at all.
+		 * IPv6 SSL must be configured as a socket
+		 */
 		$lighttpd_ls_conf .= <<<EOF
 
-\$SERVER["socket"]	== "0.0.0.0:{$lightsquid_config['lighttpd_ls_port']}" {
 {$lighttpd_ls_conf_sslcommon}
-}
+
 
 \$SERVER["socket"]	== "[::]:{$lightsquid_config['lighttpd_ls_port']}" {
 {$lighttpd_ls_conf_sslcommon}
--- a/www/pfSense-pkg-Lightsquid/Makefile
+++ b/www/pfSense-pkg-Lightsquid/Makefile
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-Lightsquid
-PORTVERSION=	3.0.5
+PORTVERSION=	3.0.6
 CATEGORIES=	www
 MASTER_SITES=	# empty
 DISTFILES=	# empty


www/pfSense-pkg-Lightsquid/files/usr/local/pkg/lightsquid.inc
```diff
@@ -328,12 +335,16 @@ ssl.honor-cipher-order	= "enable"
...
 EOF;
 
-		// Configure IPv4 and IPv6 sockets
+		/*
+		 * IPv4 HTTPS bind must NOT be a socket; otherwise the retarded
+		 * lighttpd thing yet again binds to 80/HTTP no matter what.
+		 * You enable SSL-only IPv4 by not configuring bind for IPv4 at all.
+		 * IPv6 SSL must be configured as a socket
+		 */
 		$lighttpd_ls_conf .= <<<EOF
```
gstrauss: Yes, server.port defaults to port 80 and lighttpd will bind to \*:80 or whatever server.port is set to.  If you do not want the default behavior of binding to port 80, then you need to specify server.port and, optionally, server.bind.  You can also optionally specify a repeat of the implicit bind to "\*:$port" in a $SERVER["socket"] block.

You do not need to repeat $lighttpd_ls_conf_sslcommon inside this block if it is already present in the global scope.

```
server.port		= "{$lightsquid_config['lighttpd_ls_port']}"
# \$SERVER["socket"]	== "0.0.0.0:{$lightsquid_config['lighttpd_ls_port']}" { }
\$SERVER["socket"]	== "[::]:{$lightsquid_config['lighttpd_ls_port']}" { }
{$lighttpd_ls_conf_sslcommon}
```

doktornotor: That block absolutely MUST be ther exactly as noted in the comments. Once for global scope and for each and every socket. I've damaged two keyboards while messing with this POS, trust me. Omit that block and the socket will be HTTP and not HTTPS.

gstrauss: My mistake.  You are correct that you need to repeat the ssl config in each $SERVER["socket"] (and in the global scope if you want HTTPS-only).  I'll see about fixing that behavior in lighttpd 1.4.46.

Regarding your second comment, how do things break with break with IPv6 if you set `server.bind`?

doktornotor: The server.bind thing yet again created an unwanted HTTP/80 when IPv4/HTTPS it was configured as a socket (So, I moved the port to global scope here and stopped doing a bind there altogether). We want this either HTTP only or HTTPS only on wildcard IPv4 and IPv6.

To summarize this: 
- I've spent ~8 hours on this.
- There are Debian/Ubuntu bugs about this madness and inability of the package maintainers to produce OOB working configuration.
- There are questions on StackOverflow and none of them answers this kind of questions, such as "how to stop repeating that SSL block for each socket, or how to skip HTTP to stop clashing with already existing webserver on the same machine.)

Guys, dunno, but configuring a basic webserver configuration is absolutely NOT supposed to be such incredible PITA. 
- The code should respect the configuration. When I create two sockets specifically bound to a particular port and protocol, last thing I expect is that the webserver will "smartly" create an unconfigured HTTP/80 bind and commit a suicide because 80 is already in use
- Configuring an IPv6 wildcard bind is a royal pain (again, see the Debian/Ubuntu/etc. bugs)
- If that was not enough, things get gradually worse when ```net.ipv6.bindv6only = 1``` comes into play. Again, see Debian and other bugs where people complain about lighttpd listening on IPv6 *only* all of a sudden.
- And the wiki article is completely useless joke which lacks any working examples of common configurations and only confuses things further with version-specific stuff and the  ```net.ipv6.bindv6only = 1``` thing.

gstrauss: Thanks for the constructive feedback.  I'll do what I can to address them in the next release of lighttpd, specifically the current need to repeat the ssl config for each socket.

Regarding the behavior of lighttpd to default bind to port 80, that is intended as a user-friendly part of a simple config.  The global scope binds to a socket.  If you don't specify one in `server.bind`, it is "\*".  If you do not want the default port 80, then specify `server.port`.
```
server.port		= "{$lightsquid_config['lighttpd_ls_port']}"
\$SERVER["socket"]	== "0.0.0.0:{$lightsquid_config['lighttpd_ls_port']}" { 
    {$lighttpd_ls_conf_sslcommon}
}
\$SERVER["socket"]	== "[::]:{$lightsquid_config['lighttpd_ls_port']}" {
    {$lighttpd_ls_conf_sslcommon}
}
```

Regarding dual-stack IPv6, there is plenty of vitriol on the internet about IPv6 and dual-stack.  Still, I'll look into the possibility of simplifying, and will attempt to come up with a simple option for a commonly-desired config, again, with the complication of attempting reasonable backwards compatibility with existing lighttpd configs.

FYI: In the meantime, I have been spending plenty of time (much more than 8 hours :)) trying to get package maintainers for various distributions to update lighttpd and to accept patches for older versions.  Older version of lighttpd in stable Debian/Ubuntu releases are missing more than a few patches which address more than a few reported issues, though not (yet) the issues you reported above.

doktornotor: OK, it's really good to trigger some upstream response here, thanks. 

- My expectation is that if I don't configure ```server.bind``` well that should mean really any, and dual-stack if available, no? As in "I don't want to restrict the bind in any specific way, why configure it then? If I wanted, I'd put it into the config". The wiki somehow suggests that should be the case, but not really in the end. It just doesn't work that way. Not for the use case here, and not for many others.
- The "user-friendly" part cost me a lot of hair and two keyboards. Document how to skip it. Yeah, the ```server.port``` thing does avoid it, but that info is nowhere to be found.
- And when you've somehow managed that, you are still in the protocols mixup pitfall with dual stack. (As you've already acknowledged, the global ssl config is not really respected, so you end up with those repeated blocks of redundant code.)

P.S. As for backward compatibility - well yeah if you break the code we have here in future, I'll scream and vent here again. :rofl:

gstrauss:
The following will trigger a warning from lighttpd each time it is started up, but may be the dual-stack config you desire for HTTPS-only, which might work for you on *some* platforms (see further below):
```
server.bind = "[::]"
server.port = "{$lightsquid_config['lighttpd_ls_port']}"
server.set-v6only = "disable"
{$lighttpd_ls_conf_sslcommon}
```
The above will warn about server.set-v6only potentially being removed in the future.  Open to debate, and maybe I'll remove the warning in future releases.  However, the reason the warning exists is because there was an attempt at IPv6 dual-stack sanity (which turned out not to be possible/portable).  (Further explanation at bottom)

If you use the below config, you'll get the warning: "warning: please use server.use-ipv6 only for hostnames, not without server.bind / empty address; your config will break if the kernel default for IPV6_V6ONLY changes"
```
server.port = "{$lightsquid_config['lighttpd_ls_port']}"
server.set-v6only = "disable"
server.use-ipv6 = "enable"
{$lighttpd_ls_conf_sslcommon}
```
(Note the replacement of `server.bind`, now left empty)

Depending on the OS and its kernel config for dual-stack, which might be different between distros and even between distro major releases, the dual-stack config can be difficult to get "right" without being explicitly specified.  That is because the socket option IPV6_V6ONLY specifies IPv6-only bind, whereas absence of IPV6_V6ONLY means "whatever the operating system is configured to do by default".  That is the fragile part and why lighttpd issues a warning when not *explicitly* configured.  I suppose that lighttpd could define its behavior and then query socket options and try to "do the right thing" if it needs to issue multiple socket bind() commands to achieve the same result, but then lighttpd will also need special logic if the lighttpd.conf has $SERVER["socket"] conditions, since the now-multiple socket binds must map back to the same $SERVER["socket"], even though they are separate bind() to "different" addresses (one IPv4 and one IPv6).  This is all theoretically possible, but as you can see, it is not "simple".  Also, there is a difference between IPV6_V6ONLY bound sockets and IPv4 sockets, versus non-IPV6_V6ONLY sockets in that stringifying remote addresses for CGI and FastCGI environments can lead to v4-mapped IPv6 addresses which are different strings from the plain IPv4 string, though indicate identical addresses (e.g. "::ffff:127.0.0.1" versus "127.0.0.1").  Which is the "one" correct representation for everyone?  (none of the above)  Herein touches the surface why there are so many expletives surrounding attempts to configure dual-stack.  The only (partially) correct answer is to "do the right thing" when the user has explicitly specified what is wanted for "the right thing".

If you find the above more coherent than https://redmine.lighttpd.net/projects/lighttpd/wiki/IPv6-Config then I'll update the wiki, too.

doktornotor:
OK, thanks for the hint, however that doesn't work for us (and probably not for most people on FreeBSD because it kills IPv4. Both variants result in IPv6-only connectivity.

Suspect this to be a pretty widespread problem, on Debian at least it seems to cause the same unwanted result - https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=560837 

So, I guess I think I'll leave the code as is for now, as it seems to work and until things settle, it'd not be a good idea to upset users again with experiments breaking their config. Apparently, we still in search of an easy dual-stack wildcard bind config. Dunno, how about ```server.bind = "any"``` or whatever similar?

gstrauss:
http://stackoverflow.com/questions/5587935/cant-turn-off-socket-option-ipv6-v6only
> FreeBSD since 5.x has disabled IPv4 mapped on IPv6 addresses and thus unless you turn that feature back on by setting the required configuration flag in rc.conf you won't be able to use it. – X-Istence

Also, the Debian link you provided (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=560837) suggests that the default in Debian sets sysctl net.ipv6.bindv6only=1, so modern FreeBSD,  OS X, and Debian are IPV6_V6ONLY, no matter how `server.set-v6only` is set.

Ack: you'd like an easy way to configure dual stack, and in this case, having `ssl.*` directives inherited from the global config appears to be the missing piece for the current use case (not all use cases).

In order to preserve existing behavior with lighttpd configs, I do not want to simply have `ssl.*` directives in global scope inherited by all sockets and need some way to flag that the ssl config should be inherited, perhaps by `ssl.engine = "enable"` in the socket config, but no other `ssl.*` directives in that scope.

Proposal: if implemented as above, that would make your *future* lighttpd 1.4.46 config look like:
```
server.port		= "{$lightsquid_config['lighttpd_ls_port']}"
\$SERVER["socket"]	== "0.0.0.0" { ssl.engine = "enable" }
\$SERVER["socket"]	== "[::]"    { ssl.engine = "enable" }
{$lighttpd_ls_conf_sslcommon}
```
Not quite as simple as the nginx config, but it also does not break existing lighttpd configs because setting `ssl.engine = "enable"` without setting `ssl.pemfile` is currently an error in lighttpd.

@doktornotor: What do you think?

In addition, I might consider extending the $SERVER["socket"] syntax, though I am still mulling that over.

doktornotor:
@gstrauss -Well yes, that definitely looks *much* better, and also it's something people would expect to be a working config, instead of throwing errors, I'd say.

Nice to have some progress here, thanks!

gstrauss:
I updated https://redmine.lighttpd.net/projects/lighttpd/wiki/IPv6-Config with some background info on dual-stack and more examples for TLS/SSL (even though that page is IPv6-Config) and I have queued up a change for lighttpd 1.4.46 which inherits ssl.* config from global scope if the $SERVER["socket"] block contains `ssl.engine = "enable"` but no other ssl.* directives.  Cheers.

PR #285, opened 2017-02-01
darkstat - Hardcode to HTTP, add optional custom hostname/IP address configuration

doktornotor:
Hardcoding HTTP still has effect no on people using HTTPS, due to the hardcoded HSTS header in nginx (https://redmine.pfsense.org/issues/6650). Was broken before with ```$myurl```, still exactly same broken now, no regression, except that the port is configurable.

As a workaround, I added a field where people can put their own hostname or IP used for the redirect, instead of the pfSense hostname. Pretty much required for everyone on HTTPS.


doktornotor:
OK, that's all I could figure out here.

doktornotor:
OK, that's all I could figure here...

PR #298, opened 2017-02-07
escape % in squid help text method

kcrawley:
fixes regression in pfSense haproxy package


kcrawley:
https://redmine.pfsense.org/issues/7232

kcrawley:
LA signed

rbgarga:
@kcrawley Please bump PORTVERSION or PORTREVISION on pfSense-pkg-haproxy/Makefile.

Also, did you check if pfSense-pkg-haproxy-devel have the same problem?

cc: @PiBa-NL

rbgarga:
Manually committed, thanks!

kcrawley:
@rbgarga did not fix in `devel`, was not aware of that package. a little covered up at the moment, but I'll look at it in the next 24 hrs.

PR #299, opened 2017-02-08
Moved pid unlink. Added Reasons for script. Added signals

marjohn56:
pid unlink was called before all processes were complete. This could leave dhcp6c still running sending and waiting for release but any check for process by PID returned false, moved to just before exit(0); This is probably the reason for cases of multiple dhcp6c instances. The original method of killing dhcp6c using SIGTERM, SIGTERM, SIGKILL  would fail as there was no pid by the time it got to SIGKILL.

Env var REASONS only returned NBI, now returns various reasons for script call, these can be:

REASON=INFO
REASON=REPLY
REASON=RENEW
REASON=RELEASE
REASON=REBIND
REASON=EXIT
REASON=OTHER

EXIT will be signalled when the client exits.
OTHER is a catchall and should never happen.

Signals SIGUSR1 and SIGUSR2 are added. These are used to force an exit ignoring the state of the no-release flag. SIGUSR1 forces a no-release exit. SGUSR2 forces a release exit. SIGUSER1 is very useful as
it kills dhcp6c quickly without sending and waiting for release on a WAN down event.

Logging info changes to text case and text; logging change to indicate PD or IA in info mode rather than debug.

This client will work without script updates being required.

### DIFF
--- a/net/dhcp6/patch-common.c
+++ b/net/dhcp6/patch-common.c
@@ -1,0 +1,11 @@
+--- common.c.orig	2016-12-19 08:16:42 UTC
++++ common.c
+@@ -3344,7 +3344,7 @@ ifaddrconf(cmd, ifname, addr, plen, plti
+ 	}
+ #endif
+ 
+-	d_printf(LOG_DEBUG, FNAME, "%s an address %s/%d on %s", cmdstr,
++	d_printf(LOG_INFO, FNAME, "%s an address %s/%d on %s", cmdstr,
+ 	    addr2str((struct sockaddr *)addr), plen, ifname);
+ 
+ 	close(s);
--- a/net/dhcp6/patch-config.h
+++ b/net/dhcp6/patch-config.h
@@ -1,0 +1,11 @@
+--- config.h.orig	2016-12-19 08:16:42 UTC
++++ config.h
+@@ -162,7 +162,7 @@ struct dhcp6_serverinfo {
+ 
+ /* client status code */
+ enum {DHCP6S_INIT, DHCP6S_SOLICIT, DHCP6S_INFOREQ, DHCP6S_REQUEST,
+-      DHCP6S_RENEW, DHCP6S_REBIND, DHCP6S_RELEASE, DHCP6S_IDLE};
++      DHCP6S_RENEW, DHCP6S_REBIND, DHCP6S_RELEASE, DHCP6S_IDLE, DHCP6S_EXIT};
+ 
+ struct prefix_ifconf {
+ 	TAILQ_ENTRY(prefix_ifconf) link;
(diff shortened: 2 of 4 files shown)


rbgarga:
I would like to hear from @hrs-allbsd about https://github.com/hrs-allbsd/wide-dhcpv6/pull/5 before proceed with this change

marjohn56:
So would I, it's been there a week.

bimmerdriver:
I took the liberty of contacting @hrs-allbsd. He will try to look at the PR and merge it on Friday.

bimmerdriver:
Friday came and went and the PR has still not been merged. I sent another email to @hrs-allbsd, but I haven't heard back yet. Considering that the WIDE / KAME projects shut down in 2005, it's not surprising that there have been virtually no changes to WIDE-DHCPv6 (other than another recent fix for pfsense a few months ago) for 10 or 11 years. However, there have been many revisions to the RFCs that were in place in 2005 when WIDE-DHCPv6 was released, as well as new RFCs. WIDE-DHCPv6 is a dead project comprising one person (@hrs-allbsd) and it's clearly not his priority. On the other hand, KEA DHCP is being actively developed and will replace ISC DHCP. Maybe pfsense should move to KEA DHCP.

marjohn56:
Kea is a v4 \ v6 server, i cannot see any mention of a dhcpv6 client; it also appears to be c++ which means using extra libraries in the same way as  Dibbler when i got that to work for Kev. At present there is not a really good v6 client. Dibbler is not bad, but has not been fully ported to *BSD and is missing some things that make it hard to use with pfSense, or any BSD system. 
Its not pfSense in this case that's the problem, its FreeBSD.


 
Sent from: YOGA Tablet 2-1050F 
On 26 Feb 2017 18:29, bimmerdriver <notifications@github.com> wrote:
Friday came and went and the PR has still not been merged. I sent another email to @hrs-allbsd, but I haven't heard back yet. Considering that the WIDE / KAME projects shut down in 2005, it's not surprising that there have been virtually no changes to WIDE-DHCPv6 (other than another recent fix for pfsense a few months ago) for 10 or 11 years. However, there have been many revisions to the RFCs that were in place in 2005 when WIDE-DHCPv6 was released, as well as new RFCs. WIDE-DHCPv6 is a dead project comprising one person (@hrs-allbsd) and it's clearly not his priority. On the other hand, KEA DHCP is being actively developed and will replace ISC DHCP. Maybe pfsense should move to KEA DHCP.

bimmerdriver:
I didn’t notice that. If so, it makes sense.

bimmerdriver:
@rbgarga, I still have not received any subsequent update from @hrs-allbsd. His original reply was that he was busy preparing for AsiaBSDCon, which is March 9-12. Is there a "Plan B"? The fix developed by @marjohn56 has been tested by several users over a long time and is working well. Can the code be patched from within pfsense?

rbgarga:
I've cherry-picked https://github.com/freebsd/freebsd-ports/commit/3a3ac4aa08839a6056bbf61fa5c2a8e63cefd5d2 to RELENG_2_3 and devel, it will be available on next round of snapshots.

PR #304, opened 2017-02-13
Allow NS/MX records to have hostnames with pfSense-pkg-bind9

geekmug:
Previously, the BIND package would blank out the hostname field and always use "@" for either MX or NS records. However, this is an arbitrary limitation. If you want to delegate a sub-domain, then you need to be able to provide a NS record with a hostname. Additionally, providing MX records for hosts is a valid record as well.

This patch continues to allow a blank hostname field (automatically replacing with "@"), so this should be backward compatible for existing users.

### DIFF
--- a/dns/pfSense-pkg-bind9/files/usr/local/pkg/bind.inc
+++ b/dns/pfSense-pkg-bind9/files/usr/local/pkg/bind.inc
@@ -67,9 +67,6 @@ function bind_zone_validate($post, &$input_errors) {
 							$input_errors[] = 'On reverse zones, valid record types are NS or PTR';
 						}
 					}
-					if (preg_match("/(MX|NS)/", $_POST["hosttype$i"])) {
-						$_POST["hostname$i"] = "";
-					}
 					if (!preg_match("/(MX|NS)/", $_POST["hosttype$i"]) && $_POST["hostname$i"] == "") {
 						$input_errors[] = 'Record cannot be empty for '.$_POST["hosttype$i"].' type ';
 					}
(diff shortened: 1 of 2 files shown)


pfSense-git-updates:

dns/pfSense-pkg-bind9/files/usr/local/pkg/bind.inc
```diff
@@ -513,7 +510,10 @@ EOD;
 							}
 						}
 						for ($y = 0; $y < sizeof($zone['row']); $y++) {
-							$hostname = (preg_match("/(MX|NS)/", $zone['row'][$y]['hosttype']) ? "@" : $zone['row'][$y]['hostname']);
+							$hostname = $zone['row'][$y]['hostname'];
+							if (preg_match("/(MX|NS)/", $zone['row'][$y]['hosttype']) && trim($hostname) == false) {
```
rbgarga: In which case you expect trim($hostname) to return false?

geekmug: The old code checked if the hosttype was MX or NS, then it would always substitute "@" for the hostname. The trim($hostname) check is so that this substitution only occurs if hostname was left empty, which makes this change backward compatible. Otherwise, the hostname provided by the user is kept.

rbgarga: According to trim docs it will never return false, only trimmed string.

http://php.net/manual/en/function.trim.php

geekmug: "==" is not a strict operator in PHP, so the string will be converted to a boolean value for the comparison, and any non-empty string converts to "true," otherwise "false" for an empty string. So, it is the case that ('' == false) is true, however, I see the prevailing style in this particular script is to test equality to an empty string, so I will change it.

PR #308, opened 2017-02-15
FreeRADIUS - Add input validation (Bug #7263)

doktornotor:
This adds input validation to FreeRADIUS package (Bug #7263).


doktornotor:
Wasn't done on purpose, because it'd result in conflicts with https://github.com/pfsense/FreeBSD-ports/pull/306.

rbgarga:
Yeah, I imagined :)

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -4409,4 +4415,834 @@ EOD;
...
+
+	// Session Timeout
+	if ($post['varuserssessiontimeout'] != '' && !is_numericint($post['varuserssessiontimeout'])) {
+		$input_errors[] = "The 'Session Timeout' field must contain an integer value.";
+	}
+
+	// Possible Login Times
+	// TODO: Produce some regex or better check here
+	if ($post['varuserslogintime'] && !preg_match("/^[a-zA-Z0-9,|-]*$/", $post['varuserslogintime'])) {
+		$input_errors[] = "The 'Possible Login Times' field may only contain a-z, A-Z, 0-9, comma, vertical bar and hyphen (regex /^[a-zA-Z0-9,|-]*$/)";
```
jim-p: You aren't going to split that up and validate each time entry in the list? ;-)

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -4409,4 +4415,834 @@ EOD;
...
+	}
+
+	// Maximum Bandwidth Up
+	if ($post['varmacsmaxbandwidthup'] != '' && !is_numericint($post['varmacsmaxbandwidthup'])) {
+		$input_errors[] = "The 'Maximum Bandwidth Up' field must contain an integer value.";
+	}
+
+	// Accounting Interim Interval
+	if ($post['varmacsacctinteriminterval'] != '') {
+		if (!is_numericint($post['varmacsacctinteriminterval'])) {
```
jim-p: I'd combine both of these tests and print the same message for either case, such as "The 'Accounting Interim Interval' field must contain an integer with a value larger than 60 (seconds)"

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -4409,4 +4415,834 @@ EOD;
...
+	}
+
+	// Maximum Requests Per Server
+	if ($post['varsettingsmaxrequestsperserver'] != '' && !is_numericint($post['varsettingsmaxrequestsperserver'])) {
+		$input_errors[] = "The 'Maximum Requests Per Server' field must contain an integer value.";
+	}
+
+	// OTP Lifetime
+	if ($post['varsettingsmotpenable'] == 'on') {
+		if ($post['varsettingsmotptimespan'] === 0) {
```
jim-p: If you are going to test using ===, you should cast to (int) or similar on the variable since it would be a string.

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -4409,4 +4415,834 @@ EOD;
...
+			$input_errors[] = "The 'OTP Lifetime' field must not be 0.";
+		} elseif (empty($post['varsettingsmotptimespan'])) {
+			$input_errors[] = "The 'OTP Lifetime' field must not be empty.";
+		} elseif (!is_numericint($post['varsettingsmotptimespan'])) {
+			$input_errors[] = "The 'OTP Lifetime' field must contain an integer value.";
+		} elseif ($post['varsettingsmotptimespan'] > 12) {
+			$input_errors[] = "The 'OTP Lifetime' field should contain only sane secure values. Values higher than 12 (~120 seconds) are not allowed.";
```
jim-p: I'd remove "sane" here, since "secure" gets the point across on its own.

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -4409,4 +4415,834 @@ EOD;
...
+		}
+	}
+
+	// Number of Invalid Password Attempts
+	if ($post['varsettingsmotpenable'] == 'on') {
+		if ($post['varsettingsmotppasswordattempts'] === 0) {
```
jim-p: Same as above about casting to (int)

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -4409,4 +4415,834 @@ EOD;
...
+			$input_errors[] = "The 'Invalid Password Attempts' field must not be 0.";
+		} elseif (empty($post['varsettingsmotppasswordattempts'])) {
+			$input_errors[] = "The 'Invalid Password Attempts' field must not be empty.";
+		} elseif (!is_numericint($post['varsettingsmotppasswordattempts'])) {
+			$input_errors[] = "The 'Invalid Password Attempts' field must contain an integer value.";
+		} elseif ($post['varsettingsmotppasswordattempts'] > 20) {
+			$input_errors[] = "The 'Invalid Password Attempts' field should contain only sane secure values. Values higher than 20 are not allowed.";
```
jim-p: Same as above, re: s/sane//

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -4409,4 +4415,834 @@ EOD;
...
+	 */
+
+}
+
+/* EAP settings input validation */
+function freeradius_validate_eap($post, &$input_errors) {
+
+	// Disable Weak EAP Types sanity check
+	if ($post['vareapconfdisableweakeaptypes'] == 'on') {
+		if ($post['vareapconfdefaulteaptype'] == 'md5' || $post['vareapconfdefaulteaptype'] == 'gtc' || $post['vareapconfdefaulteaptype'] == 'leap') {
```
jim-p: For three tests this could go either way but it would be easier/cleaner to use in_array here, e.g. `in_array($post['vareapconfdefaulteaptype'], array('md5', 'gtc', 'leap'))`.

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -4409,4 +4415,834 @@ EOD;
...
+			$input_errors[] = "When using the deprecated FreeRADIUS Cert Manager, the 'SSL Revocation List' field must be set to 'none'.";
+		}
+		if ($post['ssl_server_cert'] != 'none') {
+			$input_errors[] = "When using the deprecated FreeRADIUS Cert Manager, the 'SSL Server Certificate' field must be set to 'none'.";
+		}
+	}
+
+	// EAP-TLS Fragment Size
+	if ($post['vareapconffragmentsize'] != '') {
+		if (!is_numericint($post['vareapconffragmentsize'])) {
```
jim-p: I'd combine the test and print the same error for both, mentioning the type and limit.

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -4409,4 +4415,834 @@ EOD;
...
+}
+
+/* SQL settings input validation */
+function freeradius_validate_sql($post, &$input_errors) {
+
+	// SQL Server Address
+	if ($post['varsqlconfincludeenable'] == 'on') {
+		if (empty($post['varsqlconfserver'])) {
+			$input_errors[] = "The 'Server Address' field for SQL Server 1 must not be empty when 'SQL Support' for Server 1 is enabled.";
+		} elseif (!is_ipaddr($post['varsqlconfserver']) && !is_hostname($post['varsqlconfserver']) && !is_domain($post['varsqlconfserver'])) {
```
jim-p: is_hostname() also returns true for domain names (it uses is_domain() internally as part of its validation), so the is_domain() test isn't needed, just is_hostname() will catch short hostnames, FQDNs, and domain names.

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -4409,4 +4415,834 @@ EOD;
...
+			$input_errors[] = "The 'Server Address' field for SQL Server 1 must contain a valid FQDN or IP address.";
+		}
+	}
+	if ($post['varsqlconf2includeenable'] == 'on') {
+		if (empty($post['varsqlconf2server'])) {
+			$input_errors[] = "The 'Server Address' field for SQL Server 2 must not be empty when 'SQL Support' for Server 2 is enabled.";
+		} elseif (!is_ipaddr($post['varsqlconf2server']) && !is_hostname($post['varsqlconf2server']) && !is_domain($post['varsqlconf2server'])) {
```
jim-p: See above, the is_domain() test can be removed

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -4409,4 +4415,834 @@ EOD;
...
+
+/* LDAP settings input validation */
+function freeradius_validate_ldap($post, &$input_errors) {
+	global $config;
+
+	// LDAP Server Address
+	if ($post['varmodulesldapenableauthorize'] == 'on') {
+		if (empty($post['varmodulesldapserver'])) {
+			$input_errors[] = "The 'Server Address' field for LDAP Server 1 must not be empty when 'LDAP Authorization Support' for Server 1 is enabled.";
+		} elseif (!is_ipaddr($post['varmodulesldapserver']) && !is_hostname($post['varmodulesldapserver']) && !is_domain($post['varmodulesldapserver'])) {
```
jim-p: See above, re: is_domain() test

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -4409,4 +4415,834 @@ EOD;
...
+			$input_errors[] = "The 'Server Address' field for LDAP Server 1 must contain a valid FQDN or IP address.";
+		}
+	}
+	if ($post['varmodulesldap2enableauthorize'] == 'on') {
+		if (empty($post['varmodulesldap2server'])) {
+			$input_errors[] = "The 'Server Address' field for LDAP Server 2 must not be empty when 'LDAP Authorization Support' for Server 2 is enabled.";
+		} elseif (!is_ipaddr($post['varmodulesldap2server']) && !is_hostname($post['varmodulesldap2server']) && !is_domain($post['varmodulesldap2server'])) {
```
jim-p: See above, re: is_domain() test

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
doktornotor: The FreeRADIUS config syntax for similar things drives me mad, frankly... (Ditto the account expiry format, WTF is wrong with format like YYYY/MM/DD.)

net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
doktornotor: It doesn't work anyway, I guess it's impossible in PHP to distinguish between 0 and empty string in input. Will replace it with a single test for empty().

doktornotor:
@jim-p - Should be done. (No, I didn't split the login times string 🤣 )

As for gettext(), are packages getting translated now as well? Anyway, this should be done in the whole package, not just for the validation part, so yeah - lots of work and largely unrelated to this; leaving that for next time.

jim-p:
I'm not sure what the current plan of action is for translation in packages, but it seemed like low-hanging fruit since it was being modified anyhow. I'll look the changes over shortly.

PR #309, opened 2017-02-17
softflowd package cleanup, add privileges settings

doktornotor:
- Split PHP code to softflowd.inc
- Improve input validation
- Clean up descriptions, set required fields
- Add privileges settings
- Do not double the data in certain setups (https://redmine.pfsense.org/issues/4731#note-6)
- Add enable/disable toggle

BTW, while touching this, could anyone tell me WTH is that ```cleanup_config_softflowd()``` function doing?

### DIFF
--- a/net-mgmt/pfSense-pkg-softflowd/files/usr/local/pkg/softflowd.inc
+++ b/net-mgmt/pfSense-pkg-softflowd/files/usr/local/pkg/softflowd.inc
@@ -34,7 +34,7 @@ function sync_package_softflowd() {
 		$cf = array();
 	}
 	$interface_list = explode(",", $cf['interface']);
-	if (!empty($cf['host']) && !empty($interface_list) && $cf['enable_softflowd'] != "off") {
+	if (!empty($cf['host']) && !empty($interface_list) && $cf['enable'] != "off") {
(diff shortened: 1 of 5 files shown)


net-mgmt/pfSense-pkg-softflowd/files/usr/local/pkg/softflowd.inc
```diff
@@ -0,0 +1,163 @@
...
+	global $config;
+
+	if (is_array($config['installedpackages']['softflowd']['config'][0])) {
+		$cf = $config['installedpackages']['softflowd']['config'][0];
+	} else {
+		$cf = array();
+	}
+	$interface_list = explode(",", $cf['interface']);
+	if (!empty($cf['host']) && !empty($interface_list) && $cf['enable'] != "off") {
+		conf_mount_rw();
```
rbgarga: Instead of call conf_mount_rw() inside if and else statements, just move it to a single call outside

doktornotor: Moved. The thing is completely no-op, why's it even still being used...

PR #312, opened 2017-02-23
acme, store certificate after renew action with dns-manual

PiBa-NL:


idarlund:
tested. certs now gets pushed to Certificate Manager.

PR #318, opened 2017-02-28
acme, update to upstream acme.sh 2.6.7 version

PiBa-NL:


jim-p:
You need to be sure you aren't clobbering changes we've fixed. In particular, this change reverts my fixes in acme/dnsapi/dns_nsupdate.sh  that allow it to work for multiple entries, and probably other changes.

PiBa-NL:
Thanks for checking :) . Seems like that was the only change made to 'native' acme.sh files that needed to stay around.
Re-pushed the commit without undoing the nsupdate changes. Hope its alright like this.

jim-p:
I'll check it over shortly. I'm curious if my change might be viable to upstream, but it may depend too much on our other code to make sense for them as-is.

idarlund:
#312 should also be included in this update. both are marked as .13

PR #321, opened 2017-03-03
acme, bump version to 0.1.14 to get previous change into the package

PiBa-NL:


jim-p:
Is something missing from the package? This shouldn't be necessary since I merged both of those other PRs at the same time, the same bump should have brought in everything.

doktornotor:
@jim-p It's on 0.1.13 on 2.3.4 snapshots.

jim-p:
2.3.4 and 2.4 only get new package versions when a new snapshot compiles and uploads, so it might not show up yet there. It will whenever a new snap is made.

doktornotor:
Ah, OK. Thanks.

PR #324, opened 2017-03-06
Updated with quotation marks

craCH:
Because if you have a password with non standard characters (like !) you need some quotations marks around it. Otherwise the service breaks with a syntax error.

### DIFF
--- a/net/pfSense-pkg-OpenBGPD/files/usr/local/pkg/openbgpd.inc
+++ b/net/pfSense-pkg-OpenBGPD/files/usr/local/pkg/openbgpd.inc
@@ -124,10 +124,10 @@ function openbgpd_install_conf() {
 							$conffile .= "\tneighbor {$neighbor['neighbor']} {\n";
 							$conffile .= "\t\tdescr \"{$neighbor['descr']}\"\n";
 							if ($neighbor['md5sigpass']) {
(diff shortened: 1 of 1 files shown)


pfSense-git-updates:

jim-p:
Unless this changed recently that probably won't work. If both sides are OpenBGPD it may appear to work because the passwords still match but if you try to connect to something else, it considered the surrounding quotes as part of the password itself. That was some time ago but IIRC that's why we don't quote it currently. It would fail against other vendors like Cisco in those cases. If that has been fixed then it could be considered.

If you can confirm it works with single quotes, use escapeshellarg(). Using double quotes around it manually could break if the password itself contains a double quote. Otherwise you'd have to find `"` in the string and escape it, assuming OpenBGPD allows that.

craCH:
We've connected to a Cisco ASR9k today ... no connection without quotes.

jim-p:
That doesn't answer the question sufficiently.

* Is the password also quoted on the Cisco side?
* If you set both to a password without special characters, does it still connect when only the OpenBGPD side is quoted (with and without quotes, actually)?
* Does it work the same with single quotes as double quotes?

craCH:
The password isn't quoted on the cisco side (that's what the config engeneer said) ... but I'm not able to do these tests. Sorry.

rbgarga:
4 months has passed and CLA was not signed. Closing.

PR #327, opened 2017-03-08
Add AJAX to Squid AV widget

doktornotor:


www/pfSense-pkg-squid/files/usr/local/www/widgets/widgets/squid_antivirus_status.widget.php
```diff
@@ -159,3 +161,26 @@ function squid_antivirus_statistics() {
...
+		url: '/widgets/widgets/squid_antivirus_status.widget.php',
+		dataType: 'html',
+		dataFilter: function(raw){
+			// We reload the entire widget, strip this block of javascript from it
+			return raw.replace(/<script>([\s\S]*)<\/script>/gi, '');
+		},
+		success: function(data){
+			$('#squidav_status').html(data);
+		}
+	});
```
sbeaver-netgate: It's good practice to include an error handler. In this case it could return an HTML error message?

doktornotor:
Not exactly sure what it should show there; what kind of error you mean?

sbeaver-netgate:
If the AJAX call is successful, squidav_status is updated and all is well, but if for some reason the AJAX call does not succeed, nothing happens. An jQuery AJAX call can report an error, but we don't do anything about it, and it seems like we should :)

Perhaps:
`   
    success: function(data){
        $('#squidav_status').html(data);
    },
    error: function() {
        $('#squidav_status').html('<div class="alert alert-danger">Unable to retrieve status</div>');
    }`

doktornotor:
Well OK, there's a bunch of Ajax calls across the pfSense code, pretty much none of those does anything on failure. The one here is very much a verbatim copy of the [gmirror widget code](https://github.com/pfsense/pfsense/blob/master/src/usr/local/www/widgets/widgets/gmirror_status.widget.php#L36)

Perhaps you could do some mass commits there, I'm definitely much less familiar with JS.

sbeaver-netgate:
Yes. Whenever I work on them I try to add an error handler, but I haven't got to many of them yet. Whenever we add new code, I think it should it should use error handling.

PR #330, opened 2017-03-11
Suricata improvements

securitym0nkey:
This are some improvements  for the suricata SID content mgmt.

It resolves 2 bugs in the SID content mgmt:

- Only the last SID in a list get modified
- Modification to all SIDs not working. Neither via the wildcard '*' nor with just the search term and replacement as showed in the example config.

Additionally it adds support to modify all signatures from a certain category

### DIFF
--- a/security/pfSense-pkg-suricata/files/var/db/suricata/sidmods/modifysid-sample.conf
+++ b/security/pfSense-pkg-suricata/files/var/db/suricata/sidmods/modifysid-sample.conf
@@ -1,7 +1,7 @@
 # example modifysid.conf
 #
 # formatting is simple
-# <sid or sid list> "what I'm replacing" "what I'm replacing it with"
+# <sid, category, list of sids&categories> "what I'm replacing" "what I'm replacing it with"
 #
 # Note that this will only work with GID:1 rules, simply because modifying
 # GID:3 SO stub rules would not actually affect the rule.
@@ -19,5 +19,13 @@
 # "HTTP_PORTS" "HTTPS_PORTS"
 
 # multiple sids can be specified as noted below:
-# 302,429,1821 "\$EXTERNAL_NET" "\$HOME_NET"
+# 302,429,1821 "$EXTERNAL_NET" "$HOME_NET"
 
+# modify all signatures in a category. Example: replace "$EXTERNAL_NETS" with "any" to be alerts on insider threats as well
+# emerging-scan "$EXTERNAL_NET" "any"
+
+# modify all signatures in multiple categories
+# emerging-scan,emerging-sql "$EXTERNAL_NET" "any"
+
+# modify all signatures for a category and specific SIDs from other categories
+# emerging-sql,2100691,2009817 "$EXTERNAL_NET" "any"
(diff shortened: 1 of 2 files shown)


pfSense-git-updates:

security/pfSense-pkg-suricata/files/var/db/suricata/sidmods/modifysid-sample.conf
```diff
@@ -19,5 +19,13 @@
 # "HTTP_PORTS" "HTTPS_PORTS"
 
 # multiple sids can be specified as noted below:
-# 302,429,1821 "\$EXTERNAL_NET" "\$HOME_NET"
+# 302,429,1821 "$EXTERNAL_NET" "$HOME_NET"
```
doktornotor: Pretty sure these should remain escaped as it was before, and any added example should have those escaped, otherwise it's never going to match things in the rulesets.

security/pfSense-pkg-suricata/files/var/db/suricata/sidmods/modifysid-sample.conf
```diff
@@ -19,5 +19,13 @@
...
 
+# modify all signatures in a category. Example: replace "$EXTERNAL_NETS" with "any" to be alerts on insider threats as well
+# emerging-scan "$EXTERNAL_NET" "any"
+
+# modify all signatures in multiple categories
+# emerging-scan,emerging-sql "$EXTERNAL_NET" "any"
+
+# modify all signatures for a category and specific SIDs from other categories
+# emerging-sql,2100691,2009817 "$EXTERNAL_NET" "any"
```
doktornotor: See previous comment.

security/pfSense-pkg-suricata/files/var/db/suricata/sidmods/modifysid-sample.conf
securitym0nkey: The code is escaping it itself. See line 2506. Actually that way you can't use regex for replacement. pulledpork supports replacement by regex in such case you need to escape it. I thought of removing the preg_quote so that the current example would be correct. But dropped that idea to no break the config for people who have noticed that escaping is not needed as displayed in the example.

rbgarga:
Can you take a look at these changes @bmeeks8 ?

security/pfSense-pkg-suricata/files/var/db/suricata/sidmods/modifysid-sample.conf
bmeeks8: Yeah, currently the code performs internal escaping of the regex.

bmeeks8:
I am OK with the changes submitted.  Thank you to the submitter for improving the Suricata package!

rbgarga:
@securitym0nkey can you please rebase your fork and bump PORTREVISION so I can get it merged

rbgarga:
@bmeeks8 if you believe it's easier for you to include these changes together with the update you are working on just let me know

bmeeks8:
I just submitted the pull request for the new Suricata 3.2.1 package update, but it was before I saw this notice.  My update is not touching the files referenced in this pull request.

rbgarga:
Manually merged it, thanks!

PR #333, opened 2017-03-16
Don't delete safebrowsing definitions when manual configuration is enabled

doktornotor:


www/pfSense-pkg-squid/Makefile
```diff
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-squid
 PORTVERSION=	0.4.36
-PORTREVISION=	1
+PORTREVISION=	2
```
rbgarga: package is currently in PORTREVISION 2, could you please rebase your fork and bump it to 3?

doktornotor:
Not sure whether it worked, can just use the GitHub web GUI from a tablet ATM.

PR #336, opened 2017-04-02
pfSense-pkg-net-snmp: Fix services

dbaio:
- ~~Service name must not use hyphens to not get conflicted with [pfSenseHelpers.js](https://github.com/pfsense/pfsense/blob/master/src/usr/local/www/js/pfSenseHelpers.js#L658-L659)~~
- Fix a typo in `snmptrapd` executable name
- Bump PORTREVISION

### DIFF
--- a/net-mgmt/pfSense-pkg-net-snmp/Makefile
+++ b/net-mgmt/pfSense-pkg-net-snmp/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-net-snmp
 PORTVERSION=	0.1.2
-PORTREVISION=	1
+PORTREVISION=	2
 CATEGORIES=	net-mgmt
 MASTER_SITES=	# empty
 DISTFILES=	# empty
--- a/net-mgmt/pfSense-pkg-net-snmp/files/usr/local/pkg/net-snmp.xml
+++ b/net-mgmt/pfSense-pkg-net-snmp/files/usr/local/pkg/net-snmp.xml
@@ -49,7 +49,7 @@
 	<service>
 		<name>net-snmptrapd</name>
 		<rcfile>net-snmptrapd.sh</rcfile>
-		<executable>snmpdtrapd</executable>
+		<executable>snmptrapd</executable>
 		<description><![CDATA[NET-SNMP Trap Listening Daemon]]></description>
 	</service>
 	<tabs>


doktornotor:
Not exactly sure since when does hyper conflict with a piece of JS but it's seem whole lot better to me to fix that broken piece of JS.

dbaio:
Maybe since [this commit](https://github.com/pfsense/pfsense/commit/74b0f23e25caa7f0495be4d4eae152b9bf241d5d) and [this one](https://github.com/pfsense/pfsense/commit/29fda3d3cea4bd800568f82a1203db8cd13a82e2).

It is used in Status / Services page:

```
<a href="#" id="restartservice-net_snmptrapd" ><i class="fa fa-repeat" title="Restart Service"></i></a>
<a href="#" id="stopservice-net_snmptrapd"><i class="fa fa-stop-circle-o" title="Stop Service"></i></a>
<a href="#" id="restartservice-zabbix_agentd_lts" ><i class="fa fa-repeat" title="Restart Service"></i></a>
<a href="#" id="stopservice-zabbix_agentd_lts"><i class="fa fa-stop-circle-o" title="Stop Service"></i></a>
```

[Status/Services --> Ajax](https://github.com/pfsense/pfsense/blob/master/src/usr/local/www/status_services.php#L33-L54)
[Status/Services](https://github.com/pfsense/pfsense/blob/master/src/usr/local/www/status_services.php#L117)
[service-utils.inc --> get_service_control_links](https://github.com/pfsense/pfsense/blob/master/src/etc/inc/service-utils.inc#L532-L545)

That code can be fixed, yes, but changing the service name is also valid to keep services working even in not updated machines.

doktornotor:
This also breaks (at least)
- [ftp-proxy](https://github.com/pfsense/FreeBSD-ports/blob/devel/ftp/pfSense-pkg-FTP_Client_Proxy/files/usr/local/pkg/ftpproxy.xml#L40)
- Squid [c-icap](https://github.com/pfsense/FreeBSD-ports/blob/devel/www/pfSense-pkg-squid/files/usr/local/pkg/squid.xml#L57)

and in general introducing regressions with similar totally undocumented, arbitrary restrictions suck goats nuts, to put it mildly.

dbaio:
There is no harm in changing the service names for this package because it's a new package under testing. 
But if you (and @jim-p) don't agree I can withdraw that commit and let just the typo part on the executable name.

And very nice that you opened that bug on redmine, thank you.

jim-p:
It's better not to mix changes in PRs anyhow. The typo fix should be separate.

I'd also prefer to see that service name limitation fixed, if possible, rather than rename the service.

dbaio:
ok @jim-p, done.

doktornotor:
Should be fixed with https://github.com/pfsense/pfsense/commit/21a8edd7836baf1e4ee2f347fdc1bafb4d96c9d9 so this one shouldn't be needed any more.

PR #338, opened 2017-04-05
Openvpn-client-export, add the support of openvpn up and down script

hugosxm:
Hi,

hope i write this to the right place. Someone on the IRC suggested me to post my idea here. 

Here is the official documentation of openvpn-gui : 

Run Connect/Disconnect/Preconnect Scripts

====> There are three different scripts that OpenVPN GUI can execute to help with different tasks like mapping network drives.

Preconnect If a file named "xxx_pre.bat" exist in the config folder
where xxx is the same as your OpenVPN config file name, this will be executed BEFORE the OpenVPN tunnel is established.
Connect If a file named "xxx_up.bat" exist in the config folder
where xxx is the same as your OpenVPN config file name, this will be executed AFTER the OpenVPN tunnel is established.
Disconnect If a file named "xxx_down.bat" exist in the config folder
where xxx is the same as your OpenVPN config file name, this will be executed BEFORE the OpenVPN tunnel is closed.   <====

As pfsense can create a executable installer with the GUI, the config and the cert/key file, the idea is the add the possibility to bundle up and down script in the installer provided by pfsense.

I did a small piece of code, in php, first time writing in php for me, sorry... Wich read the content from some template file and write it to a file, with the same name as config, like explained up here. 

[php_patch.txt](https://github.com/pfsense/FreeBSD-ports/files/896359/php_patch.txt)

the two file are then bundled with the config and certs. 

I also modified the openvpn-postinstall.exe so it will copy the new file, i just added this to the NSI and recompiled the thing : 

"

	DetailPrint "Installing network drive scripts"
	${Locate} ".\config" "/L=F /M=*.bat" "CopyConfFile"
	${Locate} ".\config" "/L=F /M=*.vbs" "CopyConfFile"

"

I also have this little vbs script that is used to wait some second before trying to map a drive for exemple. I copy it along with the script and the config / certs files.

[sleep.txt](https://github.com/pfsense/FreeBSD-ports/files/896373/sleep.txt)

Now the openvpn client automaticatly run the scripts on connect and disconnect to map the drives.

I think this is a really USEFULL feature for pfsense ! 

Maybe we can add two text form to the openvopn-client-export page, then it will be possible to past in the form the content of the up_script and down_script. 

If the form are empty, there is no script to copy. 

And i think the sleep.vbs script has to be bundled with the script cause everyone that map a drive will be using this...

Sorry for my bad, i am not a dev, hope it will help !


jim-p:
Something is very broken with this PE and it's also not an appropriate place to post a feature request. Start by opening a feature request on https://redmine.pfsense.org/

Though I do not see that sort of feature being added to the export package. That goes beyond the intended purpose of the package.

PR #339, opened 2017-04-10
Add UPS Power (in Watts) to the UPS Status page

oobayly:
As the title suggests - this adds a new row that shows how many Watts are being drawn from the UPS.
This will differ from the current UPS Load percentage, as that is calculated in VA, not Watts. I'm open to discussion whether real power (W) or apparent power should be used.

### DIFF
--- a/sysutils/pfSense-pkg-nut/files/usr/local/www/nut_status.php
+++ b/sysutils/pfSense-pkg-nut/files/usr/local/www/nut_status.php
@@ -31,15 +31,20 @@ function print_row($desc, $value) {
 	print '</tr>';
 }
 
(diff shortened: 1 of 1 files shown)


pfSense-git-updates:

dennypage:
UPS power is already displayed on this page in the UPS Detail section (assuming the UPS provides it). Scroll down and you should see it.

FWIW, the reason UPS power is not in the top section is because a) it's rare compared to UPS load which is almost universal, and b) it's not used as part of the runtime calculation. The top section and the widget contain only those items pertaining to failure (quality and runtime).

oobayly:
Fair enough

PR #344, opened 2017-04-23
FreeRADIUS - put CRL on a new line (Bug #7491), add some messages for missing certificates

doktornotor:
Also added some messages and input validation regarding EAP certificates here (see https://redmine.pfsense.org/issues/7479).

### DIFF
--- a/net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
+++ b/net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
@@ -1126,7 +1126,7 @@ function freeradius_eapconf_resync($restart_svc = true) {
 			} else {
 				$check_crl="check_crl = no";
 			}
-			file_put_contents(FREERADIUS_ETC . "/raddb/certs/ca_cert.pem", base64_decode($ca_cert['crt']). $crl);
+			file_put_contents(FREERADIUS_ETC . "/raddb/certs/ca_cert.pem", base64_decode($ca_cert['crt']) . "\n" . $crl);
 			$conf['ssl_ca_cert'] = FREERADIUS_ETC . "/raddb/certs/ca_cert.pem";
 		}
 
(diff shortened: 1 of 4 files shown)


net/pfSense-pkg-freeradius2/files/usr/local/pkg/freeradius.inc
```diff
@@ -4517,6 +4531,15 @@ function freeradius_validate_eap($post, &$input_errors) {
 		$input_errors[] = "The 'Maximum Sessions Tracking Per Server' field must contain an integer value.";
 	}
 
+	// Certificates for TLS
+	if ($post['ssl_ca_cert'] == 'none') {
```
jim-p: If "None" is not valid why even offer it in the drop-down? Does that break anything?

doktornotor: Well probably because showing random first thing returned by the lookup did not seem like the best idea around? (Generic problem with these drop-downs in pfS.) Been there for ages, personally I have no clue what's been the idea with this. The package has never worked without certificates, originally it's been bootstrapping some self-signed one there using the removed bundled certificate "manager". 

It could be renamed to "not configured" or even " " but there needs to be some default empty placeholder, otherwise users will end up saving a random cert/CA combo that doesn't match each other.

jim-p: Ah, OK. I didn't know it was generating its own if 'None' was selected before. 

If these are a hard requirement now, I think I'd prefer to see the TLS settings moved to the Settings page. Or maybe have "None" trigger generation and selection of a new self-signed cert entry in the cert manager, to bridge the old/new behavior. When they return to the page it would show something like "FreeRADIUS-auto-<refid>" for the name, similar to the automatic GUI cert. That seems like the path of least resistance for users. That or maybe fall back to selecting the same cert as the GUI?

Either way I'd prefer not to make the user go out of their way if we can avoid it. I can work on that if you'd like, but it might be early next week before I can get to it. We can import this in the meantime.

doktornotor: I've been thinking about grabbing the code from the stunnel package to generate a self-signed certificate that shows in the pfSense cert manager on install (https://github.com/pfsense/FreeBSD-ports/blob/devel/security/pfSense-pkg-stunnel/files/usr/local/pkg/stunnel.inc#L90), but yeah, I'd rather get out something that at least points users to the required configuration now and polish things later.

Definitely a good idea to move the settings from the EAP tab, kinda weird place for that.

PR #350, opened 2017-05-05
Make the HAproxy widget settings panel open

phil-davis:
The id for the settings panel-footer changed its format a bit to support multiple copies of a widget on the dashboard.


phil-davis:
See redmine issue https://redmine.pfsense.org/issues/7523

phil-davis:
Note: This change goes with pfSense 2.4 and later. It is not for use on 2.3.* systems.

rbgarga:
@phil-davis just to make it easier to maintain, can you do a change that check pfSense version and work both on 2.3 and 2.4?

phil-davis:
@rbgarga I discovered that even better is to just use the $widgetname var, which I had already made upwardly compatible with the dashboard changes in 2.4
See #352

PR #351, opened 2017-05-06
Make the HAproxy-devel widget settings panel open

phil-davis:
The id for the settings panel-footer changed its format a bit to support multiple copies of a widget on the dashboard.


phil-davis:
Note: This change goes with pfSense 2.4 and later. It is not for use on 2.3.* systems.

rbgarga:
@phil-davis just to make it easier to maintain, can you do a change that check pfSense version and work both on 2.3 and 2.4?

phil-davis:
@rbgarga see #353

PR #352, opened 2017-05-09
Make haproxy widget settings open

phil-davis:
pfSense 2.4 supports potentially multiple copies of a widget on the dashboard. For this, the widget name is no longer just like "haproxy", but is "haproxy-0" (and could be "haproxy-1" etc in future, if haproxy wanted to allow multiple copies).

Make haproxy use the passed-in value of $widgetname to generate the id of the widget panel footer. That works in both 2.3.* and 2.4.


rbgarga:
@phil-davis can you rebase your fork please?

phil-davis:
To master? I see that devel is the default branch of this repo.

rbgarga:
just rebase devel. JimP did a commit that already bumped PORTREVISION of haproxy

phil-davis:
Should be good to go now.

PR #354, opened 2017-05-13
Redmine #7548 Add absolute offset to NTP stats

phil-davis:
Plot and list the absolute value of the NTP time offset, as well as the existing plus/minus value.
See Redmine https://redmine.pfsense.org/issues/7548
Forum https://forum.pfsense.org/index.php?topic=130468.0
for details.

In order to do add this, it was easier to also refactor the way the index of the output array was kept track of. Seems to be simpler to understand now. This whole code could do with refactoring - I added more duplicated code to what was already there. But I am not starting down that path right now.

### DIFF
--- a/sysutils/pfSense-Status_Monitoring/files/usr/local/www/rrd_fetch_json.php
+++ b/sysutils/pfSense-Status_Monitoring/files/usr/local/www/rrd_fetch_json.php
@@ -58,11 +58,11 @@ $rrd_info_array = rrd_info($rrd_location . $right . ".rrd");
 $right_last_updated = $rrd_info_array['last_update'];
 
 //grab the older last updated time of the two databases
-if(empty($right_last_updated)) {
+if (empty($right_last_updated)) {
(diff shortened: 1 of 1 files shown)


NOYB:
The fix and other changes should be separate commits and/or PR.
What is the effect of removing the left and right adjusts?

phil-davis:
All those left/right/adjusts variables were just counting up an array index. $ds_key_out does the job now with 1 variable that gets incremented at the appropriate places.
It was not going to be easy to patch in more left/right/adjust variable fiddles in order to add the new row of data, and would have been another step of added confusion.

NOYB:
Thank you Phil.  The separate commits makes it much easier to follow and see what each what does.  Really appreciate it.  Thank you.

phil-davis:
Once I started on the separate commits, I thought "what the heck" and went all the with the refactoring. So the other PR has a lot (more,better) refactored code than here.

PR #355, opened 2017-05-14
Redmine #7548 Add absolute offset to NTP stats

phil-davis:
Plot and list the absolute value of the NTP time offset, as well as the existing plus/minus value.
See Redmine https://redmine.pfsense.org/issues/7548
Forum https://forum.pfsense.org/index.php?topic=130468.0
for details.

In order to do add this, it was easier to also refactor the whole code to get rid of the various counters and huge block of duplicated code for left and right. Those refactorings are in the various commits here, before the last one that implements the actual feature.

### DIFF
--- a/sysutils/pfSense-Status_Monitoring/files/usr/local/www/rrd_fetch_json.php
+++ b/sysutils/pfSense-Status_Monitoring/files/usr/local/www/rrd_fetch_json.php
@@ -292,6 +292,34 @@ foreach ($side as $settings) {
 			}
 
 			$obj[] = $entry;
+
+			if ($ds == 'offset') {
+				// Make an entry for the absolute value of NTP time offset
+				// Start with the existing entry and just modify it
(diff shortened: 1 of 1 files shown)


rbgarga:
Can you please have a look @jdillard ?

phil-davis:
And let me know which part of the version/portrevision should be bumped for a change like this.

jdillard:
Thanks, @phil-davis! I had to merge this one manually, in e4c9e5f47bfd379d6691ca736351fdd9631b9d4b, because of the `unknown repository` issue. I also revved the version to 1.7.0, since it added a feature and was refactored (much appreciated).

phil-davis:
Yeh, I forgot all about this it has been there so long. Last week I cleaned up the large number of repo forks that I had from various projects - removing the ones that I do not use very often. Glad that GitHub does not delete the code waiting in PRs.

PR #358, opened 2017-05-22
Update Snort binary to 2.9.9.0_1

bmeeks8:
This updates the Snort binary to version 2.9.9.0_1 to stay in sync with upstream.


rbgarga:
Manually cherry picked update + spoink patch. Thanks!

PR #359, opened 2017-05-25
Avoid e-mail spam on NUT communication errors

nagyrobi:
With many UPS models (especially some APC models like RS800), sometimes NUT misinterprets the communication link status, it changes from COMMOK to COMMBAD and vice-versa multiple times a second, and this can continue for several seconds. When one has e-mail notifications enabled, spam occurs, about a dozen of e-mails are thorwn to the user at once. This change removes COMMOK and COMMBAD notification flags from the e-mail notification command. These flags will still cause them to be logged, they are only removed from the e-mailing command.


dennypage:
Suppressing emails for communication errors across the board is not the way to handle flakey UPS or USB connections. Before you do anything else, please look at the NUT tunables.

nagyrobi:
Thanks jim-p, done.

nagyrobi:
@dennypage, I agree with your comments, though there's no such "tunable" in NUT configuration, as far as I could see. There is "upssched", but that is not really implemented yet in this package, and that won't supress these message, it would only make the system send you one message instead of many, and with a configurable delay, but you'd still get them.

The best option would be, IMHO, to expose all the NOTIFYFLAG options in the web interface to the user, and allow custom selections via checkboxes for each flag. But that's a whole different story.

Another option maybe to parse the contents of "Additional configuration lines for upsmon.conf" from the web interface and match them with the contents of nut.inc, so that double configuration items would not appear - but that's not how it works now. Currently all the configs entered here are added to what's in nut.inc - which is pretty wrong, since nut.inc already contains most of what upsmon.conf can ever contain. Any config line entered here in the web interface is just simply added to upsmon.conf - resulting in duplicate config items.

This is just a "quick and dirty fix" for now, until proper web interface configuration will be created either way. It's not just me affected.

dennypage:
This is not a quick and dirty fix. For most users this change would introduce a serious defect. 

You have a flakey ups or a flakey USB connection. If you are unable to fix this core issue, please explore the common NUT work-arounds. Look at the polling interval.

nagyrobi:
Sorry but pollinterval is not something that could be effective here.

dennypage:
Have you tried it? It's commonly used to help flakey connections.

dennypage:
While we're at it, some things you should try to address your core issue:

- Try a different USB port. Pay attention to whether the port is native 1.1 or bridged. Native ports generally works better than bridged. If bridged, see if you have ports on a different bridge chip. 

- USB hubs usually cause problems with UPS units. However, if the problem is a flakey motherboard bridge chip, they can actually help by introducing a different bridge chip. You can also buy cables with built-in bridge chips, but a hub is usually cheaper. Be sure to use a powered one. 

- Try a different USB cable. Higher quality.

dennypage:
Fwiw, I looked around and didn't find people complaining about issues using the rs-800 with nut, so I would assume that the problem is with the cable, motherboard or USB driver.

jim-p:

Thanks!

rbgarga:
No answers from submitter since last @jim-p comment on last July. If you decide to work on it again please re-open or re-submit

PR #360, opened 2017-05-25
Allow Quagga to run BGP with MD5

quadrinary:
Hi all - 

The modifications here reflect what we've been doing in our environment since last fall. 

Key features:
1. BGP MD5 password implementation at the OS level
2. View and easily compare running and saved raw configs


quadrinary:
Good deal. Let me know if you have any questions, it's been running stable for us since the beginning of March on 12 separate instances.

quadrinary:
This being my first direct contribution to pfSense, what happens next? 

Thanks!

quadrinary:
Hey all - I've made a significant change since this last pull and implemented a CARP failover mode that shifts OSPF cost on selected interfaces instead of just shutting down the service. The option still exists to simply shutdown the plugin as  before, but now users are presented with a choice, with the benefit that standby routers are still fully part of OSPF and can be reached via routes learned, not just static routes or NAT tricks.  [Another user outlined this a few years back ](https://forum.pfsense.org/index.php?topic=99897.0) and after testing this simple method for a couple weeks, i decided it best to approach it programatically from inside the Quagga plugin itself with (imho) pretty great outcomes.

PR #362, opened 2017-05-27
Avoid e-mail spam on NUT communication errors #359

nagyrobi:
Bump revision number because of: https://github.com/pfsense/FreeBSD-ports/pull/359/files


rbgarga:
Please do it in the same PR adding a new commit

PR #366, opened 2017-06-06
Ntopng config UI changes to allow arbitrary local networks list

madpilot78:
ntopng allows an arbitrary list of networks to be considered as "local".

I've added ability to configure such an arbitrary list to the pfsense ntopng UI.

To accommodate for such options I have reorganised a little the UI layout.

I'm submitting it hoping it can be accepted in your official tree. If any changes in UI layout or functionality is required please comment on this pull request.

Thanks!


net/pfSense-pkg-ntopng/files/usr/local/pkg/ntopng.inc
```diff
@@ -121,6 +121,17 @@ function ntopng_sync_package() {
 				$local_networks = "--local-networks " . escapeshellarg(gen_subnet(get_interface_ip('lan'), get_interface_subnet('lan')) . '/' . get_interface_subnet('lan'));
 			}
 			break;
+		case "custom":
+			$nets = array();
+			foreach ($ntopng_config['row'] as $net) {
+				if (is_subnet($net['cidr'])) {
+					$nets[] = $net['cidr'];
```
jim-p: To be on the safe side, you should use `trim()` around the right side, to be absolutely certain there is no whitespace around any of the values.

madpilot78: You're right. I should have thought about this right away.

Fixing it shortly.

madpilot78:
I noticed the pfSense-pkg-ntopng package version has already been bumped to 0.8.7.

Should I add a commit bumping it further to 0.8.8?

jim-p:
Yes, you'll want to rebase and then bump it again.

madpilot78:
Oops, I just wanted to rebase and bump portrevision but made a git mess somehow.

Not sure how to fix it.

Any suggestions?

doktornotor:
Perhaps someone could commit this before it gets messed up again? 😛

dennypage:
Guys, there's a couple of bugs here. Even though the custom networks option is not selected, the custom networks input field is still shown. More significantly, the custom networks input field is checked for validity even if the custom networks option is not selected. An empty list is not considered valid. This means that everyone's prior config from 0.8.8 is considered invalid and they cannot re-save a configuration without entering a valid cidr.

dennypage:
And yes, putting a valid CIDR in the field does allow the config to be saved. I've only tested a couple, but a common one, "192.168.0.0/16" works fine.

madpilot78:
Just for the record, I have submitted pull request #372 which contains a fix for this issue.

PR #368, opened 2017-06-24
NRPE: Allowing user to define custom shell commands

mintsoft:
Hello,

Currently there's no way of easily extending the monitoring within PFSense, if it's not exposed by an existing check_* or via SNMP it's very difficult to be able to check from external monitoring like Nagios. 

If the user has the ability to inject a shell command that gets dropped into the nrpe.conf then you can define new checks pretty easily by adding commands like  `/bin/tcsh -c 'grep ok /path/to/the/statefile/here && exit 0'` into the GUI. It's not exactly the easiest thing for people to do, but it does at least add a seam for extension.

Feedback welcome!

Thanks,
Rob

### DIFF
--- a/net-mgmt/pfSense-pkg-nrpe/files/usr/local/pkg/nrpe2.inc
+++ b/net-mgmt/pfSense-pkg-nrpe/files/usr/local/pkg/nrpe2.inc
@@ -208,7 +208,7 @@ function nrpe2_custom_php_service() {
 function nrpe2_get_commands() {
 	$nagios_check_path = NRPE_BASE . "/libexec/nagios";
 	$commands = glob("{$nagios_check_path}/check_*");
-	$cmdarr = array();
+	$cmdarr = array(array("command" => "/bin/tcsh -c "));
 	foreach ($commands as $cmd) {
 		$cmdarr[]["command"] = basename($cmd);
 	}
@@ -230,5 +230,3 @@ function nrpe2_custom_php_validation_command($post, &$input_errors) {
 		}
 	}
 }
-
-?>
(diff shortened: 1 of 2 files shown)


pfSense-git-updates:

mintsoft:
ICLA just signed

mintsoft:
@rbgarga @jim-p ping! Any chance of a review on this?

net-mgmt/pfSense-pkg-nrpe/files/usr/local/pkg/nrpe2.inc
```diff
@@ -154,8 +154,14 @@ function nrpe2_custom_php_write_config() {
 		$sudo = (isset($cmd['sudo']) && is_executable($sudo_bin)) ? "{$sudo_bin} " : "";
 		$wcmd = !empty($cmd['warning']) ? "-w {$cmd['warning']}" : "";
 		$ccmd = !empty($cmd['critical']) ? "-c {$cmd['critical']}" : "";
-		if (is_executable("{$nagios_check_path}/{$cmd['command']}")) {
-			$cmds[] = "command[{$cmd['name']}]={$sudo}{$nagios_check_path}/{$cmd['command']} {$wcmd} {$ccmd} {$cmd['extra']}\n";
+
+		$path_to_command = "{$nagios_check_path}/{$cmd['command']}";
+		$is_custom_shell_command = ($cmd['command'] === 'Custom shell command');
```
jim-p: This variable is only used once, better to not set it and use the value directly if it isn't reused.

mintsoft:
Fixed @jim-p

PR #370, opened 2017-07-01
Correct unblock src/dest IP method call.

renaudholcombe:
This is to correct Bug #7578 (unblock links do not unblock on suricata alerts page).

Bug for reference: https://redmine.pfsense.org/issues/7578


rbgarga:
Can you check this out please @bmeeks8 ?

bmeeks8:
The suggested fix is good.  Thank you @renaudholcombe for the submission.
Bill Meeks

rbgarga:
Note: Merge combined with #380

renaudholcombe:
Please apply this pull request (#370) before #379.

PR #371, opened 2017-07-14
Cellular Package Fixes

svenauhagen:
Hi,

we fixed multiple problems in the cellular package:

1. Carrier was not displayed in Widget
2. Signal strength scale was off by 1
3. AT commands are tried for 3 times since the port is sometimes busy

Best
Sven

### DIFF
--- a/net/pfSense-pkg-cellular/Makefile
+++ b/net/pfSense-pkg-cellular/Makefile
@@ -2,7 +2,6 @@
 
 PORTNAME=	pfSense-pkg-cellular
 PORTVERSION=	1.1.5
-PORTREVISION=	1
 CATEGORIES=	net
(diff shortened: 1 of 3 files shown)


net/pfSense-pkg-cellular/Makefile
```diff
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-cellular
-PORTVERSION=	1.1.4
+PORTVERSION=	1.1.5
 PORTREVISION=	1
```
rbgarga: Remove PORTREVISION line when bumping PORTVERSION

svenauhagen:
Hi, 

I removed the portrevision line.

Best
Sven

PR #372, opened 2017-07-14
Only check custom networks CIDRs validity when actually using the cus…

madpilot78:
…tom list mode.

My commit fc2f34fb39 contains a mistake, it always checks the CIDR address list, even if it is not actually  using it, causing an empty field to be refused.

This commit addresses this issue. Sorry for my mistake.

Can this fix be "fast tracked"?

Thanks!

### DIFF
--- a/net/pfSense-pkg-ntopng/Makefile
+++ b/net/pfSense-pkg-ntopng/Makefile
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-ntopng
-PORTVERSION=	0.8.9
+PORTVERSION=	0.8.10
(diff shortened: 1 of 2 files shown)


dennypage:
Agreed, this should be integrated as quickly as possible.

jim-p:
In the future, it's much easier to merge these if you keep the forked copy of the repo with the same name (e.g. FreeBSD-ports stays FreeBSD-ports and not pfSense-FreeBSD-ports). Had to merge it manually rather then being able to use our usual tools.

madpilot78:
I did not know the repository name played a role. Sorry it caused difficulties for you.

I'll avoid renaming repositories in thee future.

PR #373, opened 2017-07-15
Bug 7616 - Barnyard2 webui configuration updates result in ****** written to the config for the password

renaudholcombe:
The logic of the password input type will send a hardcoded '********' post value when the value has not been edited which was not handled. The changes are to escape when the DMYPWD constant is being passed in. Another aspect of this is that when the value has not changed, it needs to be re-encoded to base64 since it was decrypted during page load.

This is to address [Bug-7616](https://redmine.pfsense.org/issues/7616).

### DIFF
--- a/security/pfSense-pkg-snort/files/usr/local/www/snort/snort_barnyard.php
+++ b/security/pfSense-pkg-snort/files/usr/local/www/snort/snort_barnyard.php
@@ -178,7 +178,14 @@ if ($_POST['save']) {
 		if ($_POST['barnyard_dbhost']) $natent['barnyard_dbhost'] = $_POST['barnyard_dbhost']; else unset($natent['barnyard_dbhost']);
 		if ($_POST['barnyard_dbname']) $natent['barnyard_dbname'] = $_POST['barnyard_dbname']; else unset($natent['barnyard_dbname']);
 		if ($_POST['barnyard_dbuser']) $natent['barnyard_dbuser'] = $_POST['barnyard_dbuser']; else unset($natent['barnyard_dbuser']);
-		if ($_POST['barnyard_dbpwd']) $natent['barnyard_dbpwd'] = base64_encode($_POST['barnyard_dbpwd']); else unset($natent['barnyard_dbpwd']);
+
+		// The password field will return '********' if no changes are made and needs to be escaped.
+        if ($_POST['barnyard_dbpwd'] && ($_POST['barnyard_dbpwd'] != DMYPWD)) $natent['barnyard_dbpwd'] = base64_encode($_POST['barnyard_dbpwd']);
+        else
+			// Because of the base64 encoding/decoding, in the case of a valid value that hasn't changed, it needs to be re-encoded to base64.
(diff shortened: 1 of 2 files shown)


bmeeks8:
Thanks for this bug fix as well.  It had been on my TODO list, but I had not gotten around to it.

Bill

renaudholcombe:
@bmeeks8, not a problem, glad to help. 

Off-topic, but I've been digging around in the redmine issues/roadmap to look for quick hits; if there are other issues of higher priority you'd like me to look at when I've got a little time, let me know. I'm new to pfSense development so I'm still on the learning curve, but happy to dig into items you see as more important.

bmeeks8:
I suspect the same two bugs you fixed in the Snort GUI package are also present in the Suricata GUI package since those two packages share a lot of identical code.  If you have some time and want to take a look for the same code issues in the Suricata package, I would appreciate it.  I have been very busy the last year with some outside consulting/contract work and that has taken all of my time I used to be able to dedicate to the Snort and Suricata packages.

Thanks again for jumping in and helping out with the package maintenance!

Bill

renaudholcombe:
Good thinking, I'll take a look.

security/pfSense-pkg-snort/files/usr/local/www/snort/snort_barnyard.php
```diff
@@ -178,7 +178,14 @@
...
 		if ($_POST['barnyard_dbname']) $natent['barnyard_dbname'] = $_POST['barnyard_dbname']; else unset($natent['barnyard_dbname']);
 		if ($_POST['barnyard_dbuser']) $natent['barnyard_dbuser'] = $_POST['barnyard_dbuser']; else unset($natent['barnyard_dbuser']);
-		if ($_POST['barnyard_dbpwd']) $natent['barnyard_dbpwd'] = base64_encode($_POST['barnyard_dbpwd']); else unset($natent['barnyard_dbpwd']);
+
+		// The password field will return '********' if no changes are made and needs to be escaped.
+        if ($_POST['barnyard_dbpwd'] && ($_POST['barnyard_dbpwd'] != DMYPWD)) $natent['barnyard_dbpwd'] = base64_encode($_POST['barnyard_dbpwd']);
+        else
+			// Because of the base64 encoding/decoding, in the case of a valid value that hasn't changed, it needs to be re-encoded to base64.
+            if ($_POST['barnyard_dbpwd'] != DMYPWD) unset($natent['barnyard_dbpwd']);
+            else $natent['barnyard_dbpwd'] = base64_encode($natent['barnyard_dbpwd']);
```
rbgarga: Please fix style

bmeeks8: This change is essentially identical to the one submitted for Suricata.  The same issue with re-encoding an already encoded password upon a "save" operation exists in Snort since both packages share much of the same GUI code.  I am OK with this fix, but I do agree the if/then/else statement indents need fixing up.

Bill

PR #374, opened 2017-07-16
Bug #7674 - Issue Downloading Snort Alert Log Download

renaudholcombe:
As part of preparing the response to the download request we were incorrectly setting the location header and exiting the operation.

This addresses [Bug 7674](https://redmine.pfsense.org/issues/7674).

### DIFF
--- a/security/pfSense-pkg-snort/files/usr/local/www/snort/snort_alerts.php
+++ b/security/pfSense-pkg-snort/files/usr/local/www/snort/snort_alerts.php
@@ -385,15 +385,13 @@ if ($_POST['download']) {
 			header("Cache-Control: private, must-revalidate");
 		}
 		header("Content-Type: application/octet-stream");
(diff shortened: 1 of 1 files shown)


bmeeks8:
Thank you for the bug fix submission.

Bill

renaudholcombe:
@rbgarga, I'll bump the version. I'm assuming we'll need to do the same PORTVERSION change for the Suricata pull requests; how do we want to handle it given there are 2 requests each for Snort and Suricata?

rbgarga:
@renaudholcombe When you submit multiple PRs changing the same port you can change the version only in one of them and add a note saying "apply this # before this #" so we can apply first the one without version change and create a single update package containing both

rbgarga:
Note: merge combined with #373

renaudholcombe:
Please apply this pull request (#374) before #373.

PR #379, opened 2017-07-22
Correct barnyard mysql password behavior in Suricata WebUI configuration

renaudholcombe:
The problem was that upon saving the settings without modifying the mysql password field for barnyard2, the value was being rewritten to the configuration file in it's base64-encoded form.

This change updates the form input type of the field from 'Input' to 'Password' for consistency and applies logic upon post to handle the default behavior of that form element.

This corrects [Bug-7716](https://redmine.pfsense.org/issues/7716).


renaudholcombe:
@bmeeks8, You were dead on, a variant of this issue existed for suricata. I also crosschecked to verify that neither of the other issues I've looked at exist in their companion pages.

security/pfSense-pkg-suricata/files/usr/local/www/suricata/suricata_barnyard.php
```diff
@@ -177,7 +177,14 @@
...
 		if ($_POST['barnyard_dbname']) $natent['barnyard_dbname'] = $_POST['barnyard_dbname']; else unset($natent['barnyard_dbname']);
 		if ($_POST['barnyard_dbuser']) $natent['barnyard_dbuser'] = $_POST['barnyard_dbuser']; else unset($natent['barnyard_dbuser']);
-		if ($_POST['barnyard_dbpwd']) $natent['barnyard_dbpwd'] = base64_encode($_POST['barnyard_dbpwd']); else unset($natent['barnyard_dbpwd']);
+
+		// The password field will return '********' if no changes are made and needs to be escaped.
+		if ($_POST['barnyard_dbpwd'] && ($_POST['barnyard_dbpwd'] != DMYPWD)) $natent['barnyard_dbpwd'] = base64_encode($_POST['barnyard_dbpwd']);
+		else
+			// Because of the base64 encoding/decoding, in the case of a valid value that hasn't changed, it needs to be re-encoded to base64.
+			if ($_POST['barnyard_dbpwd'] != DMYPWD) unset($natent['barnyard_dbpwd']);
+			else $natent['barnyard_dbpwd'] = base64_encode($natent['barnyard_dbpwd']);
```
rbgarga: Please fix style of this entire block respecting the same rules used in this file. It's really hard to read it the way it is.

@bmeeks8 can you check these changes please?

bmeeks8:
I am OK with the logic of the submitted changes.  They are required to fix improper behavior with the stored password.  As the submitter said, the old code (since Bootstrap conversion) was essentially re-encoding and already encoded password upon saving and thus corrupted it.

I do agree the indent style for the added if/then/else statements need fixing.

Bill

renaudholcombe:
@bmeeks8, @rbgarga, sure I can make the formatting changes. I used that style due to the nested nature of the logic from habit and not finding any similarly structured statements within the code. 

Are we looking for the entire set of statements to be on a single line or do we want to split them into multiple lines by 'if' statements, keeping the 'else' portions on the next line (if so, what sort of indentation for the nested statements do we want)?

rbgarga:
@renaudholcombe The rules we use in pfSense are defined in https://doc.pfsense.org/index.php/Developer_Style_Guide#PHP_Specific_Rules 

They are basically the same as FreeBSD defines in style(9)

renaudholcombe:
I'm sorry about spilling so much ink on this topic, but @rbgarga, I took a look at the style guide and that suggests the sort of multi-line indentation that I was using initially (in the 'Indent Style' section). I'll make changes to better match the convention within the file using new line indents per nested 'if' statements; would that be the right approach?

rbgarga:
I know there are code that doesn't follow the rules all around pfSense source, but we try to don't introduce new code like this.

Basically always use brackets `{` and `}` and use hard TABs, with size of 8, trying to stay in 80 columns, like:
```php
		if ($_POST['barnyard_dbpwd'] &&
		    $_POST['barnyard_dbpwd'] != DMYPWD) {
			$natent['barnyard_dbpwd'] = base64_encode(
			    $_POST['barnyard_dbpwd']);
		} else {
			/*
			 * Because of the base64 encoding/decoding, in the case
			 * of a valid value that hasn't changed, it needs to be
			 * re-encoded to base64.
			 */
			if ($_POST['barnyard_dbpwd'] != DMYPWD) {
				unset($natent['barnyard_dbpwd']);
			} else {
				$natent['barnyard_dbpwd'] = base64_encode(
				    $natent['barnyard_dbpwd']);
			}
		}
```

rbgarga:
Note: Merge combined with #380

renaudholcombe:
Alright @bmeeks8, @rbgarga, I think I've gotten everything sorted out across the pull requests, let me know if there are any changes you'd like. Thanks again for guiding me through this!

PR #380, opened 2017-07-22
adding extended custom http eve.json fields

fgrosswig:
why not adding this fields to the custom fields? maybe needs a suboption at the interface or could stay as it is to extend the http eve.json information as default.

### DIFF
--- a/security/pfSense-pkg-suricata/files/usr/local/pkg/suricata/suricata_generate_yaml.php
+++ b/security/pfSense-pkg-suricata/files/usr/local/pkg/suricata/suricata_generate_yaml.php
@@ -375,10 +375,21 @@ if (($suricatacfg['eve_log_alerts'] == 'on') && ($suricatacfg['eve_log_alerts_pa
 
 if ($suricatacfg['eve_log_http'] == 'on') {
 	$eve_out_types .= "\n        - http:";
-	if ($suricatacfg['http_log_extended'] == 'on')
-		$eve_out_types .= "\n            extended: yes";
-	else
-		$eve_out_types .= "\n            extended: no";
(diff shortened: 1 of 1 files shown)


pfSense-git-updates:

rbgarga:
@bmeeks8 can you check this out, please?

fgrosswig:
well, signed it ... whats next?

rbgarga:
We just need to wait for @bmeeks8 to take a look at it too

bmeeks8:
I am fine with this addition.  It will make the EVE output logging in Suricata better.  Thank you to @fgrosswig for the submission!

Bill

securitym0nkey:
I know i'm late to the game as this is already merged. But i think that are too many fields to be logged as default just by enabling extended http log. It might put unneeded load the log/even storage. For example an ELK stack where elastic would index all those fields. @fgrosswig would you be happy if we had an option to define the fields one is interested in via the gui? To have more granular control over the eve-output is something i'm working on anyway. You can have a look [here](https://github.com/securitym0nkey/FreeBSD-ports/tree/suricata-granular-eve-output-settings)

fgrosswig:
@securitym0nkey: first i would say what people going to store on elk will do with the "extend" logg option anyway extended infos. So why limiting it not takeing advantage of the built in "extended" options in exact this case. I was wondering why "extended" logging to http does not have the whole set. And second, as mentioned above, it would be a best thing to have a textfield or multiple choice to exactly choose what specific option you are going to store on elk. But this is the higher class and would need some additional tweaks on the interface i was too lazy to take also care of - beause as said: if i check amr extended then i expect the "whole" extended and not limitin extended output.

PR #383, opened 2017-08-02
pfSense-pkg-suricata: Suricata eve flow patch

johannrichard:
Suricata can (also) log flow info in EVE, which can be used when analysing with an ELK stack. This change enables this in the interface editor as well as creates the corresponding `YAML`. 

Since enabling this option potentially generates larg(er) log files, including packet capture, the option is disabled by default.

### DIFF
--- a/security/pfSense-pkg-suricata/Makefile
+++ b/security/pfSense-pkg-suricata/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-suricata
 PORTVERSION=	3.2.3
-PORTREVISION=   1
+PORTREVISION=	1
 CATEGORIES=	security
 MASTER_SITES=	# empty
 DISTFILES=	# empty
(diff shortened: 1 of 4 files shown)


rbgarga:
@bmeeks8 can you take a look please?

bmeeks8:
I am fine with this logging option addition.  Thank you to @johannrichard for the contribution!  There have been lots of folks contributing to improving the package recently.  Good to see the interest and willingness to chip in and help improve the package.

Bill

security/pfSense-pkg-suricata/Makefile
```diff
@@ -2,6 +2,7 @@
 
 PORTNAME=	pfSense-pkg-suricata
 PORTVERSION=	3.2.3
+PORTREVISION=	1
```
johannrichard: Added `PORTREVISION`

PR #384, opened 2017-08-03
pfSense-pkg-freeradius3: Allow 127.0.0.1 for interface selection

jwessel:
The configuration of using the radius server and a single client on
the loopback device should be allowed.  This allows a secure
method for using OTP tokens on a single firewall.

Signed-off-by: Jason Wessel <jason@wesselworks.com>

### DIFF
--- a/net/pfSense-pkg-freeradius3/Makefile
+++ b/net/pfSense-pkg-freeradius3/Makefile
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-freeradius3
-PORTVERSION=	0.8
+PORTVERSION=	0.9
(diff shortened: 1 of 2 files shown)


jwessel:
No problem.  I bumped the PORTVERSION.

net/pfSense-pkg-freeradius3/Makefile
```diff
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-freeradius3
-PORTVERSION=	0.8
+PORTVERSION=	0.9
 CATEGORIES=	net
```
jwessel: Bumped PORTVERSION from 0.8 to 0.9.

PR #385, opened 2017-08-03
Table BGColor

NonSecwitter:
the table is unreadable with current color settings and pfSense portal in dark theme.

Here are the colors I've tested: https://snag.gy/VFxeiK.jpg
#919191 is the line with "[blk_BL_automobile_bikes]"
Note that the word "access" at the end of the line is black to reflect light theme colors.
The lowest row, just above the developer tools, is the original presentation. The text is unreadable.

### DIFF
--- a/www/pfSense-pkg-squidGuard/files/usr/local/pkg/squidguard.inc
+++ b/www/pfSense-pkg-squidGuard/files/usr/local/pkg/squidguard.inc
@@ -659,7 +659,7 @@ function squidguard_before_form_dest(&$pkg) {
 function make_grid_general_items($id = '') {
 	global $squidguard_config;
 
-	$bg_color = 'bgcolor="#dddddd"';
+	$bg_color = 'bgcolor="#919191"';
 	$res = '';
 	$res .= '<table width="100%">';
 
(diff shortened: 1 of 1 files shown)


NonSecwitter:
I uploaded the wrong screen shot, but the color #919191 looks fine with black text as well.

sbeaver-netgate:
It would be a lot nicer to use rules from the pfSense css files to control the colors. Use a tool like Firebug to examine pfSense pages (rather than squidguard) and use the same rules for your changes.

NonSecwitter:
Understandable, I'm not much of a web developer. I don't understand CSS rules, so I just replicated the current method.

doktornotor:
@sbeaver-netgate - While I understand it'd be nice to use CSS properly here, this package is a piece of unreadable buggy spaghetti code that barely holds together and only (mostly) works by some miracle. Unless someone volunteers to rewrite it from scratch, I'd just change the hardcoded colors so that the text is readable by default and move on.

sbeaver-netgate:
Agreed. Lipstick on a pig. I'm OK with the PR as it is, but it's going to break one day when someone adds a new color scheme :)

doktornotor:
Yeah, it will eventually break again. I did not dare to install it on 2.4 yet to test whether it still produces some action (people jhave still been using this on 2.3.x due to lack of alternatives, so it must have been doing something useful for them). I fear it'd eat my disk and make all my beer stock vanish from the fridge, LOL.

rbgarga:
@NonSecwitter Please bump PORTVERSION *or* PORTREVISION in Makefile so a new package will be produced

rbgarga:
Remote repository is gone

PR #387, opened 2017-08-07
pfSense-pkg-suricata: Delete rotated `eve.json` log files

johannrichard:
The automatic log removal under `suricata_check_dir_size_limit` currently does not clean-up the `eve.json.*` log. This can lead to a situation where the other rotated logs are cleared but the overall directory size can still grow beyond the hard limit. 

For example, if `eve.json` log retention is set to more than a few days and with `flow` logging enabled: there you can end up with dozens of files that are not cleared in time. 

With this change, that shouldn't happen anymore. 

This change is especially relevant with respect to changes introduced in #383 that result in large `eve.json` files if `flow` is enabled.


bmeeks8:
@johannrichard :  would you consider also including the code in the patch a user submitted on the Redmine bug site for pfSense here:  https://redmine.pfsense.org/issues/7756#change-33404?

He included a nice patch that makes the log cleanup code much smarter, and it also corrects a few oversights.

Thanks,
Bill

johannrichard:
I'm fine with the log cleanup implementation in #389 which is indeed much more comprehensive! I therefore close this request. Thanks, @opoplawski.

PR #392, opened 2017-08-16
Suricata-4.0.0 Update

bmeeks8:
This updates the Suricata binary to version 4.0.0 to match the latest upstream version.

**New Features**
A new configurable parameter has been added to the _alert-pf_ plugin used to insert offending IP addresses into the "snort2c" table in the packet filter of pfSense.  The new parameter is named "block-drops-only" and can be set to 'yes' or 'no' (with 'no' being the default when no setting is specified).  When set to 'yes', this new setting causes the plugin to only insert offending IP addresses into the "snort2c" table when the firing rule signature has a rule action of DROP.  Rules with action ALERT will only cause alerts with no blocks if this new option is set to 'yes'.

The _alert-pf_ plugin's suricata.yaml configuration settings are shown below
```
    # alert-pf blocking plugin
    - alert-pf:
        enabled: yes/no            # "yes" to enable blocking plugin
        kill-state: yes/no         # "yes" to kill open state table entries associated with blocked IP addresses (default is "yes")
        pass-list: <filename>      # complete path and filename for txt file of single IP addresses or CIDR networks that should never be blocked
        block-ip: src/dst/both     # which IP in packet to block (default is BOTH)
        pf-table: <pf table name>  # name of packet filter firewall table where block IP addresses should be added.  This table must exist!
        block-drops-only: yes/no   # only insert blocks in packet filter firewall table for rules having DROP action keyword (default is "no")
```

### DIFF
--- a/security/suricata/Makefile
+++ b/security/suricata/Makefile
@@ -1,8 +1,8 @@
 # Created by: Patrick Tracanelli <eksffa@freebsdbrasil.com.br>
-# $FreeBSD$
+# $FreeBSD: head/security/suricata/Makefile 447598 2017-08-09 08:15:21Z joneum $
 
 PORTNAME=	suricata
-PORTVERSION=	3.2.3
+PORTVERSION=	4.0.0
 CATEGORIES=	security
 MASTER_SITES=	http://www.openinfosecfoundation.org/download/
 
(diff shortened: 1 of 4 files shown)


bmeeks8:
There will be a matching update for the Suricata GUI package coming soon, but until then there are no compatibility issues with this update and the current GUI package.  The only side effect of using the 4.0.0 binary with the 3.2.3 GUI package is that upon startup an error message will be logged about the missing "block-drops-only" configuration for that new parameter, but the plugin will default to "no" for the setting which is the legacy behavior anyway.  Thus there will be no impact to Suricata's operation.

rbgarga:
@bmeeks8 I've imported it manually to devel and RELENG_2_3 because I did a cherry-pick of original commit + a manual commit adding patch-alert-pf.diff. New binary will be available soon. Thanks!!

PR #394, opened 2017-08-18
Bug Fix for Snort-2.9.9.0 Binary Package

bmeeks8:
## Snort-2.9.9.0_2

This update simplifies the CSV output module patch used with Snort on pfSense.  The simplifed patch will be easier to maintain.  Also added some missing null pointer checks within the CSV patch that were likely causing some user-reported Signal 11 errors in certain circumstances.  Finally, added back the missing _attribute_table.dtd_ file that should get copied to the **etc/snort** directory upon installation.  This file is used when the Host Attribute Table preprocessor is enabled.

This update is backwards compatible with the existing Snort GUI package, so the new binary can be built immediately.


bmeeks8:
I will be submitting another pull request in the near future to fix some Snort GUI package bugs, but this binary package update is not dependent on that updated GUI package.  This binary update can be built and made available to users immediately to give them the benefit of the Signal 11 bug fixes.

Bill

PR #395, opened 2017-08-18
pfBNG: Add a work-around for Bug #6603

doktornotor:
Prevents unbound from bombing out when the file does not exist [(Bug #6603).](https://redmine.pfsense.org/issues/6603) Note that the same issue completely breaks fresh 2.4 installs when you restore a config with pfBNG's DNSBL feature enabled due to [Bug #7604](https://redmine.pfsense.org/issues/7604).

@BBcan177 - do whatever you want with this; I'm simply fed up with this issue.

Alternative solution for pfSense itself suggested [in Bug #6603 comment 11](https://redmine.pfsense.org/issues/6603#note-11) however I can see a problem with that due to the server: clause being required for some options, while not for others, so, a mere `include: /var/unbound/conf.d/*.conf` in default ```unbound.conf``` does not seem exactly viable (and skipping it when required would break unbound start as well.)


doktornotor:
The real solution I could see here would be somewhere along the lines of implementing ```includeifexists: /foo/bar/baz.conf``` in Unbound itself, or simply making the failed include non-fatal there. Nothing that would get solved short-term unfortunately. 😞

jim-p:
I think letting a glob handle the work here is good enough for now, also this would allow splitting directives into multiple files if needed without having to specify them all individually.

BBcan177:
I haven't tested this patch and will not be able to test for another 2-3 weeks. However, it should be ok to merge with the changes indicated below.

Please test it fully and if it breaks, please fix it promptly :)

1.
The glob might need to be changed to .*conf as there are other pfb_dnsbl.??? files that could be in the /var/unbound folder. Should unbound be reloaded during a DNSBL cron update, the following files can be present in the /var/unbound folder and can't be included in the glob.

    pfb_dnsbl.raw
    pfb_dnsbl.tsp
    pfb_dnsbl.dup

If debug mode is enabled in pfblockerng.sh (Next release) there are these file references:

    pfb_dnsbl.conf.final
    pfb_dnsbl.raw.orig
    pfb_dnsbl.bkr
    pfb_dnsbl.bkraw

and the following new file indicator:

    pfb_dnsbl.reload

2. In Line 3834 of pfblockerng.inc, this will also need to be changed to reflect the new glob:

       if (isset($conf) && !strstr(implode($conf), 'pfb_dnsbl.conf')) {

3. Please remove the "broken" reference in line 1034 and replace that with "previous".

4. In Line 1034 of pfblockerng.inc, change the `$log =` to a `$log .=` to append the log variable.

**Note**: 

This will also **not fix any existing backups** that contain the previous `pfb_dnsbl.conf` reference until a Force Update or Cron event takes place. So it still might result in a failed unbound start in this instance. I still feel that this is best resolved in Unbound to skip a missing include file, or in pfSense Unbound code to touch the include file if not found... <my 2 cents>

Thanks for the patch Dok!

doktornotor:
Changed as requested by @BBcan177.

Regarding existing backups - there's really not much that could be done there unless someone wants to add some magic to `/etc/inc/upgrade_config.inc`

PR #396, opened 2017-08-18
pfSense-pkg-suricata:  Granular eve output settings

securitym0nkey:
These changes allow more granular control over the eve output.


bmeeks8:
@rbgarga:

I am OK with these updates going into the package.

Bill

PR #397, opened 2017-08-22
pfSense-pkg-snort-3.2.9.5 - GUI Package Update to v3.2.9.5

bmeeks8:
## Snort GUI Package v3.2.9.5
This updates the _pfSense-pkg-snort_ GUI package to version 3.2.9.5.  One new feature and several reported bug fixes are included in this update.

**New Features**
1. The ARP Spoofing preprocessor is now exposed in the GUI as a configurable option on the PREPROCESSORS tab.  Settings are available for toggling the enabled state of the preprocessor and for enabling detection of unicast ARP requests.  A multi-entry table is provided that allows for entry of MAC address-to-IP address pairs to be monitor for ARP spoofing incidents.  New MAC/IP address pairs can be added to the table and existing MAC/IP pairs can be edited or deleted from the table.

**Bug Fixes**
1. Fix display of IPv6 addresses so they wrap correctly when displayed in the SRC IP and DST IP columns on the ALERTS tab.

2. Fix the DOWNLOAD button on the ALERTS and BLOCKS tabs so that they work.  Also fix the ALERTS download so that it only includes alert logs and not all log files in the directory.

3. Restore the installation of the _attribute_table.dtd_ validation file to the Snort conf directory.

4. Do not show Reverse DNS Lookup and Track-by-IP icons on the ALERTS tab for alert entries that do not contain an IP Header and thus no IP addresses (usually from alerts generated by the ARP Spoof preprocessor).

5. Improve package uninstall procedure by manually cleaning up files created or altered by the GUI code.  The default _pkg_ uninstall code does not remove files modified by others.

6. Remove the shared object rules files when uninstalling the package.  This should fix errors during package upgrades when older versions of these files still exist.

7.  Add a warning to the OpenAppID rules file download section on the GLOBAL SETTINGS tab about Geo-IP blocking at the volunteer hosting web site.  This hosting block may impact users in some countries when they attempt to enable the OpenAppID rules download.  (NOTE - this is for the rules only.  The OpenAppID detectors are maintained by the Snort VRT and should always download and install.  However, the detectors need the rules in order to be fully functional.)

### DIFF
--- a/security/pfSense-pkg-snort/Makefile
+++ b/security/pfSense-pkg-snort/Makefile
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-snort
-PORTVERSION=	3.2.9.4
+PORTVERSION=	3.2.9.5
 CATEGORIES=	security
 MASTER_SITES=	# empty
 DISTFILES=	# empty
--- a/security/pfSense-pkg-snort/files/usr/local/pkg/snort/snort_generate_conf.php
+++ b/security/pfSense-pkg-snort/files/usr/local/pkg/snort/snort_generate_conf.php
@@ -928,6 +928,21 @@ preprocessor appid: \
 
 EOD;
 
+/* def ARP Spoof preprocessor */
+$arpspoof_preproc = "# ARP Spoof preprocessor #\n";
+if ($snortcfg['arp_unicast_detection'] == 'on') {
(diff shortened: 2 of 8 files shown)


bmeeks8:
An update to the Snort binary package is coming right behind this GUI package update, but the two pull requests are not dependent on each other.  The binary package update will finally finish fixing a bug that was partially addressed in the 2.9.9.0_2 update.

Bill

PR #398, opened 2017-08-22
Snort 2.9.9.0_3 -- Bug Fix for custom blocking module

bmeeks8:
## Snort 2.9.9.0_3
This update corrects a bug in the custom blocking plugin for the Snort binary.  Failure to validate that the IP header information in an alerting packet is valid before attempting to insert the IP addressess into the "snort2c" table could cause Signal 11 faults in the Snort binary.  Certain alerts, particularly from the ARP Spoofing detection preprocessor, may not contain valid IP header information.  The IP header information is now validated before attempting to insert the addresses into the "snort2c" table.  Alert packets with empty or invalid IP header information are now ignored by the blocking plugin since it could not do anything useful with them anyway.

### DIFF
--- a/security/snort/Makefile
+++ b/security/snort/Makefile
@@ -3,7 +3,7 @@
 
 PORTNAME=	snort
 PORTVERSION=	2.9.9.0
-PORTREVISION=	2
+PORTREVISION=	3
 CATEGORIES=	security
 MASTER_SITES=	https://snort.org/downloads/snort/ \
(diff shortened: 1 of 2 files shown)


bmeeks8:
Hopefully this finally kills Signal 11 bug in the Snort binary.  A partial fix was in version 2.9.9.0_2 where the CSV log output was beefed up with NULL pointer checks.  This pull request should close the door on the Signal 11 faults.

Bill

PR #400, opened 2017-08-22
pfBlockerNG, fix curl proxy options

PiBa-NL:


PiBa-NL:
BBCan177, will add a fix for this to his updated version, closing this request.

PR #407, opened 2017-08-26
pfSense-pkg-apcupsd: Pull in apcupsd-3.14.14_2

Vynce:
My primary motivation for pulling in the rev 2 package of apcupsd-3.14.14 is the MODBUS USB driver option (I just cherry-picked f7e0540ed53e2225730aa36ddd00b037ad38e369). I wasn't sure if the pfSense-pkg-apcupsd port revision also needed to be bumped up to trigger a rebuild of the package with the updated apcupsd dependency.

CC: @dbaio


pfSense-git-updates:

PR #411, opened 2017-08-30
net/pfSense-pkg-freeradius3: Revert broken change that breaks configuration

doktornotor:
Not sure what was the change supposed to fix but the escaping in heredoc breaks the thing altogether.

```
# radiusd -X
/usr/local/etc/raddb/policy.d/pfs_custom_policies[2]: Reference "${policy.mac-addr-regexp}" not found
/usr/local/etc/raddb/policy.d/pfs_custom_policies[2]: Parse error expanding ${...} in condition
Errors reading or parsing /usr/local/etc/raddb/radiusd.conf
```
@jim-p Perhaps make use of the example here: https://wiki.freeradius.org/guide/mac-auth#plain-mac-auth_raddb-policy-conf ?

### DIFF
--- a/net/pfSense-pkg-freeradius3/Makefile
+++ b/net/pfSense-pkg-freeradius3/Makefile
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-freeradius3
-PORTVERSION=	0.14
+PORTVERSION=	0.15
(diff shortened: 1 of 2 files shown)


doktornotor:
https://github.com/pfsense/FreeBSD-ports/pull/412 has a proper fix.

PR #413, opened 2017-08-31
Adding OSPF cost-shifting as a CARP failover mechanism for scenarios requiring rapid failover in HA pairs.

quadrinary:
This latest modification to the Quagga package allows users to select the previously existing CARP mode that outright disables the Quagga service or use a hot/warm configuration wherein routers in a HA pair have their OSPF interface costs shifted according to CARP status, allowing for rapid transition from one router to another. Crosslinked in redmine: [https://redmine.pfsense.org/issues/8181](url)


quadrinary:
Per Jim P, I am also working this into FRR, I just figured I'd get this out there now since I've had it in my repo for a month or two now.

jim-p:
The quagga parts look OK but the FRR commits cannot be on the same PR.

quadrinary:
There we go, I reverted the FRR changes, they are no longer part of this pull request.  I do have some thoughts before this gets merged -- I was considering the possibility that users currently using the original form of CARP failover would suddenly find themselves without any CARP monitoring, as there would be no method defined. Is this a major concern or should I work on a method to compensate for the upgrade?

jim-p:
That is a significant concern. It should probably default to the old behavior when it is not set, and the drop-down should default to the old method to cover that unset case as well. You could maybe rig up some upgrade code but that doesn't seem worth it.

quadrinary:
I'll sleep on it and think over coding a fast upgrade. Ideally, what I envision is checking for the presence of the _$ospfd_conf['carpmode']_ parameter. If it does NOT have an assigned value of any kind but _$ospfd_conf['carpstatusvid']_ has a value, I would have it force _$ospfd_conf['carpmode'] == "quaggadisable"_.  This should be pretty easy, but I'll need to thoroughly test it first.

quadrinary:
I've added an upgrade path as discussed in my previous commented and tested it during an upgrade from a prior version.

quadrinary:
Corrected for one last case, wherein the "carpstatusvid" parameter is not set or is set to a non-valid value. Code now verifies ""carpstatusvid" begins with the valid "_vip" value before applying the upgrade patch.

quadrinary:
For testing, I'll suggest a few things. 

When selecting your active/standby OSPF costs, I generally give the standby unit lesser values in each field. For example, on my primary units, I set the OSPF "active" cost to 9 and the "backup" cost to 19000. On the standby unit, I set OSPF "active" cost to 5 and the "backup" cost to 15000. 

That way, if there is a CARP event, the LSA containing the updated cost values from the _standby_ unit will always be more preferable when it is taking over; if both units had the same "active" cost, then there is a chance that connected routers may attempt equal cost load balancing, which does not play well with Quagga, and is anyway not what you want when you've just experienced a CARP event. 

Likewise, when you fail to "backup" state, you want your costs to be more extreme on the primary unit so that it takes itself out of the OSPF equation if the secondary unit is taking a bit to get its LSAs out.

quadrinary:
Any idea on the timeline for this?

jim-p:
We'll be checking into PRs and merging again once we have 2.4.0 released and things settle down here. We're all tied up at the moment.

quadrinary:
Makes sense, thanks for the info!

rbgarga:
@quadrinary Please rebase your fork

quadrinary:
Done.

jim-p:

Thanks!

quadrinary:
https://redmine.pfsense.org/issues/8181

net/pfSense-pkg-Quagga_OSPF/pkg-descr
```diff
@@ -1,2 +1,7 @@
+<<<<<<< HEAD
+OSPF routing protocol using Quagga with manual configuration of BGP and OSPFv6.
+Conflicts with OpenBGPD; both packages cannot be installed at the same time.
+=======
 OSPF routing protocol using Quagga.
 Conflicts with OpenBGPD and FRR; these packages cannot be installed at the same time.
+>>>>>>> upstream/devel
```
rbgarga: Leftover from rebase

rbgarga:
Assigning for @sdj38 for testing

Valentin-N:
Are there any plans to get this functionality committed into FRR?

quadrinary:
rbgarga - I hadn't realized that there was a conflict, I must've missed the alert email from Github.

I'm not sure what I need to do to resolve the conflict, it looks like it's just a build number issue on the makefile; can you give me any guidance?

quadrinary:
Valentin-N - I've been in the midst of a number of changes at work lately that have hinged on the cost shifter in Quagga and once that project is complete this summer, I hope to circle back and get this functionality into FRR as well. It won't take a whole lot of work but I'll need to find time to test before submitting.

rbgarga:
No answers from submitter in 3 months.

quadrinary:
@rbgarga - what answers were expected? I did not see any questions asked of me regarding this.

rbgarga:
@quadrinary your branch had conflicts and could not be merged. If you want to re-submit it please do and we can review again.

PR #414, opened 2017-09-01
Cellular Update

svenauhagen:
Hi,

this is a larger update to the cellular package.
The main changes are:

1) Add a fixed mount point for a cellular modem /dev/cuaZ99.0 (connection) and /dev/cuaZ99.1 (data) when the modem is added (with a devd conf symlinked to the original location)
2) A script to trigger a modem reconnect or reset if not connected that can be enabled in the GUI
3) An option to reset the modem in the GUI

This will make cellular modems much easier to handle and recoverable on problems.

Please let me know what you think.

Best
Sven


net/pfSense-pkg-cellular/files/usr/local/pkg/cellular.inc
```diff
@@ -60,7 +64,34 @@ function cellular_sync() {
 
 	foreach (["port", "baudrate", "timeout", "initstring"] as $key) {
 		$key_val = ($cellular[$key] != "default" && $cellular[$key] != "") ? $cellular[$key] : "";
-		exec(PYTHON_BIN . ' ' . INTERFACE_BIN . ' $device $baudrate $timeout setcfg Interface $key $key_val signal');
+		exec(PYTHON_BIN . ' ' . INTERFACE_BIN . ' setcfg Interface ' . $key . ' ' . $key_val);
+	}
+
+	#Cron
+	install_cron_job('cellular_check.php', false);
```
doktornotor: This will trigger write_config() on every package resync. How about using else when the package not enabled?

svenauhagen: The reason why I did it this way is that the cron job script has different command line options like -r -i and so on.
If the user changes them in the GUI and the cron job options change I have no way of detecting the cron job that is already in place and replace it.
I will try and find a better way to detect the old cron job.

svenauhagen: Ok, I pushed a check for this, please let me know if that is better.

Best
Sven

doktornotor:
I think you've missed the point. There's no need to redo what install_cron_job() already does on it's own. But with the original PR, when the cronjob was enabled and was supposed to stay enabled, you always disabled it, then re-enabled again. Just needed moving the one line.

svenauhagen:
Let me try and explain the problem in another way. 
The cron command checks for the exact command only. 
My code generates different commands depending on the user input.
To show the problems let's me give an example:

Command1: mycommand -I eth0 -r

Now user changes the interface:
Command2: mycommand -I eth1 -r

If I implement your change I will have both entries in cron because cron will not delete the first command when I add the second. I do not have a way of reconstructing command1 after the user changed the GUI option. 
Therefore I need to search for the base command remove all entries and readd the correct in order to only have one entry on cron. 

I hope that explains the problem better. 

You are correct if the command is always the same but mine isn't.

net/pfSense-pkg-cellular/files/usr/local/bin/cellular_dev.py
```diff
@@ -0,0 +1,98 @@
...
+
+data_file = "cuaZ99.0"
+data_path = "/dev/" + data_file
+mgm_file = "cuaZ99.1"
+mgmt_path = "/dev/" + mgm_file
+
+def remove_link():
+
+    if os.path.lexists(data_path):
+        os.system("rm " + data_path)
```
jim-p: Why do this instead of using os.remove() / os.unlink()? Using native functions is better than shell exec. Same applies to other instances using this syntax.

net/pfSense-pkg-cellular/files/usr/local/bin/cellular_dev.py
```diff
@@ -0,0 +1,98 @@
...
+args.silent = False
+
+if args.add:
+    if args.device:
+
+        #Make sure the links are gone
+        remove_link()
+
+        #Before we start kill 3gstat as it blocks us
+        os.system("/bin/ps auxww | /usr/bin/grep \"[3]gstats\" | /usr/bin/awk '{print $2}' | /usr/bin/xargs kill")
```
jim-p: Using pkill here would be better, everything can be done with that one command, for example:

`/bin/pkill -fx '.*[3]gstats.*'`

net/pfSense-pkg-cellular/files/usr/local/bin/cellular_dev.py
```diff
@@ -0,0 +1,98 @@
...
+
+        dev_pre = args.device[:-1]
+        dev_post = args.device[-1]
+        dev_ug = commands.getoutput("sysctl -n dev." + dev_pre + "." + dev_post + ".ttyname")
```
jim-p: Use the full path to commands, `/sbin/sysctl`

net/pfSense-pkg-cellular/files/usr/local/bin/cellular_dev.py
```diff
@@ -0,0 +1,98 @@
...
+        path_ug = "/dev/cua" + dev_ug
+
+        data_ug = path_ug + ".0"
+        if os.path.exists(data_ug):
+            os.system("ln -s " + data_ug + " " + data_path)
```
jim-p: Use `os.link()` instead of a shell exec.

net/pfSense-pkg-cellular/files/usr/local/bin/cellular_dev.py
```diff
@@ -0,0 +1,98 @@
...
+
+        mgmt_ug = path_ug + ".2"
+        if os.path.exists(mgmt_ug):
+            os.system("ln -s " + mgmt_ug + " " + mgmt_path)
```
jim-p: Use `os.link()` instead of a shell exec.

net/pfSense-pkg-cellular/files/usr/local/pkg/cellular.inc
```diff
@@ -60,7 +64,50 @@ function cellular_sync() {
 
 	foreach (["port", "baudrate", "timeout", "initstring"] as $key) {
 		$key_val = ($cellular[$key] != "default" && $cellular[$key] != "") ? $cellular[$key] : "";
-		exec(PYTHON_BIN . ' ' . INTERFACE_BIN . ' $device $baudrate $timeout setcfg Interface $key $key_val signal');
+		exec(PYTHON_BIN . ' ' . INTERFACE_BIN . ' setcfg Interface ' . $key . ' ' . $key_val);
```
jim-p: Use `escapeshellarg()` on `$key_val`, so `escapeshellarg($key_val)`

net/pfSense-pkg-cellular/files/usr/local/pkg/cellular.inc
```diff
@@ -60,7 +64,50 @@ function cellular_sync() {
...
+	}
+
+	#Cron
+	if ($cellular["enable_cellular_check"] == "on") {
+
+		$check_interface = '-i ' . $cellular["check_interface"];
```
jim-p: Use `escapeshellarg()` on `$cellular["check_interface"]`.

net/pfSense-pkg-cellular/files/usr/local/pkg/cellular.inc
```diff
@@ -60,7 +64,50 @@ function cellular_sync() {
...
+		$check_interval = $cellular["check_interval"];
+		$enable_reset_modem = '';
+
+		if ($cellular["enable_reset_modem"]) {
+			$enable_reset_modem = '-r ' . $cellular['reset_count'];
```
jim-p: Use `escapeshellarg()` on `$cellular["reset_count"]`.

net/pfSense-pkg-cellular/files/usr/local/sbin/cellular.py
```diff
@@ -34,524 +34,527 @@
...
+    args.silent = False
+    if args.device:
+        interface.set_device(args.device)
+    if args.baudrate:
+        interface.set_baudrate(args.baudrate)
+    if args.timeout:
+        interface.set_timeout(args.timeout)
+
+    #Before we start kill 3gstat as it blocks us
+    os.system("/bin/ps auxww | /usr/bin/grep \"[3]gstats\" | /usr/bin/awk '{print $2}' | /usr/bin/xargs kill")
```
jim-p: See my note above about `pkill`.

net/pfSense-pkg-cellular/files/usr/local/sbin/cellular.py
```diff
@@ -34,524 +34,527 @@
 import ConfigParser
 import serial
 
-__version__ = "1.1.0_B22"
+__version__ = "1.1.6"
 
 huawei = (lambda a: a.startswith("Huawei"))
 
 class CellularInterface:
-	"""
```
jim-p: Is there a compelling reason for changing this entire file from tabs to spaces? It makes the changes difficult to find and review.

net/pfSense-pkg-cellular/files/usr/local/sbin/cellular_check.php
```diff
@@ -0,0 +1,75 @@
...
+                        if (is_file($reset_file)) {
+                                $count = file_get_contents($reset_file);
+                                $count += 1;
+                        } else {
+                                $count = 1;
+                        }
+
+                        if ($count >= $reset) {
+                                print "Resetting Modem";
+                                shell_exec('echo "AT^RESET" > ' . $port);
```
jim-p: Use `escapeshellarg()` on `$port`. Or even consider writing to that file using PHP directly rather than using a shell exec.

net/pfSense-pkg-cellular/files/usr/local/sbin/cellular.py
svenauhagen: Hi,
no sorry I just fixed up the indentation which was a little all over the place.
The line that you marked was the only one that changed/was added in this file.

svenauhagen:
@jim-p 
I added the changes from yesterday

svenauhagen:
@jim-p 
ah yes, I did not know that.
I changed unlink to remove.

Thanks

jim-p:

Thanks!

svenauhagen:
@jim-p 
https://redmine.pfsense.org/issues/8193

svenauhagen:
Hi @jim-p ,

any update on this PR?

Best and thanks
Sven

jim-p:
It should be merged before too long. At the moment we're still dealing with issues from the latest quarterly ports branch merge so we haven't been looking at PRs.

svenauhagen:
Thanks for the update. Would be great if it can be merged in the near future.

PR #416, opened 2017-09-03
Move BIND command channel port to avoid clashing with Unbound (Bug #7271)

doktornotor:
Improve check for rndc-confgen and add error handling while here. Always regenerate the template to avoid stale config issues on package upgrades.

### DIFF
--- a/dns/pfSense-pkg-bind9/Makefile
+++ b/dns/pfSense-pkg-bind9/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-bind
 PORTVERSION=	9.11
-PORTREVISION=	8
+PORTREVISION=	9
 CATEGORIES=	dns net ipv6
 MASTER_SITES=	# empty
 DISTFILES=	# empty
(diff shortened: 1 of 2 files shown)


hb9cwp:
Finally, no more worries about this issue that plagues me up to this day, thanks! Will update to snapshot and test in the coming week.

Still, I believe it would have been more intuitive to move Unbound's port back to its documented default 8953, and then Bind could use its default 953 :-)
https://redmine.pfsense.org/issues/7271

doktornotor:
You can apply it manually using https://github.com/pfsense/FreeBSD-ports/commit/5fb1095cd2a5949f9f44fc985c0fe8aab77185ef.patch URL for System Patches until the patch is merged.

doktornotor:
Yeah, this is just out of scope of this PR that's supposed to deal with Bug 7271.

davidjwood:
Fair enough - I thought it was a worthwhile change whilst you were touching this code anyway. Of course, it's not necessary to fix the bug.

It's time that something was committed to fix this long standing bug that has such a trivial fix - hopefully this is it!

doktornotor:
@davidjwood I have a bunch of other things to clean up for BIND in the queue, will add this for next PR.

doktornotor:

PR #417, opened 2017-09-05
net/pfSense-pkg-frr: Add prefix-list to bgpd.conf

robsonmantovani:


jim-p:
Is there something that breaks without this? I don't doubt that is possible, given how some other areas have behaved, but in my (admittedly limited) BGP testing it was OK having the prefix lists in the zebra config and not duplicated here.

robsonmantovani:
The configuration "IP prefix-list" is needed to be set on bgpd.conf to the filter works on bgp.

Take a look on bgp configuration example: https://www.frrouting.org/user-guide/BGP-Configuration-Examples.html#BGP-Configuration-Examples

In the actual package, when I create a prefix-list, the Configurator create an "IP prefix-list" on zebra.conf but not in bgpd.conf. It's seems not Good

When I create a route-map, the Configurator create on both. It's seems Good.

In this scenario the filter rules don't work for me. 

To work, I need to "prefix-list" and "route-map" on bgpd.conf

robsonmantovani:
@jim-p thank you very much for this package. 

I've been waiting a long time for a package like this. OSPF + BGP together

jim-p:
OK, that's all I needed to know. If the filter rules don't work with out it, we'll get this merged soon.

rbgarga:
Please rebase your fork to fix conflicts

robsonmantovani:
Done!

jim-p:
See also: https://redmine.pfsense.org/issues/7935

PR #419, opened 2017-09-16
Updated to acme.sh 2.7.3 and added HE support

Kreeblah:
Updated to acme.sh 2.7.3 and added support in the GUI for Hurricane Electric DNS verification.


Kreeblah:
This needs more work.

PR #420, opened 2017-09-17
Update to upstream acme.sh 2.7.3 and add Hurricane Electric DNS auth

Kreeblah:


jim-p:
Looks OK at a quick perusal. Doesn't seem to change anything we've done to the code on purpose (e.g. nsupdate changes). Will definitely need testing though.

Kreeblah:
Yeah, the only two changes I was able to find that you made to the code were the nsupdate changes and two lines in acme.sh.  The nsupdate module hasn't been changed upstream since December, so I didn't touch it in what I included here, and I restored the two lines in acme.sh in my second commit there.

Testing would definitely be good, though.  I was able to test the Hurricane Electric option I added, and it seems to work for an end-to-end workflow, but I didn't test any of the other options.

rbgarga:
@Kreeblah please rebase your fork

Kreeblah:
Sure.  The only conflict was the port version bump I made vs. the port revision bump, and given the change to a newer upstream version, I opted to keep the port version bump.

PR #422, opened 2017-09-25
japanese

brittos:


rbgarga:
This PR is clearly wrong

brittos:
Sorry, man. It was my mistake.
Obrigado, tenha um bom dia!

PR #427, opened 2017-10-03
Fix SQUIDGUARD DATE RANGE

isacporter:
This update is a FIX a issue on set set DATE RANGE on squidguard configurator.
Before, it was not possible to create break dates. The validation mask reported error.

line 285 - This change will fix a typo in a variable
line 1838 - This change will fix the date comparison variable in case of date range
line 1841 - This change will fix the date range comparison
line 1846 - This change is the fix to set date and date range on squidguard TIMES. It will make it possible to create the date range in all possible squidguard patterns.

Now we can have timerange in pfsense's Squidguard configurator.


rbgarga:
Duplicate of #428

PR #404, opened 2017-08-26
sysutils/pfSense-Status_Monitoring: Add reason to write_config() calls

doktornotor:
Part of https://redmine.pfsense.org/issues/204

PR #432, opened 2017-10-08
pfSense-Status_Monitoring - fix other cases of "unknown" step value (Bug #6860)

doktornotor:
Based on the [description of the issue](https://forum.pfsense.org/index.php?topic=117036.msg662322#msg662322)  from people that are able to reproduce it (I definitely am not one of them), I added *all* the halved values, to prevent more revisits to the code here.


doktornotor:
Note: Port version intentionally NOT bumped to prevent yet another conflict with https://github.com/pfsense/FreeBSD-ports/pull/404

PR #435, opened 2017-10-18
Squid - Do not try to stop c-icap via fifo unless the process is actually running (Bug #7954)

doktornotor:


www/pfSense-pkg-squid/files/usr/local/pkg/squid_antivirus.inc
```diff
@@ -705,13 +705,17 @@ function squid_write_cicap_rcfile() {
 	$cicap_start_cmd = "{$cicap_bin} -f {$cicap_conf}";
 	$cicap_stop_cmd = <<< EOF
 
-if [ -p "{$cicap_pipe}" ]; then
-	/bin/echo -n "stop" > /var/run/c-icap/c-icap.ctl
-else
-	/usr/bin/killall -9 c-icap 2>/dev/null
+if /bin/pgrep -aq c-icap 2>/dev/null; then
+	if [ -p "{$cicap_pipe}" ]; then
+		/bin/echo -n "stop" > /var/run/c-icap/c-icap.ctl
```
jim-p: Now that I'm looking closer, couldn't this also be:  `/bin/echo -n "stop" > {$cicap_pipe}`?

doktornotor: Probably, been this way for ages, not sure why... 👍

jim-p: Indeed. Not a big deal though. Fix if you want, or let it go.

doktornotor:
@jim-p Changed, it's more consistent like this. 

While here, I noticed those useless conf_mount_ro()/conf_mount_rw() calls here, how about killing those as well while we are touching this? There's a bunch of them in squid.inc as well.

jim-p:
I am always good with wiping out deprecated code. Though maybe let's get this merged first and do that in a separate PR.

doktornotor:
OK, let's leave this for another PR.

doktornotor:

PR #436, opened 2017-10-20
Update acme.inc

topkool:
Add the support of the Gandi Live API on the pfSense GUI.
The current version of acme.sh already supports the Gandi Live API, but there was no GUI in pfSense to setup the Gandi API key.


topkool:
Done in https://github.com/pfsense/FreeBSD-ports/pull/473
Is it OK?

jim-p:
It should be an additional commit to this PR/branch and not a separate PR

jim-p:
I'll manually roll this into the larger provider update I'm about to push, it will be up later today, thanks!

PR #437, opened 2017-10-20
[pfSense-pkg-bind9] Fix XMLRPC Sync by assuring the timeout is an int

Qonstrukt:
After upgrading to pfSense 2.4, we found that the sync functionality of BIND broke the package.

The exact error was reported here by someone else:
https://redmine.pfsense.org/issues/7944

The bug was probably introduced on feb 15 in this commit:
https://github.com/pfsense/FreeBSD-ports/commit/22a3d6cb10641129cecf72444278bef46c038a62#diff-06c502c5c9e6a781e00da74f11873ef8L1008


rbgarga:
Merged manually. Thanks

PR #438, opened 2017-10-21
Squid - Allow "all" ACL in Do Not Cache, add some hints to GUI, remove useless remount code

doktornotor:
No version bump to avoid conflicts with https://github.com/pfsense/FreeBSD-ports/pull/435


doktornotor:

PR #441, opened 2017-10-21
pfSense-pkg-AutoConfigBackup - Remove useless conf_mount_{ro,rw} calls, add missing include

doktornotor:
While here, added a more explicit wording to use portal username and not email address (requested in #7688).


doktornotor:

PR #465, opened 2017-10-21
pfSense-pkg-pfBlockerNG - Remove useless conf_mount_{ro,rw} calls

doktornotor:


doktornotor:
Merged with https://github.com/pfsense/FreeBSD-ports/pull/470 - closing.

PR #467, opened 2017-10-21
pfSense-pkg-suricata - Remove useless conf_mount_{ro,rw} calls, Fix SID mods list download

doktornotor:
@bmeeks8 - I (hopefully] fixed the SID mods list download with https://github.com/pfsense/FreeBSD-ports/pull/467/commits/dad6474f550c588644ac25c3222e3804668cabfa, it was appending HTML source code to the individual files, plus the gzip download did not work at all, because the same was appended to the tar.gz breaking it completely.

There are still multiple issues with the code here, e.g. 
- when you click the Download button next to lists, the Edit no longer works and turns into download as well. 
- when you click the Download button next to lists, the "Download all" button that's supposed to create the gzip instead keeps downloading the last individual file you downloaded previously.

I don't really understand the javascript stuff being done here but it's badly broken. Frankly it'd be a whole lot better to store the files base64-encoded in config.xml, they are not exactly so huge that it'd be a significant problem.

### DIFF
--- a/security/pfSense-pkg-suricata/Makefile
+++ b/security/pfSense-pkg-suricata/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-suricata
 PORTVERSION=	4.0.0
-PORTREVISION=  2
+PORTREVISION=	3
 CATEGORIES=	security
 MASTER_SITES=	# empty
 DISTFILES=	# empty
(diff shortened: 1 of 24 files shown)


bmeeks8:
Thanks for the fixes.  I will take a look at the JavaScript code in the SID MODS tab file to see what's going on.  Storing the SID mod files content in _config.xml_ is OK with me, but it may take me a bit of time to get it coded up.  One thing the current code allows that I would like to continue is the ability to have multiple named configurations (different _disablesid.conf_ files with differing names, for example) to pick and choose from at will.  Maintaining this when stored as Base64 encoded text in _config.xml_ would mean using an array to store each entry so you can have name and then content for each item.

doktornotor:

PR #469, opened 2017-10-22
adds outgoing interface option to general settings

GoOnNowGit:


doktornotor:
I must be missing something... Why would you be hiding WAN from the interface list, why would this default to LAN and why would this be global?

http://www.squid-cache.org/Doc/config/tcp_outgoing_address/

GoOnNowGit:
you're right; this doesn't make sense.

PR #476, opened 2017-11-06
Support final UPS final power kill during shutdown.

dennypage:
Support final UPS final power kill during shutdown.

Note that this PR uses a new feature introduced in pfsense PR #3867.


dennypage:
Thanks Jim. FWIW, merging in advance of https://github.com/pfsense/pfsense/pull/3867, or backporting, won’t have any negative effect. It will just create an extra file that won’t be used until https://github.com/pfsense/pfsense/pull/3867 is merged.

PR #478, opened 2017-11-16
Update Let's Encrypt Agreement URL

areian:
The agreement was updated on November 15th, and certificate issuance fails if the old agreement URL is provided.

Not sure if this is the right place to submit this patch. If not, please point me in the right direction, if the issue has not already been fixed there.

Also, not sure if I need to bump the package version, or if that is handled by a release manager.


jim-p:
I merged this manually since the Makefile did need the version updated, and also the URL was incorrect in this PR (ended in .pdf twice). It should be showing up for users to install shortly, thanks!

areian:
Woops! Copy/paste error 😥 
Thanks for fixing and merging

PR #489, opened 2017-12-12
Add working JavaScript toggles

miken32:
Resolves [redmine issue 8189](https://redmine.pfsense.org/issues/8189)

### DIFF
--- a/dns/pfSense-pkg-bind9/files/usr/local/pkg/bind.inc
+++ b/dns/pfSense-pkg-bind9/files/usr/local/pkg/bind.inc
@@ -741,12 +741,18 @@ EOD;
 function bind_print_javascript_type_zone() {
 	$js = <<<'JS'
 	<script>
+	events.push(function(){
+		on_type_zone_changed();
+		$("textarea#resultconfig").attr("disabled", true);
+		$("textarea#dsset").attr("disabled", true);
+		$("select#type").change(function(){on_type_zone_changed();});
+
 		function on_type_zone_changed() {
 			var tipo = $("select#type").val();
 			switch (tipo) {
 				case 'master':
 					$("input#slaveip").attr("disabled", true);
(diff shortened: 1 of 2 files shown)


dns/pfSense-pkg-bind9/files/usr/local/pkg/bind.inc
```diff
@@ -739,113 +739,128 @@ EOD;
...
-	<?php
+JS;
+	echo $js;
 }
 
 function bind_print_javascript_type_zone2() {
-	print("<script language=\"text/javascript\">on_type_zone_changed(); document.iform.resultconfig.disabled = 1; document.iform.dsset.disabled = 1;</script>\n");
+	$js = <<<'JS'
+	<script>
+	// we can't put this at the end of the body, so wait until jQuery is loaded
```
sbeaver-netgate: Elsewhere throughout pfSense we use events.push() to ensure JS is executed after the include files (jQuery etc) are loaded. Would that work here? I would be better to use a single approach.

Here is a small example:

`<script type="text/javascript">
//<![CDATA[
events.push(function() {
	// Has the Use ramdisk checkbox changed state?
	if (<?=(int)$use_mfs_tmpvar_changed?> && confirm("<?=$ramdisk_msg?>")) {
		postSubmit({override : 'yes'}, 'diag_reboot.php')
	}

	// source track timeout field is disabled if sticky connections not enabled
	$('#lb_use_sticky').click(function () {
		disableInput('srctrack', !$(this).prop("checked"));
	});

	disableInput('srctrack', !$('#lb_use_sticky').prop("checked"));

});
//]]>
</script>`

miken32: Updated to use `events.push()`; also consolidated the script into one block and got rid of ugly 1990's `onchange` attributes!

PR #497, opened 2018-01-23
Snort v2.9.11.1 Update

bmeeks8:
## Snort 2.9.11.1
This updates the Snort binary to version 2.9.11.1 to keep pace with Snort upstream.  This patch for pfSense includes a compiler directive to disable compiler optimizations when compiling the package for the _armv6_ or _armv7_ CPUs.  With optimizations enabled, the compiler generates a binary that will crash with a SIGBUS error due to unaligned memory access when Snort is starting up.


rbgarga:
Manually cherry-picked commits from FreeBSD repos

PR #500, opened 2018-02-13
Cellular Package change link to symlink

svenauhagen:
Hi,

it seems that pythons os.link does not work reliably os.symlink does seem to work.
This patch changes the link to symlink.

Best
Sven


svenauhagen:
https://redmine.pfsense.org/issues/8329

PR #501, opened 2018-02-15
pfSense-pkg-softflowd: Added additional options now available in softflowd-0.9.9_1

SysError956:
Updated package version to 1.2.3

Includes new 'VLAN' flow tracking level

Includes new 'IPFIX' protocol option

Flows will now include a unique ID (or index) to differentiate between multiple instances of softflowd

The indexes will be displayed in an info box at the top of the settings page when softflowd is enabled

### DIFF
--- a/net-mgmt/pfSense-pkg-softflowd/files/usr/local/pkg/softflowd.inc
+++ b/net-mgmt/pfSense-pkg-softflowd/files/usr/local/pkg/softflowd.inc
@@ -38,6 +38,7 @@ function sync_package_softflowd() {
 	if (!empty($cf['host']) && !empty($interface_list) && $cf['enable'] != "off") {
 		$cf['host'] = is_ipaddrv6($cf['host']) ? "[{$cf['host']}]" : $cf['host'];
 		$start = "/usr/bin/killall -9 softflowd";
+		$if_index = 1;
 		foreach ($interface_list as $interface_friendly) {
 			if (empty($interface_friendly)) {
 				continue;
@@ -46,8 +47,9 @@ function sync_package_softflowd() {
 			if (empty($interface)) {
 				continue;
 			}
+			
 			$start .= "\n\t/usr/local/sbin/softflowd ";
-			$start .= " -i " . escapeshellarg($interface);
+			$start .= " -i " . $if_index++ . ":" . escapeshellarg($interface);
 			$start .= " -n " . escapeshellarg("{$cf['host']}:{$cf['port']}");
 			if (is_numericint($cf['maxflows'])) {
 				$start .= " -m " . escapeshellarg($cf['maxflows']);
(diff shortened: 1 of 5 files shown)


net-mgmt/pfSense-pkg-softflowd/files/usr/local/pkg/softflowd.inc
```diff
@@ -159,4 +161,27 @@ function cleanup_config_softflowd() {
...
+	global $config;
+
+	if (is_array($config['installedpackages']['softflowd']['config'][0])) {
+		$cf = $config['installedpackages']['softflowd']['config'][0];
+	} else {
+		$cf = array();
+	}
+	
+	/* Compile and display the list of interface indices if softflowd is active */
+	$interface_list = explode(",", $cf['interface']);
```
rbgarga: Make sure `$cf['interface']` is defined before pass it to explode()

net-mgmt/pfSense-pkg-softflowd/pkg-descr
```diff
@@ -5,7 +5,7 @@ by listening on a network interface or by reading a
 packet capture file. These flows may be reported via
 NetFlow to a collecting host or summarised within
 softflowd itself.<br/> Softflowd supports Netflow
-versions 1, 5 and 9 and is fully IPv6-capable - it can
+versions 1, 5, 9 and 10 (IPFIX) and is fully IPv6-capable - it can
 track IPv6 flows and send export datagrams via IPv6. It
 also supports export to multicast groups, allowing for
-redundant flow collectors.
+redundant flow collectors.
```
rbgarga: Please keep newline `\n` at the end of pkg-descr, otherwise portlint will complain

net-mgmt/pfSense-pkg-softflowd/files/usr/local/pkg/softflowd.inc
```diff
@@ -46,8 +47,9 @@ function sync_package_softflowd() {
 			if (empty($interface)) {
 				continue;
 			}
+			
 			$start .= "\n\t/usr/local/sbin/softflowd ";
-			$start .= " -i " . escapeshellarg($interface);
+			$start .= " -i " . $if_index++ . ":" . escapeshellarg($interface);
```
jim-p: Does that actually work properly? It will construct the string as, for example with igb1: `-i 1:'igb1'`. It looks like it might be better as: 
`$start .= " -i " . escapeshellarg($if_index++ . ":" . $interface);`

Which would result in `-i '1:igb1'

jim-p:
I'm not seeing a link to a Redmine issue for this, was one created? If not, please make an entry at https://redmine.pfsense.org

SysError956:
I apologize for the missing Redmine link. Here it is: https://redmine.pfsense.org/issues/8345

net-mgmt/pfSense-pkg-softflowd/files/usr/local/pkg/softflowd.inc
SysError956: Good catch; I clearly hacked that line together too quickly. It did however work with just the interface between the quotes. The "getopt" call in softflowd actually strips out the quotes (even the one following the colon right before the interface name); that really surprised me! Nevertheless, it was definitely an error on my part, looks better now.

net-mgmt/pfSense-pkg-softflowd/pkg-descr
SysError956: I've added the \n at the end of the pkg-descr as per your suggestion. However, oddly enough, when I run portlint (2.17.15), it complains (with a warning) of a "unnecessary blank lines at the last part".

rbgarga: Yeah, you can't have a bank line in the end of file. It's mostly an editor setting usually called "newline at EOF". Please remove the blank line

net-mgmt/pfSense-pkg-softflowd/files/usr/local/pkg/softflowd.inc
```diff
@@ -159,4 +161,29 @@ function cleanup_config_softflowd() {
...
+	global $config;
+
+	if (is_array($config['installedpackages']['softflowd']['config'][0])) {
+		$cf = $config['installedpackages']['softflowd']['config'][0];
+	} else {
+		$cf = array();
+	}
+	
+	/* Compile and display the list of interface indices if softflowd is active */
+	if (isset($cf) && isset($cf['interface'])) {
```
rbgarga: isset($cf) will be always true here, so it can be removed. Only $cf['interface'] must be checked

PR #503, opened 2018-03-07
Fix dns_nsupdate wildcard bug

pulcov:
With wildcard domain validation, script dns_nsupdate.sh search **nsupdate_acme-challenge._yourdomain_.key** and failed. File exists but name is **nsupdate_acme-challenge.*._yourdomain_.key**


jim-p:
Neither the key name nor domain name should have "*." in them. If anything, this shouldn't be masked, it should be rejected.

jim-p:
I misspoke there, the domain name should, but the key name should be manually set to an alternate value that doesn't contain "*.", which may or may not match the domain name so it isn't always safe to assume. What should be rejected is a wildcard domainname without a separate key name.

pulcov:
I do not know if that's the right way to correct the problem but with 0.2.3 version, dns domain validation fail when i use DNS-NSupdate and wildcard signed certificate. With this patch, problem is solved.

jim-p:
Merged manually (ACMEv2 is now live, so I had to push another change anyhow)

PR #508, opened 2018-03-24
Update Suricata to v4.0.4 and fix RTM decoding bug in custom plugin

bmeeks8:
# Suricata v4.0.4
This update to version 4.0.4 of the Suricata binary on pfSense matches the current Suricata upstream release.  Additionally, this update includes a fix for a kernel routing message (RTM) decoding bug that was preventing the automatic firewall interface IP address change monitoring thread from working correctly.

## New Features:
See upstream release notes at [https://suricata-ids.org/2018/02/14/suricata-4-0-4-available/](url).

## Bug Fixes:
1. A kernel routing message decoding bug existed in the _alert-pf_ custom plugin used to implement Legacy Mode blocking in Suricata on pfSense.  The bug prevented the plugin from detecting when a firewall interface IP address changed.  This could lead to blocking of a firewall interface IP because the automatic internal interface IP pass list would not get updated with the new interface IP address.

Bill


rbgarga:
Manually imported. Thanks!

PR #513, opened 2018-04-15
pfBlockerNG v2.1.2_3 - Add tag for DEVEL/REL conflicts

BBcan177:
Added a makefile CONFLICT tag to avoid users from installing both DEVEL and REL versions at the same time.

### DIFF
--- a/net/pfSense-pkg-pfBlockerNG/Makefile
+++ b/net/pfSense-pkg-pfBlockerNG/Makefile
@@ -13,7 +13,7 @@ COMMENT=	pfSense package pfBlockerNG
 
 LICENSE=	APACHE20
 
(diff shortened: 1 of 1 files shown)


net/pfSense-pkg-pfBlockerNG/Makefile
```diff
@@ -13,6 +13,8 @@ COMMENT=	pfSense package pfBlockerNG
 
 LICENSE=	APACHE20
 
+CONFLICTS=	pfBlockerNG-devel
```
rbgarga: The correct package name for conflict is `pfSense-pkg-pfBlockerNG-devel`

BBcan177: Fixed

PR #516, opened 2018-04-20
Add port specification to BIND package

overhacked:
Able to specify BIND DNS port and control port,
which default to 53 and 953 respectively.

Slightly restructured the BIND package settings
screen to have an "Advanced Options" panel where
the new port settings are and also moved "Custom Options"
and "Global Settings" overrides to this panel.

### DIFF
--- a/dns/pfSense-pkg-bind9/files/usr/local/pkg/bind.inc
+++ b/dns/pfSense-pkg-bind9/files/usr/local/pkg/bind.inc
@@ -116,7 +116,7 @@ function bind_sync() {
 	// Create rndc
 	$rndc_confgen = "/usr/local/sbin/rndc-confgen";
 	$rndc_port = $bind['controlport'] == "" ? "953" : $bind['controlport'];
-	$rndc_confgen_opts = "-p$rndc_port";
+	$rndc_confgen_opts = "-p " . escapeshellarg($rndc_port);
 	if (!file_exists(BIND_LOCALBASE."/etc/rndc-confgen.pfsense") && file_exists($rndc_confgen)) {
 		exec("$rndc_confgen $rndc_confgen_opts ", $rndc_conf);
 		$confgen_file = "";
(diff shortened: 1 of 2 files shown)


dns/pfSense-pkg-bind9/files/usr/local/pkg/bind.inc
```diff
@@ -112,10 +112,13 @@ function bind_sync() {
 
 	global $config;
 	conf_mount_rw();
+	$bind = $config['installedpackages']['bind']['config'][0];
 	// Create rndc
 	$rndc_confgen = "/usr/local/sbin/rndc-confgen";
+	$rndc_port = $bind['controlport'] == "" ? "953" : $bind['controlport'];
+	$rndc_confgen_opts = "-p$rndc_port";
```
jim-p: For this style of reference, you should use braces around inline variables, for example this would be `"-p {$rndc_port}"`

However, since this is passing user input to a shell command, it must be escaped better, so do this instead:

    $rndc_confgen_opts = '-p ' . escapeshellarg($rndc_port);

dns/pfSense-pkg-bind9/files/usr/local/pkg/bind.inc
```diff
@@ -227,10 +232,12 @@ EOD;
 	$bind_listenonv6 = $bind_listenonv6 == "" ? "none;" : $bind_listenonv6;
 	$bind_listenon = $bind_listenon == "" ? "none;" : $bind_listenon;
 	// Print "<pre>$bind_listenonv6 $bind_listenon";
+
+	$bind_listenport = $bind['listenport'] == "" ? "53" : $bind['listenport'];
 	if (array_key_exists("ipv6allow", $config['system'])) {
-		$bind_conf .= "\tlisten-on-v6 { $bind_listenonv6 };\n";
+		$bind_conf .= "\tlisten-on-v6 port $bind_listenport { $bind_listenonv6 };\n";
```
jim-p: Use braces around inline variable references (the new one and fix the older reference as well)

dns/pfSense-pkg-bind9/files/usr/local/pkg/bind.inc
```diff
@@ -227,10 +232,12 @@ EOD;
...
 	}
-	$bind_conf .= "\tlisten-on { $bind_listenon };\n";
+	$bind_conf .= "\tlisten-on port $bind_listenport { $bind_listenon };\n";
```
jim-p: Use braces around inline variable references (the new one and fix the older reference as well)

PR #518, opened 2018-05-02
Implements #8490 - "pfSense-pkg-acme: acme_certificates_edit.php - Ad…

mr-intj:
…d ability to specify (vs generate) private key"

This is required to allow certificates to work in environments where HPKP is deployed across subdomains.

UI (Services / Acme / Certificates / Edit Certificate Options:
==============================================================

* Added "Custom..." option at the end of the Key Size drop-down.

* Renamed this field from "Key Size" to "Private Key", since it now contains both generate and paste-custom options.

* Selecting "Custom..." from the list adds (makes visible) a "Custom Private Key" textarea field in the next section.

* Selecting any other of the other (key-size) options from the Private Key drop-down hides the "Custom Private Key"
  field again.

* Key sizes in the drop-down have been renamed for clarity, e.g. "2048" => "2048-bit RSA", "ec-384" => "384-bit ECDSA", etc.

* Fixed a few typos/misspellings in the files I worked on (e.g. "atempting" => "attempting")

Back End:
=========

* Like other user-entered X.509 PEM formatted data, the user-entered Custom Private Key text is stored base64_encoded
  in the configuration.

* The private-key-generation logic is skipped when the user pastes in their private key.

* When visible, "Custom Private Key" is a required field, and is also verified to contain "-----BEGIN PRIVATE KEY-----"
  and "-----END PRIVATE KEY-----" strings.

Additional Notes:
=================

* security/pfSense-pkg-acme/Makefile

  - I bumped the version number from 0.2.8 to 0.2.9. If you're using semver (i.e. last digit is bug fixes), then that
    should probably be 0.3.0.

* security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc:795 - getCertificatePSK function

  - This function appears to be the locus of a bug (needs investigating): Once a private key has been generated for
    a certificate, it will not be re-generated when the certificate is edited, even if a user changes the key-size.

  - This function is misnamed; it doesn't get a Pre-Shared Key, it generates a private key.

### DIFF
--- a/security/pfSense-pkg-acme/Makefile
+++ b/security/pfSense-pkg-acme/Makefile
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-acme
-PORTVERSION=	0.2.8
+PORTVERSION=	0.2.9
 PORTREVISION=	2
 CATEGORIES=	security
 MASTER_SITES=	# empty
--- a/security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc
+++ b/security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc
@@ -676,11 +676,12 @@ $acme_newcertificateactions = array(
 #end
 
 $a_keylength = array();
-$a_keylength['2048'] = array('name' => "2048");
-$a_keylength['3072'] = array('name' => "3072");
-$a_keylength['4096'] = array('name' => "4096");
-$a_keylength['ec-256'] = array('name' => "ec-256", 'ecc' => true);
-$a_keylength['ec-384'] = array('name' => "ec-384", 'ecc' => true);
+$a_keylength['2048'] = array('name' => "2048-bit RSA");
+$a_keylength['3072'] = array('name' => "3072-bit RSA");
+$a_keylength['4096'] = array('name' => "4096-bit RSA");
+$a_keylength['ec-256'] = array('name' => "256-bit ECDSA", 'ecc' => true);
+$a_keylength['ec-384'] = array('name' => "384-bit ECDSA", 'ecc' => true);
+$a_keylength['custom'] = array('name' => "Custom...");
 
 function set_cronjob() {
 	global $config;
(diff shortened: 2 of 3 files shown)


jim-p:
Redmine link: https://redmine.pfsense.org/issues/8490

jim-p:
I have seen that key size change issue before when I was working on https://redmine.pfsense.org/issues/8305 but at the time it seemed like a natural limit since you wouldn't normally change the private key after creating a certificate. It should be possible to detect the change and regenerate, but that might be worthy of a confirmation dialog to make sure it doesn't happen accidentally.

plinss:
> Looks OK, needs some testing though.

I've tested locally and works as advertised.

This was also very useful as I have HPKP deployed on my root domain and it finally allows me to use Let's Encrypt certificates on the firewall.

plinss:
> I have seen that key size change issue before when I was working on https://redmine.pfsense.org/issues/8305 but at the time it seemed like a natural limit since you wouldn't normally change the private key after creating a certificate

Changing the private key periodically is standard practice. Some people replace the private key every time they renew the certificate, though keeping it for a longer period is certainly valid and useful when using HPKP and/or DANE (especially with short-lived certificates like from Let's Encrypt).

However, when the user explicitly changes the key size it would be unexpected for the key to not be regenerated to reflect the new size (or type). A confirmation dialog would be fine.

Additionally, it would be useful to to be able to regenerate the private without having to delete and re-create the certificate (or awkwardly toggle through different sizes). Either an explicit control to regenerate the key (using the same size) and/or and expiration time when the key will automatically be regenerated would be better than simply keeping the same private key forever.

jim-p:
I agree it would be good to have a way to change that I just think it needs a higher barrier than simply changing the size. Perhaps a button that at least does a JS confirm to regenerate a new private key. That's for a different redmine issue/PR though, it's not related closely enough to this to warrant tacking it on.

jim-p:
Tried it out and it seems to test OK here, too.

sbeaver-netgate:
Please resolve the conflicts so that I can merge this PR.

PR #527, opened 2018-06-11
pfSense-pkg-pfBlockerNG-devel v2.2.1_1

BBcan177:
- Fix Extraneous characters
- Adjust CoinBlocker Feed URLs from Github to Gitlab
- Fix EasyList Indonesian and Slovak Feed Category definitions


jim-p:
This was committed, but the PR didn't auto close. See https://github.com/pfsense/FreeBSD-ports/commits/devel

PR #528, opened 2018-06-14
Fix Automagic for TCP based OpenVPN and for port aliases

RolfWojtech:
I stumbled upon two issues with the wonderful Automagic feature:
1) It currently does not create a remote line at all even if you have NAT rules that point to the TCP server.
2) It behaves unexpectedly when dealing with multiple ports inside an alias: it will only use the first one.
This merge request fixes both issues.

### DIFF
--- a/security/pfSense-pkg-openvpn-client-export/files/usr/local/pkg/openvpn-client-export.inc
+++ b/security/pfSense-pkg-openvpn-client-export/files/usr/local/pkg/openvpn-client-export.inc
@@ -994,11 +994,7 @@ function openvpn_client_export_find_port_forwards($targetip, $targetport, $targe
                 $proto_short = strtolower(substr($targetproto, 0, 3));
                 if (!isset($natent['disabled']) && ($natent['target'] == $targetip) && ($natent['local-port'] == $targetport) && ($natent['protocol'] == $proto_short)) {
  	        	$dest['proto'] = $targetproto;
-
-			// Could be multiple ports... But we can only use one.
-			$dports = is_port($natent['destination']['port']) ? array($natent['destination']['port']) : filter_expand_alias_array($natent['destination']['port']);
-			$dest['port'] = $dports[0];
-
+			
 			// Could be network or address ...
 			$natif = (!$natent['interface']) ? "wan" : $natent['interface'];
 
@@ -1034,7 +1030,13 @@ function openvpn_client_export_find_port_forwards($targetip, $targetport, $targe
 				}
 			}
 
-			$destinations[] = $dest;
+                        $dports = is_port($natent['destination']['port']) ? array($natent['destination']['port']) : filter_expand_alias_array($natent['destination']['port']);
+ 
+			// Could be multiple ports, we add all of them
+			foreach ($dports as $dport) {
+				$dest['port'] = $dport;
+				$destinations[] = $dest;
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


security/pfSense-pkg-openvpn-client-export/files/usr/local/pkg/openvpn-client-export.inc
```diff
@@ -991,13 +991,10 @@ function openvpn_client_export_find_port_forwards($targetip, $targetport, $targe
...
 	foreach ($config['nat']['rule'] as $natent) {
 		$dest = array();
-		if (!isset($natent['disabled']) && ($natent['target'] == $targetip) && ($natent['local-port'] == $targetport) && ($natent['protocol'] == $targetproto)) {
-			$dest['proto'] = $natent['protocol'];
-
-			// Could be multiple ports... But we can only use one.
-			$dports = is_port($natent['destination']['port']) ? array($natent['destination']['port']) : filter_expand_alias_array($natent['destination']['port']);
-			$dest['port'] = $dports[0];
-
+                $proto_short = strtolower(substr($targetproto, 0, 3));
```
jim-p: Since this variable is only used once, skip setting it this way and use the contents in the next line directly instead.

RolfWojtech:
Thanks for your review, I made the change you requested (personally I prefer having the variable for the next person that stumbles upon this code, but I concede that it wastes a line).

RolfWojtech:
By the way I made the observation that the  network manager in Ubuntu 16.04 cannot handle multiple multiple remotes. However the existing code already generates multiple remotes if you use multi-WAN or two ports for the same server.

security/pfSense-pkg-openvpn-client-export/files/usr/local/pkg/openvpn-client-export.inc
```diff
@@ -991,13 +991,9 @@ function openvpn_client_export_find_port_forwards($targetip, $targetport, $targe
...
-		if (!isset($natent['disabled']) && ($natent['target'] == $targetip) && ($natent['local-port'] == $targetport) && ($natent['protocol'] == $targetproto)) {
-			$dest['proto'] = $natent['protocol'];
-
-			// Could be multiple ports... But we can only use one.
-			$dports = is_port($natent['destination']['port']) ? array($natent['destination']['port']) : filter_expand_alias_array($natent['destination']['port']);
-			$dest['port'] = $dports[0];
-
+                if (!isset($natent['disabled']) && ($natent['target'] == $targetip) && ($natent['local-port'] == $targetport) && ($natent['protocol'] == strtolower(substr($targetproto, 0, 3)))) {
+ 	        	$dest['proto'] = $targetproto;
+
```
rbgarga: Please keep the indentation style using hard TABs instead of replacing it by spaces

security/pfSense-pkg-openvpn-client-export/files/usr/local/pkg/openvpn-client-export.inc
```diff
@@ -1033,7 +1029,13 @@ function openvpn_client_export_find_port_forwards($targetip, $targetport, $targe
 				}
 			}
 
-			$destinations[] = $dest;
+                        $dports = is_port($natent['destination']['port']) ? array($natent['destination']['port']) : filter_expand_alias_array($natent['destination']['port']);
```
rbgarga: Same here, keep TABs

PR #529, opened 2018-06-14
Move key-direction infront of tls-auth for old network-manager-openvp…

RolfWojtech:
OpenVPN configs generated by pfsense currently do not work in network-manager-openvpn of Ubuntu 16.04 because of a bug: https://bugs.launchpad.net/ubuntu/+source/network-manager-openvpn/+bug/1643282
One could argue that pfsense generates a perfectly valid config thus no change is needed but Ubuntu 16.04 is still very widespread and supported until 2021 and it does seem to affect many users (the error message is really unclear).

### DIFF
--- a/security/pfSense-pkg-openvpn-client-export/files/usr/local/pkg/openvpn-client-export.inc
+++ b/security/pfSense-pkg-openvpn-client-export/files/usr/local/pkg/openvpn-client-export.inc
@@ -464,7 +464,7 @@ EOF;
 					$tls_directive = "tls-auth";
 					$tls_keydir = "key-direction 1{$nl}";
 				}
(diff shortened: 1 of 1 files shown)


jim-p:
Has this been tested and proven to work with other clients? I get wanting to fix Ubuntu but we need to be certain that it doesn't break any others.

RolfWojtech:
I have tested it with the Windows OpenVPN GUI and it works fine there as well. In general these directions are not supposed to have an order so yes, this Pull Request would just be working around a bug in network manager, very likely not affecting any other clients. (https://forums.openvpn.net/viewtopic.php?t=7964 )

jim-p:

Thanks!

jim-p:
I committed a slightly different version of this, it will be in snapshots for testing soon!

PR #531, opened 2018-07-03
FauxAPI v1.3

ndejong:
I'm submitting the FauxAPI package for consideration again, in the past 18 months the package and code has remained stable and reliable across 2.3.x and 2.4.x revisions.

At minimum, review the documentation to understand the approach and the underlying purpose of the "Faux" API that aims to provide a way to edit and manage a pfSense configuration via REST calls:-
https://github.com/ndejong/pfsense_fauxapi/blob/master/README.md

Regards,
N


hooray4me:
any movement on this?

gonzopancho:
We thank you for the contribution, but based on our internal roadmap for pfSense, we are unable to bring this into the fold at this time. We are more than willing to consider it in the future as time and effort permits as there may be some value in doing so.  Right now it would only detract from developmental goals of the product.  Thank you for your support.

PR #534, opened 2018-07-09
Set standard RADIUS attributes ouside of the RADIUS class

Augustin-FL:
Radius Attributes are currently not used in a common/standard way by pfSense Authentication components (currently OpenVPN, IPsec xAuth and CaptivePortal) when performing an Authentication.

In an attempt to clarify RADIUS PHP code, i suggested in pfsense/pfsense#3640 to standardize the way attributes are used, but outside of the RADIUS class, in ```auth.inc```. This would make normalization easier, and also make the code more clear (i am not sure that using functions like ```getNasIp()``` to  retrieve variables is a best practice. Shouldn't arguments be used instead)?


net/pear-Auth_RADIUS/files/patch-Auth_RADIUS.php
```diff
@@ -12,53 +12,20 @@
...
+@@ -295,11 +300,7 @@ class Auth_RADIUS {
+             $var = $GLOBALS['HTTP_SERVER_VARS'];
+         }
+                 
+-        $this->putAttribute(RADIUS_NAS_IDENTIFIER, isset($var['HTTP_HOST']) ? $var['HTTP_HOST'] : 'localhost');
+-        $this->putAttribute(RADIUS_NAS_PORT_TYPE, RADIUS_VIRTUAL);
+-        $this->putAttribute(RADIUS_SERVICE_TYPE, RADIUS_FRAMED);
+-        $this->putAttribute(RADIUS_FRAMED_PROTOCOL, RADIUS_PPP);
+-        $this->putAttribute(RADIUS_CALLING_STATION_ID, isset($var['REMOTE_HOST']) ? $var['REMOTE_HOST'] : '127.0.0.1');
++        $this->putAttribute(RADIUS_SERVICE_TYPE, RADIUS_LOGIN);
```
rbgarga: Why are you removing attributes that are assigned by the original port? Less changes we need to carry on it are better.

Augustin-FL: - For ```Service-Type``` and ```Framed-Protocol``` attributes : because these 2 attributes are wrong : they indicate that pfSense is querying Radius server in order to make a PPP connection. These two attributes were also not assigned in old ```RADIUS.inc```. 
- For ```Calling-Station-ID``` : The new RADIUS port broke this attribute : it is currently set to ```127.0.0.1```. I suggested in pfsense/pfsense#3640 to set this attribute [here in auth.inc](https://github.com/pfsense/pfsense/pull/3640/files#diff-647553ae2000651d4860da0cbf861caeR1651) _(for user manager authentication)_ and [here in captiveportal.inc](https://github.com/pfsense/pfsense/pull/3640/files#diff-764cc5453dc616c068ba7adeb68ba73eR2669) _(for captiveportal accounting)_, so i am removing it from the RADIUS class.

Augustin-FL:
@rbgarga I think it should be more clear in a packet capture : 
image

This is why i removed attributes assigned by the original port. Please let me know if i you want me to modify something in this pull request

PR #535, opened 2018-07-09
haproxy-devel 0.59, several fixes and improvements..

PiBa-NL:


satmandu:
Thanks for putting this together. The haproxy web gui as it stands now really doesn't like php 7.2 in the current dev builds.

PR #549, opened 2018-08-11
Fix PHP7 syntax issue in pfSense-pkg-cron

BBcan177:
PHP Warning:  Use of undefined constant info - assumed 'info' (this will throw an Error in a future version of PHP) in /usr/local/www/packages/cron/cron.php on line 115


BBcan177:
Updated PR to include commit for cron_edit.php

PR #553, opened 2018-08-19
acme, add challenge alias feature for dns validation with CNAME

PiBa-NL:
acme, add challenge alias feature for dns validation with CNAME (https://redmine.pfsense.org/issues/8613)
-support different validation methods within the domains for 1 certificate
-add notification and syslog message when renewal fails
-minor gui improvements

### DIFF
--- a/security/pfSense-pkg-acme/Makefile
+++ b/security/pfSense-pkg-acme/Makefile
@@ -1,8 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-acme
-PORTVERSION=	0.3.1
-PORTREVISION=	1
+PORTVERSION=	0.3.2
 CATEGORIES=	security
 MASTER_SITES=	# empty
 DISTFILES=	# empty
--- a/security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc
+++ b/security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc
@@ -655,6 +655,12 @@ $acme_domain_validation_method['dns_zilore'] = array('name' => "DNS-Zilore",
 			'description' =>"Fill in the Zilore API Key"
 		)
 	));
+$acme_domain_validation_method['anydns'] = array('name' => "notforuser",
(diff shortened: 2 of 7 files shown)


security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme_utils.inc
```diff
@@ -53,6 +53,27 @@ if(!function_exists('is_arrayset')){
 	}
 }
 
+if (!function_exists('getarraybyref')) {
```
jim-p: I know it can make things easier for portability but I'd really rather not have the same function redefined in several packages like this again. Too many bad memories of pfSense 2.2 and before with copy/pasted code.

Let's get this function into base and then only use this version of the package on 2.4.4 and later.

PiBa-NL: so how do we get this version of the package into 2.4.3 ? there is no reason to restrict the new abilities to 2.4.4 just because of one little 'helper function'..? perhaps we can mark that part of the code 'to be removed after 2.4.4 gets released' ? it would imho be nice if this could be available for everyone in a few days time.

PiBa-NL: well okay, removed from here.

jim-p:
Waiting on https://github.com/pfsense/pfsense/pull/3973 to be merged first

rbgarga:
@jim-p what if we carry a copy of `getarraybyref()` inside a `if (!function_exists())` conditional for a while to prevent users to face errors when it is not present on that specific version of snapshot? And also make the same code to work on 2.4.3 while it's still supported?

After 2.4.4 is released we can do a single commit to remove it from all packages

jim-p:
We can but I really don't like having the same code copied and pasted around like that anymore.
If we were going to start doing something like that long-term I'd say make a port with an .inc file we can have all the packages depend on rather than having dozens of copies of the same function. Too many bad memories of PBI and shared package code base days.

PR #555, opened 2018-08-21
haproxy-devel, php7 array usage.. drop init() and !isarray(x)x=array() , instead use getarraybyref() where required.

PiBa-NL:
haproxy-devel, php7 array usage.. drop init() and !isarray(x)x=array() , instead use getarraybyref() where required.

### DIFF
--- a/net/pfSense-pkg-haproxy-devel/files/usr/local/pkg/haproxy/haproxy_utils.inc
+++ b/net/pfSense-pkg-haproxy-devel/files/usr/local/pkg/haproxy/haproxy_utils.inc
@@ -102,25 +102,6 @@ if(!function_exists('is_arrayset')){
 	}
 }
 
-function &getarraybyref(&$array) {
-	if (!isset($array)) {
-		return false;
-	}
-	if (!is_array($array)) {
-		$array = array();
-	}
-	$item = &$array;
-	$arg = func_get_args();
-	for($i = 1; $i < count($arg); $i++) {
-		$itemindex = $arg[$i];
-		if (!is_array($item[$itemindex])) {
-			$item[$itemindex] = array();
-		}
-		$item = &$item[$itemindex];
-	}
-	return $item;
-}
-
 function haproxy_compareByName($a, $b) {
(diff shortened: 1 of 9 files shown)


net/pfSense-pkg-haproxy-devel/files/usr/local/pkg/haproxy/haproxy_utils.inc
```diff
@@ -102,25 +102,6 @@ if(!function_exists('is_arrayset')){
...
-	for($i = 1; $i < count($arg); $i++) {
-		$itemindex = $arg[$i];
-		if (!is_array($item[$itemindex])) {
-			$item[$itemindex] = array();
-		}
-		$item = &$item[$itemindex];
-	}
-	return $item;
-}
-
```
rbgarga: Please don't remove it. Add a `if (!function_exists())` conditional so we can cherry-pick it to 2.4.3 and keep packages the same on both branches as much as possible. It makes our lives easier to maintain packages on branches based on a single version

PiBa-NL: @jim-p & @rbgarga can you decide one way or the other, seems i'm getting mixed messages (or i misinterpreted the intended size of effect).
See same kinda topic which lead me to delete the function here from haproxy as well: https://github.com/pfsense/FreeBSD-ports/pull/553#discussion_r211390033

As for a personal preference, i did like the ```if (!function_exists())``` approach..

rbgarga: Sorry about that @PiBa-NL, we will discuss it and come up with a solution soon.

bott0r:
please approve !

net/pfSense-pkg-haproxy-devel/files/usr/local/pkg/haproxy/haproxy_utils.inc
rbgarga: Leave it as is please.

PR #556, opened 2018-08-23
pfSense-pkg-pfBlockerNG-devel -  Fix previous PHPv7 commits.

BBcan177:
The following settings 'v4suppression', 'suppression' and 'tldexclusion' are not arrays.


BBcan177:
This was incorrectly patched via the following commit:
https://github.com/pfsense/FreeBSD-ports/commit/177218c9e911522cd33c0462751098f2b7c975f5#diff-101d379a1dcad26bdf90a3b701c2ee19

PR #557, opened 2018-08-23
Adding OSPF cost-shifting as a CARP failover mechanism for scenarios requiring rapid failover in HA pairs.

quadrinary:
This is an rebased version of [Pull Request #413](https://github.com/pfsense/FreeBSD-ports/pull/413)

Code remains the same. 

This latest modification to the Quagga package allows users to select the previously existing CARP mode that outright disables the Quagga service or use a hot/warm configuration wherein routers in a HA pair have their OSPF interface costs shifted according to CARP status, allowing for rapid transition from one router to another. Crosslinked in redmine: [](https://redmine.pfsense.org/issues/8181)


net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -344,19 +344,59 @@ function quagga_ospfd_install_conf() {
...
 	fclose($fd);
 
+	//UPGRADE PATH - Checks if legacy "CARP Disable" mode was previously in use proior to 0.6.20; if previously in use, sets "carpmode" to "quaggadisable". If not in use, sets "carpmode" to "none".
+	if (!isset($ospfd_conf['carpmode'])) {
+		syslog(LOG_NOTICE, "Quagga: CARP failover mode upgrade initializing");
+		if (strpos($ospfd_conf['carpstatusvid'],'_vip') !== false) {
+			$ospfd_conf['carpmode'] = "quaggadisable";
+			syslog(LOG_NOTICE, "Quagga: CARP failover mode established as 'quaggadisable'");
+		}
+		else {
```
rbgarga: Move else to same line of }, it's desired style in pfSense

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -344,19 +344,59 @@ function quagga_ospfd_install_conf() {
...
+			$ospfd_conf['carpmode'] = "none";
+			syslog(LOG_NOTICE, "Quagga: CARP failover mode established as 'none'");
+		}
+		write_config( $write_config_only = true );
```
rbgarga: `write_config()` expects a string parameter saying the reason of that change to make it easier for users to identify changes.

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -344,19 +344,59 @@ function quagga_ospfd_install_conf() {
...
 CARP_STATUS=`/sbin/ifconfig {$carpcheckinterface} | /usr/bin/grep 'carp:' | /usr/bin/grep {$vhid} | /usr/bin/awk '{print \$2;}'`
 if [ \${CARP_STATUS} != "MASTER" ]; then
+	logger "Quagga: CARP \"Quagga Disable\" Mode - Interface {$carpcheckinterface} is NOT in MASTER state, exiting";
 	exit;
 fi
 EOF;
 	}
+	
+	$carp_ospfcost_status_check = "";
+	if ((isset($ospfd_conf['carpmode']) && $ospfd_conf['carpmode'] == "ospfcost") && (isset($ospfd_conf['carpcostvid']) && substr($ospfd_conf['carpcostvid'],0,4) != "none") && (isset($ospfd_conf['carpactivecost']) && isset($ospfd_conf['carpbackupcost'])) && ($ospfd_conf['carpactivecost'] >= 0 && $ospfd_conf['carpactivecost'] <= 65535) && ($ospfd_conf['carpbackupcost'] >= 0 && $ospfd_conf['carpbackupcost'] <= 65535)) {
```
rbgarga: Please break this really long line, it's hard to read. Use 4 spaces to indent multi-line statements

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -514,24 +557,82 @@ function quagga_ospfd_plugin_carp($pluginparams) {
...
-	$friendly = convert_real_interface_to_friendly_interface_name($iface);
-	$vip = get_configured_vip($ospfd_conf['carpstatusvid']);
-	if ($vip['vhid'] != $vhid || $vip['interface'] != $friendly) {
+	/* Verify "quaggadisable" parameters are properly set.*/
+	if ( $ospfd_conf['carpmode'] == "quaggadisable" && (!isset($ospfd_conf['carpstatusvid']) || $ospfd_conf['carpstatusvid'] == "none")) {
+		syslog(LOG_ALERT, "Quagga: Failed checking CARP mode parameters: {$ospfd_conf['carpmode']}");
 		return null;
 	}
+	/* Verify "carpmode" parameters are properly set.*/
+	if ( $ospfd_conf['carpmode'] == "ospfcost" && (!isset($ospfd_conf['carpcostvid']) || substr($ospfd_conf['carpcostvid'],0,4) == "none") || !isset($ospfd_conf['carpactivecost']) || !isset($ospfd_conf['carpbackupcost']) || ($ospfd_conf['carpactivecost'] < 0 || $ospfd_conf['carpactivecost'] > 65535) || ($ospfd_conf['carpbackupcost'] < 0 || $ospfd_conf['carpbackupcost'] > 65535)) {
```
rbgarga: Break long line

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -555,8 +656,8 @@ function write_quagga_running_config($module) {
...
 }
 
-write_quagga_running_config('bgpd');
-write_quagga_running_config('ospfd');
-write_quagga_running_config('ospf6d');
-write_quagga_running_config('zebra');
+write_quagga_running_config(bgpd);
+write_quagga_running_config(ospfd);
+write_quagga_running_config(ospf6d);
+write_quagga_running_config(zebra);
```
rbgarga: Why did you remove quotes here?

net/pfSense-pkg-Quagga_OSPF/files/usr/local/www/status_ospfd.php
```diff
@@ -78,9 +78,9 @@ function doCmdT($title, $command) {
 defCmdT("Quagga BGP IPv6 Routes", "{$control_script} bgp6 route");
 defCmdT("Quagga BGP Neighbors", "{$control_script} bgp neighbor");
 defCmdT("Quagga BGP Summary", "{$control_script} bgp sum");
-defCmdT("Quagga ospfd.conf", "/bin/cat {$pkg_homedir}/ospfd.conf");
```
jim-p: Showing the configuration file and the running configuration are two completely different and useful things. The running config is what was interpreted by quagga, so it may differ from the configuration file, and that can aid in tracking down problems with a configuration. You can add the commands to show the running configuration, but do not remove the commands to show the configuration files.

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
quadrinary: Very good question. My production versions haven't had the quotes, so they must've been added in after I submitted the original code 18 months ago that included the "write_quagga_running_config()" function. I'll do some testing on my end to ensure it doesn't impact functionality.

net/pfSense-pkg-Quagga_OSPF/files/usr/local/www/status_ospfd.php
quadrinary: I understand your reasoning here. 

My logic behind making it a pure "show running-config" vs. the file that was loaded when the daemon started is that we wanted to see at a glance if an OSPF cost had been set dynamically by viewing the running configuration directly. 

With that in mind, I can do as you suggested and show both.

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
quadrinary: Will do.

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
quadrinary: Will do.

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
quadrinary: Will do.

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
quadrinary: OK - how would this look, something along the lines of "write_config( "Applied upgrade script" )"?

I had pulled the "write_config( $write_config_only = true )" function from another part of the original plugin and wasn't entirely sure how it was to be used or what it was expecting.

quadrinary:
I've made and tested the requested changes and normalized the multi-line functions as they had been done in different manners over time.

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -78,17 +78,17 @@ function quagga_ospfd_install_conf() {
 	// generate ospfd.conf based on the assistant
 	if (is_array($config['installedpackages']['quaggaospfd']['config'])) {
 		$ospfd_conf = &$config['installedpackages']['quaggaospfd']['config'][0];
-	} elseif (isset($config['installedpackages']['quaggaospfdraw']['config'][0]['ospfd'])  ||
-	    isset($config['installedpackages']['quaggaospfdraw']['config'][0]['ospf6d']) ||
-	    isset($config['installedpackages']['quaggaospfdraw']['config'][0]['bgpd'])) {
+	} elseif (isset($config['installedpackages']['quaggaospfdraw']['config'][0]['ospfd'])
+		|| isset($config['installedpackages']['quaggaospfdraw']['config'][0]['ospf6d'])
```
jim-p: These should be [indented with four spaces](https://www.netgate.com/docs/pfsense/development/developer-style-guide.html#indent-style)

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -78,17 +78,17 @@ function quagga_ospfd_install_conf() {
...
+		|| isset($config['installedpackages']['quaggaospfdraw']['config'][0]['bgpd'])) {
 		log_error("Quagga: No assistant generated config for OSPF, but found raw config for one or more daemon");
 	} else {
 		log_error("Quagga OSPFd: No config data found.");
 		return;
 	}
 
 	if (isset($config['installedpackages']['quaggaospfdraw']['config'][0]['ospfd'])
-	    && !empty($config['installedpackages']['quaggaospfdraw']['config'][0]['ospfd'])) {
+		&& !empty($config['installedpackages']['quaggaospfdraw']['config'][0]['ospfd'])) {
```
jim-p: These should be [indented with four spaces](https://www.netgate.com/docs/pfsense/development/developer-style-guide.html#indent-style)

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -266,7 +266,7 @@ function quagga_ospfd_install_conf() {
 
 	/* Make zebra config */
 	if (isset($config['installedpackages']['quaggaospfdraw']['config'][0]['zebra'])
-	    && !empty($config['installedpackages']['quaggaospfdraw']['config'][0]['zebra'])) {
+		&& !empty($config['installedpackages']['quaggaospfdraw']['config'][0]['zebra'])) {
```
jim-p: These should be [indented with four spaces](https://www.netgate.com/docs/pfsense/development/developer-style-guide.html#indent-style)

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -291,11 +291,10 @@ function quagga_ospfd_install_conf() {
 
 	/* Make bgpd config, add password and logging */
 	if (isset($config['installedpackages']['quaggaospfdraw']['config'][0]['bgpd'])
-	    && !empty($config['installedpackages']['quaggaospfdraw']['config'][0]['bgpd'])) {
+		&& !empty($config['installedpackages']['quaggaospfdraw']['config'][0]['bgpd'])) {
```
jim-p: These should be [indented with four spaces](https://www.netgate.com/docs/pfsense/development/developer-style-guide.html#indent-style)

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -324,39 +323,81 @@ function quagga_ospfd_install_conf() {
...
 		}
-
 		if ($bgpddelmd5file != "") {
 			$fd = fopen("{$quagga_config_base}/bgpddelmd5pw.conf", "w");
 			fwrite($fd, $bgpddelmd5file);
 			fclose($fd);
 		}
 	}
-        /* Make ospf6 config */
+		/* Make ospf6 config */
```
jim-p: Why the extra tab here?

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -324,39 +323,81 @@ function quagga_ospfd_install_conf() {
...
 	if (isset($config['installedpackages']['quaggaospfdraw']['config'][0]['ospf6d'])
 	    && !empty($config['installedpackages']['quaggaospfdraw']['config'][0]['ospf6d'])) {
 		// if there is a raw config specified in the config.xml use that instead of the assisted config
 		$ospf6dconffile = str_replace("\r","",base64_decode($config['installedpackages']['quaggaospfdraw']['config'][0]['ospf6d']));
 	} else {
-        	$ospf6dconffile = "";
+			$ospf6dconffile = "";
```
jim-p: Looks like maybe one tab too many here as well

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -324,39 +323,81 @@ function quagga_ospfd_install_conf() {
...
 if [ \${CARP_STATUS} != "MASTER" ]; then
+	logger "Quagga: CARP \"Quagga Disable\" Mode - Interface {$carpcheckinterface} is NOT in MASTER state, exiting";
 	exit;
 fi
 EOF;
 	}
+	
+	$carp_ospfcost_status_check = "";
+	if ((isset($ospfd_conf['carpmode']) && $ospfd_conf['carpmode'] == "ospfcost")
+		&& (isset($ospfd_conf['carpcostvid']) && substr($ospfd_conf['carpcostvid'],0,4) != "none")
```
jim-p: All of these conditions should be [indented with four spaces](https://www.netgate.com/docs/pfsense/development/developer-style-guide.html#indent-style)

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -514,24 +558,86 @@ function quagga_ospfd_plugin_carp($pluginparams) {
 	if (is_array($config['installedpackages']['quaggaospfd']['config'])) {
 		$ospfd_conf = &$config['installedpackages']['quaggaospfd']['config'][0];
 	} else {
+		syslog(LOG_ALERT, "Quagga: CONFIGURATION DOES NOT EXIST!");
```
jim-p: Is that really worth logging an error for? The user may have it installed but not configured yet. Doesn't seem fatal.

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -514,24 +558,86 @@ function quagga_ospfd_plugin_carp($pluginparams) {
 	if (is_array($config['installedpackages']['quaggaospfd']['config'])) {
 		$ospfd_conf = &$config['installedpackages']['quaggaospfd']['config'][0];
```
jim-p: On 2.4.4 and later, this syntax can be dangerous with PHP 7.2 if the user doesn't have any settings.

Check out [init_config_arr()](https://github.com/pfsense/pfsense/blob/master/src/etc/inc/config.lib.inc#L985) for a good way around this, for example, instead of the `is_array()` test, use:

    init_config_arr(array('installedpackages', 'quaggaospfd', 'config', 0));
    $ospfd_conf = &$config['installedpackages']['quaggaospfd']['config'][0];

Do that any time there is an assignment by reference to a multi-level hash like this.

(I know this wasn't part of your commits, but since you're working on it anyhow...)

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
```diff
@@ -514,24 +558,86 @@ function quagga_ospfd_plugin_carp($pluginparams) {
...
-	$vip = get_configured_vip($ospfd_conf['carpstatusvid']);
-	if ($vip['vhid'] != $vhid || $vip['interface'] != $friendly) {
+	/* Verify "quaggadisable" parameters are properly set.*/
+	if ( $ospfd_conf['carpmode'] == "quaggadisable" && (!isset($ospfd_conf['carpstatusvid']) || $ospfd_conf['carpstatusvid'] == "none")) {
+		syslog(LOG_ALERT, "Quagga: Failed checking CARP mode parameters: {$ospfd_conf['carpmode']}");
+		return null;
+	}
+	/* Verify "carpmode" parameters are properly set.*/
+	if ( $ospfd_conf['carpmode'] == "ospfcost" && (!isset($ospfd_conf['carpcostvid'])
+		|| substr($ospfd_conf['carpcostvid'],0,4) == "none")
```
jim-p: All of these conditions should be [indented with four spaces](https://www.netgate.com/docs/pfsense/development/developer-style-guide.html#indent-style)

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
quadrinary: This error is to alert a user in the event that they have selected to follow a CARP interface for the purposes of OSPF changes, but have not actually created an OSPF configuration. Since the default state is "off" for CARP tracking, we found it useful in our logging to be sure we didn't unnecessarily have it set to follow CARP events.

jim-p: But that will be executed even if the user hasn't chosen to follow a CARP interface. It will trigger for any CARP event where the user has the package installed but not configured. That isn't an error state.

quadrinary: I see what you're saying. Sorry, it's been over a year since this was written and I haven't looked at it since then. If I remember correctly, this was some debug logging from making sure all configuration variations produced the expected behavior. We can probably safely remove this.

net/pfSense-pkg-Quagga_OSPF/files/usr/local/pkg/quagga_ospfd.inc
quadrinary: I need to think about this a bit more since init_config_arr() did not exist on earlier versions of pfsense, and would immediately break the version we are running. Which production version of pfSense was the first to receive this function? I can setup a conditional to use is_array() in older versions.

quadrinary: Based on what I'm observing, Init_config_arr() was only introduced in 2.4.4 (config revision 18.8 based on this site https://www.netgate.com/docs/pfsense/releases/versions-of-pfsense-and-freebsd.html ). I'll include a parameter to check the config revision version and if it's 18.8 or greater, it'll use init_config_arr(). If less, it'll use is_array(). 

Side note - is relying on $config['version'] a suitable solution to determine the pfSense version in play or is there a better way?

jim-p: You won't be able to install the new version of the package on versions older than 2.4.4, so a version test is unnecessary in the copy of the package that gets committed.

Aside from that, the best way to check the version is to look at `$g['product_version']` and then use one of the version compare functions like `version_compare_string()`.

quadrinary: Ok, great thanks for the clarification about version limitations, I won't worry about it then.  

Does that mean that the current version of Quagga that is installable on 2.4.4 will cause issues, since it still uses is_array()?

Digging deeper though, there are other cases where is_array() is leveraged, do all of these need to be addressed?

jim-p: I haven't seen any issues with Quagga on 2.4.4 but the array issues tend to happen on fresh installations (no related config present) and the ones I have were upgraded. It's possible there are problems lurking there but nobody has happened to hit them yet.

Cases that need addressed are primarily anything that uses assignment by reference to a multi-level array like `$foo &= $bar['baz']['blah']`. If it uses a copy instead (No `&`) then PHP is more forgiving.

quadrinary: Got it. I could see where this particular use of is_array() could cause a problem as it's called during any CARP event, whereas the rest are assuming a config is either being built or has been built. That said, I'll test it with a new install and verify init_config_arr() works as expected in this situation.

quadrinary: I just tested with a new install of 2.4.4, configured a CARP address, and added the existing Quagga version, which relies on is_array(). I added logging to the conditions in question to see what happens. I performed a number of CARP tests and no PHP crashes occurred in the cases where there was no Quagga config and in the case where there was a Quagga config. 

In looking at init_config_arr(), I'm trying to understand how it should be applied in this case, since the purpose of the if-else statement is to check if a config actually exists. If we have the init_config_arr() function create an array, then won't it artificially pass the test every time? Or would we replace the if-else with the array initialization and let it go to the subsequent tests?

jim-p: In that kind of case, the is_array() test may be OK but you still need to initialize the array before assigning by reference. It may be safe in certain cases and not in others, depends on the state of the user configuration. Just something to look out for. We've been fixing a lot of similar cases in the base and packages ever since the switch to PHP 7.2.

quadrinary: Is this the optimal place to initialize the array then? Should it be initialized somewhere else besides the CARP function call?

jim-p: As long as it gets initialized before assigning by reference, it doesn't matter much. You could initialize first and then change the `is_array()` check to an `!empty()` check.

rbgarga:
@quadrinary sorry for the long delay on this PR. Could you please rebase your fork so we can review and test it?

rbgarga:
Superseded by #769

PR #566, opened 2018-09-11
New version of the Avahi package with a redesigned interface.

dennypage:
A completely new version of the Avahi package.

The new version offers the following improvements:

- A much cleaner and simpler interface.
- Uses positive interface selection rather than negative selection. This prevents security surprises when adding interfaces.
- Addresses an information leakage issue with OpenVPN connections. 
- Removes run-time dependance upon dbus, reducing memory and improving service startup time (no more 5 second delay).
- Disables local cache which caused hostname collision issues with reflection.

This package has been tested with 2.4.3 (prod) and 2.4.4 (dev).

Thanks,
Denny


dennypage:
I've now tested with 2.4.4 release. All seems good with the package itself, but I did notice something unusual. With the package installed via the command line interface (pkg install pfSense-pkg-Avahi-2.0.0.txz), the package no longer shows up in the list of installed packages in the UI. In 2.4.3-p1, it would show up with an exclamation mark next to the version number indicating that the version installed was newer than the version in the package repo. Is this intended? Is this something I need to fix?

jim-p:
That is normal for packages that didn't come from pfSense. See the release blog post, release notes, and upgrade guide which all have a big warning about how packages from other repositories behave now on 2.4.4.

dennypage:
"Packages from alternate repositories will not appear in the Installed Packages list in the GUI, and must be entirely managed in the command line."

Sorry Jim, I just didn't read down far enough. My bad.

net/pfSense-pkg-Avahi/files/etc/inc/priv/avahi.priv.inc
```diff
@@ -3,7 +3,7 @@
  * avahi.priv.inc
  *
  * part of pfSense (https://www.pfsense.org)
- * Copyright (c) 2015 Rubicon Communications, LLC (Netgate)
+ * Copyright (c) 2018 Denny Page
```
rbgarga: Don't remove Rubicon Copyright

net/pfSense-pkg-Avahi/files/usr/local/pkg/avahi.xml
```diff
@@ -3,12 +3,12 @@
...
 <packagegui>
 	<copyright>
-<![CDATA[
+	<![CDATA[
 /*
  * avahi.xml
  *
  * part of pfSense (https://www.pfsense.org)
- * Copyright (c) 2009-2015 Rubicon Communications, LLC (Netgate)
+ * Copyright (c) 2018 Denny Page
```
rbgarga: Don't remove Rubicon Copyright

net/pfSense-pkg-Avahi/files/usr/local/pkg/avahi/avahi.inc
```diff
@@ -0,0 +1,128 @@
+<?php
+/*
+ * avahi.inc
+ *
+ * part of pfSense (https://www.pfsense.org)
+ * Copyright (c) 2018 Denny Page
```
rbgarga: Don't remove Rubicon Copyright

net/pfSense-pkg-Avahi/pkg-plist
```diff
@@ -1,7 +1,6 @@
 pkg/avahi.xml
-pkg/avahi.inc
+pkg/avahi/avahi.inc
```
rbgarga: Why move .inc file to a separate directory?

net/pfSense-pkg-Avahi/files/etc/inc/priv/avahi.priv.inc
dennypage: I didn't remove the copyright--in effect it wasn't there to begin with. This package is a complete ground up writing, unrelated to the prior package. This particular file was initialized by copying and modifying the lldld.priv.inc.

net/pfSense-pkg-Avahi/files/usr/local/pkg/avahi.xml
dennypage: I didn't remove the copyright--in effect it wasn't there to begin with. This package is a complete ground up writing, unrelated to the prior package. This particular file was initialized by copying and modifying lldld.xml.

net/pfSense-pkg-Avahi/files/usr/local/pkg/avahi/avahi.inc
dennypage: I didn't remove the copyright--in effect it wasn't there to begin with. This package is a complete ground up writing, unrelated to the prior package. This particular file was written completely from scratch.

net/pfSense-pkg-Avahi/pkg-plist
dennypage: Same pattern used in other packages, which allows the introduction of additional files in the future. Same as with the nut and lldpd packages.

PR #575, opened 2018-09-27
pfBlockerNG-devel - Revert symlinks and add force delete of php_pfb

BBcan177:


net/pfSense-pkg-pfBlockerNG-devel/files/usr/local/pkg/pfblockerng/pfblockerng_install.inc
```diff
@@ -159,9 +159,12 @@ if ($pfb['keep'] == 'on' && isset($pfb['widgets']) && strpos($pfb['widgets'], 'p
 // Create Firewall filter service and link required executables
 update_status("\nRemove any existing and create links for Firewall filter executables...");
 unlink_if_exists('/usr/local/bin/php_pfb');
-symlink('/usr/local/bin/php', '/usr/local/bin/php_pfb');
+if (file_exists('/usr/local/bin/php_pfb')) {
```
rbgarga: `unlink_if_exists()` is called above, why add this `exec()` call here?

BBcan177: Its a failsafe since some users are reporting that a `php_pfb -v` reports an older version of PHP.

It should have been deleted with the unlink() but that cmd might not do a force if its locked etc...

net/pfSense-pkg-pfBlockerNG-devel/files/usr/local/pkg/pfblockerng/pfblockerng_install.inc
```diff
@@ -158,10 +158,12 @@ if ($pfb['keep'] == 'on' && isset($pfb['widgets']) && strpos($pfb['widgets'], 'p
 
 // Create Firewall filter service and link required executables
 update_status("\nRemove any existing and create links for Firewall filter executables...");
-unlink_if_exists('/usr/local/bin/php_pfb');
-symlink('/usr/local/bin/php', '/usr/local/bin/php_pfb');
+if (file_exists('/usr/local/bin/php_pfb')) {
+	exec('/bin/rm -f /usr/local/bin/php_pfb 2>&1', $output, $retval);
+}
```
rbgarga: I'm still not convinced about replacing `unlink_if_exists()` call by a shell session running `rm -f`. I would like to understand why former is failing instead of using a hammer. Are you able to reproduce the failure? If yes, could you replace `unlink_if_exists()` call by `unlink()` and see what is the error it reports

BBcan177: I don't have a system to reproduce this issue. All I can say, is that some users when they ran "php_pfb -v", it showed the PHP5 version instead of the PHPv7 version when they transitioned to 2.4.4.

So either the hard link was not deleted/recreated, or some other error,

I did notice in some testing that trying to delete the php_pfb hard link from the shell under root, required a "y" confirmation. So I thought a force delete might be needed.

@jim-p

rbgarga: I ran this small test and it worked as expected:
```
touch("/root/file");
link("/root/file", "/root/link");
unlink("/root/link");
unlink("/root/file");
```

BBcan177: @rbgarga @jim-p 

Can we just accept this for now so that users don't continue to have the pfb_filter service show as down?

I can remove the force cmd if that helps, atleast we can fix one issue?

I would really like to get this approved soon. Thanks.

BBcan177: @rbgarga @jim-p

I have made the following changes:

1) Removed the rm -f exec()
2) Refactored the installation code for services
    * Create services
    * Stop services
    * Delete/create hard links
    * Start services
3) Add a safety belt to the pfb_filter service to recreate the hard links if the PHP version is mismatched.

jim-p: Rather than check the PHP version, you should check to see if your linked copy is actually a link to the current binary. There are ways other than the version they could be out of sync. Different compile time options, for one.

Here is a quick example of checking the link using `stat()` to read the inode number. If a file is a hard link to another file, their inode numbers will be identical. So if they differ, remove and recreate the link.

    touch("a");
    touch("b");
    link("a", "c");
    $a = stat("a");
    $b = stat("b");
    $c = stat("c");
    var_dump($a['ino']);
    var_dump($b['ino']);
    var_dump($c['ino']);  

When that is run you'll see that the `ino` of a and c are identical, but b is different.

BBcan177: Thanks Jim...  

I have the following:
https://github.com/pfsense/FreeBSD-ports/pull/575/commits/f7af7bb85f0a12bc2430415ce5f6a895fa422ed8

But can do the **stat** comparison instead:

    stat /usr/local/bin/php
    99 1846717 -rwxr-xr-x 3 root wheel 3713864 4856668 "Sep 29 17:00:18 2018" "Sep 14 07:32:41 
    2018" "Sep 29 16:15:58 2018" "May 21 13:00:08 2018" 32768 9600 0 /usr/local/bin/php

    stat /usr/local/bin/php_pfb
    99 1846717 -rwxr-xr-x 3 root wheel 3713864 4856668 "Sep 29 17:00:18 2018" "Sep 14 07:32:41 
    2018" "Sep 29 16:15:58 2018" "May 21 13:00:08 2018" 32768 9600 0 /usr/local/bin/php_pfb

Then if they don't match, I can recreate the hard link.

BBcan177: Added https://github.com/pfsense/FreeBSD-ports/pull/575/commits/d67c38f717216252cd5142e2309e29bf6d8c8eb6

Tried to use the stat -f with a custom format but had some shell syntax issues. I ended up just removing the filename from the stat output with a cut cmd.

PR #578, opened 2018-10-02
revert "FreeRadius package : Send Traffic quota to the Captive Portal (Redmine 8927)"

Augustin-FL:
Revert the previous pull request ( #576 ) as it seems to cause an issue (see https://redmine.pfsense.org/issues/8989 )

Renato or Jim, do you know any reason why [the FreeRADIUS pfSense dictionary](https://github.com/pfsense/pfsense/blob/master/src/usr/share/doc/radius/dictionary.pfsense) coud be missing or invalid ?


Augustin-FL:
Closing because of #579

PR #582, opened 2018-10-15
[pfSense-pkg-frr] Fix bgp weight

mhalden:
There is currently an error with how weight is set in route maps which not only causes it to not work, but breaks metric. This PR fixes this issue.


mhalden:
Portrevision bumped.

PR #586, opened 2018-10-29
pfBlockerNG-devel - Fix Interface selection list

BBcan177:


NanoCaiordo:
Just tested and the interfaces are correctly selected using different combination of settings.

PR #587, opened 2018-11-05
pfSense-pkg-snort-3.2.9.8 -- Snort GUI package update

bmeeks8:
### pfSense-pkg-snort-3.2.9.8
This update of the Snort GUI package provides support for version 2.9.12 of the Snort binary and also corrects a bug on the CATEGORIES tab that prevented the saving of Shared Object rules selections.

**New Features**
1.  Support for the latest 2.9.12 version of the Snort binary

**Bug Fixes**
1.  Share Object rule selections on the CATEGORIES tab cannot be saved.

### DIFF
--- a/security/pfSense-pkg-snort/Makefile
+++ b/security/pfSense-pkg-snort/Makefile
@@ -1,8 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-snort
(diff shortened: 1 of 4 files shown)


bmeeks8:
This GUI package version can work with either versions 2.9.11.1 or 2.9.12 of the Snort binary.  The correct rules tarball to download is determined programatically.

PR #591, opened 2018-11-10
Add custom command line options to ntopng

pixelsquared:
Add field to allow users to specify custom CLI options for ntopng


dennypage:
In the valuating this, it would be helpful if you gave examples of what option(s) you want to set that you currently cannot. Also, why you want to set them via arbitrary text, which my conflict with options that are already in the GUI, rather than extending the GUI to provide the necessary support.

pixelsquared:
The main option I wished to use is [--dump-flows|-F] to dump expired flows to external destinations.

The reason that I went with allowing any custom options is:

- I was unsure of the best way to display the dump expired flows option in the GUI due to their variability
- I could see users wanting to use other options and thought that it would be simpler to allow them to enter custom options until they are implemented in the gui

I do understand if you are concerned with users adding conflicting options. If you think that only adding a field for the dump expired flows option is better I could do that instead or you have a suggestion on how to better display the dump expired flows option in the GUI I could do that as well.

dennypage:
@jim-p may have a different opinion, but I would think it better to avoid an arbitrary text command line option due to potential conflicts with options that are already part of the GUI, fixed as part of the pfSense GUI package, or fixed as part of the underlying FreeBSD package.

jim-p:
I agree. I'd rather not encourage the practice. Also, as enacted, it's a giant security hole allowing a stored shell execution of arbitrary commands. It won't be merged.

Make options for each command line switch you want, and a way to validate/control their parameters, and we can look that over instead.

PR #593, opened 2018-11-11
pfblockerng autorule patch

NanoCaiordo:
By following the code it seems the autorule select box should have 3 options instead of 4.

In fact by selecting 'ar' no suffix will be added (array key 0 evaluates to false) and selecting 'AR' will suffix the number 1 (array key 1) like pfB_Text1.


NanoCaiordo:
I'm closing it.

PR #596, opened 2018-11-20
Switch menu linked tab

stephenw10:
Open the settings tab from the menu to prevent GUI issues that the log view tab can cause.


stephenw10:
https://forum.netgate.com/topic/137768/syslog-webgui-log-view-option

PR #602, opened 2018-12-12
Bug #9194 - pfSense-Status_Monitoring - Can't Save default view

joshuasign:
corrections and code cosmetics
$view_title will now be always set, so many condition can be removed so code reading will be easier.


joshuasign:
For informations, this code has just been tested, and it seems working without errors.

joshuasign:
ok, sorry for that i never use github before.
I look for the way to remove the mail report and put it in new PR... i do it as soon i find ^^

PR #604, opened 2018-12-16
Bug #9194 - Can't Save default view

joshuasign:
corrections and code cosmetics


jim-p:
Redmine issue is https://redmine.pfsense.org/issues/9194

PR #605, opened 2018-12-23
Tidy up SIP phone config download buttons

stephenw10:
Re-align buttons from this:
selection_535
To this:
selection_536


jim-p:
Redmine issue: https://redmine.pfsense.org/issues/9250

PR #607, opened 2018-12-28
Added Buypass CA ACME, and fix on registering email

mysteq:
Added support for Buypass Go ACME
https://www.buypass.com/ssl/products/acme
which also recently has been added as working in acme.sh.

Also this CA requires the email address, so fixed so it's written to accountconf.conf, which it seemed not to do in the existing code.

### DIFF
--- a/security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc
+++ b/security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc
@@ -1,7 +1,7 @@
 <?php
 /*
  * acme.inc
- * 
+ *
  * part of pfSense (https://www.pfsense.org/)
  * Copyright (c) 2016 PiBa-NL
  * All rights reserved.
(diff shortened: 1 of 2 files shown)


mysteq:
Sorry, editor that automatically removed trailing whitespace in the existing code.

And yes, I'm currently running this code on my pfsense with an Buypass certificate and with an updated acme.sh client.

jim-p:
The "and with an updated acme.sh client" part is a problem. What does that mean, exactly? Is that from acme.sh master or dev upstream or with some custom changes? I just updated from master a week or so ago, does it need something newer than that? If so, that would need to happen before this change could be accepted.

mysteq:
From acme.sh master, but yes it needs to be newer as it had changes that was commited to master recently. Should I add it to this PR, or add another PR?

jim-p:
I'll have to merge that myself, I have the local set of changes here, too easy to cause a conflict otherwise. This will need to wait. I should be able to get that in sometime next week.

PR #617, opened 2019-01-14
FRR: Show IPv4 and IPv6 BGP Summary

Napsterbater:
Change "show ip bgp summary $*" to "show bgp summary $*" so Summary page will show both IPv4 and IPv6 Info.

https://redmine.pfsense.org/issues/9244

### DIFF
--- a/net/pfSense-pkg-frr/files/usr/local/bin/frrctl
+++ b/net/pfSense-pkg-frr/files/usr/local/bin/frrctl
@@ -189,7 +189,7 @@ bgp*)
 		;;
 	sum*)
 		shift; shift;
-		daemon_command ${BGP_PORT} ${BGP_PASSWORD} "show ip bgp summary $*"
+		daemon_command ${BGP_PORT} ${BGP_PASSWORD} "show bgp summary $*"
 		;;
 	nexth*)
 		shift; shift;


jim-p:
Since other commands like routes split IPv4/IPv6 it might be better to do the same here so the output is separated better.

Also, as a part of this pull request, you must increase PORTVERSION or add a PORTREVISION line in the package Makefile. Without this version increase, the package will not be rebuilt to include the new change.

See https://www.netgate.com/docs/pfsense/development/developer-style-guide.html#ports-packages-specific-rules

Napsterbater:
Actually, only the routes are split, everything else is combined, neighbors, next hop, etc, Don't see why this should be different, IMO.

Also splitting would be way above my abilities. 

Ill have to look at PORTVERSION part tomorrow or so.

jim-p:
That's fine, then. It's not difficult to setup the new command, though. Just a couple lines of code here and there. But it's not a huge deal to have them merged. If it is, we can split it up later.

The version bump has to be done, though.

Napsterbater:
Changed PORTREVISION from 4 to 5, is that correct?

jim-p:
Yes, that's fine, thanks!

PR #622, opened 2019-01-22
pfSense-pkg-Telegraf - Add support for Ping and Netstat modules

Berzerker:
I use the ping and netstat modules for my grafana dashboard, and figured others may as well for statistical purposes.

I couldn't really find a better way, seeing as the config page is plain XML, to enter in multiple hosts, but I made the best of what I could with a list of 4 hosts, which adjusts accordingly if some are empty or not.

Feel free to comment if you think there's a better way for adding an arbitrary number of hosts.


Dazog:
Will this be merged for upcoming 2.5 beta's?

Berzerker:
Got it. Pushed a new commit with the PORTVERSION upped. 👍

PR #623, opened 2019-01-23
Fix line ending problem in security/pfSense-pkg-tinc

buergi:
Not stripping \r characters from tinc-up causes problems with custom
tinc-up scripts, e.g. the interface is added to the group "pkg_tinc\r"
instead of "pkg_tinc".

### DIFF
--- a/security/pfSense-pkg-tinc/files/usr/local/pkg/tinc.inc
+++ b/security/pfSense-pkg-tinc/files/usr/local/pkg/tinc.inc
@@ -126,7 +126,7 @@ function tinc_save() {
 	file_put_contents("{$configpath}/rsa_key.priv", base64_decode($tincconf['cert_key']) . "\n");
 	chmod("{$configpath}/rsa_key.priv", 0600);
 	if ($tincconf['tinc_up']) {
-		$_output = base64_decode($tincconf['tinc_up']) . "\n";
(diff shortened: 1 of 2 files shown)


security/pfSense-pkg-tinc/files/usr/local/pkg/tinc.inc
```diff
@@ -126,7 +126,7 @@ function tinc_save() {
 	file_put_contents("{$configpath}/rsa_key.priv", base64_decode($tincconf['cert_key']) . "\n");
 	chmod("{$configpath}/rsa_key.priv", 0600);
 	if ($tincconf['tinc_up']) {
-		$_output = base64_decode($tincconf['tinc_up']) . "\n";
+		$_output = str_replace("\r", "", base64_decode($tincconf['tinc_up'])) . "\n";
```
jim-p: Might be better to use `unixnewlines()` here instead manually picking what to remove.

buergi: The code in this file is full of `str_replace("\r","",...)`, I tried to stay consistent with the existing code. However, I agree, this is something that could be improved, however I don't know the the function you mentioned.

rbgarga: @buergi `unixnewlines()` can be found at https://github.com/pfsense/pfsense/blob/master/src/etc/inc/util.inc#L2870

PR #627, opened 2019-03-13
Enable Telegraf on ARM

robbiet480:
Telegraf is now building on ARM! This PR incorporates upstream changes at https://github.com/freebsd/freebsd-ports/commit/8a143a3599a11c8ddd10f9e82b587bd696ac4d32 to enable building telegraf on ARM as well as enabling the package to be installed on ARM.

More info: https://github.com/influxdata/telegraf/issues/5282


victorhooi:
This would be awesome!

I have several SG-3100's, and would be amazing to have this available there as well.

Are there any blockers to having this go through?

rbgarga:
`net-mgmt/telegraf` is on version 1.10.1 on pfSense 2.5.0 snapshots. I've enabled the build of `pfSense-pkg-Telegraf` for armv7 on this version as well so we can evaluate if it works as expected before think about backport it to 2.4.4. The patch in this PR is outdated so I'm closing it.

Thanks!

rbgarga:
FYI, lang/go is not building on poudriere armv7 jail. I'll re-enable Telegraf on arm when it's fixed upstream.

robbiet480:
Thanks for the update @rbgarga

victorhooi:
@rbgarga Bugger about lang/go not building - do you have a link to the upstream bug?

victorhooi:
I think this is working now?

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=221297

Was marked as fixed now. Does this mean we can pull in Telegraf now? =)

rbgarga:
> I think this is working now?
> 
> https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=221297
> 
> Was marked as fixed now. Does this mean we can pull in Telegraf now? =)

Not that simple. lang/go cannot build using qemu-user-static, which is the method we currently use. We need native builders and restructure our build process to be able to use it.

artooro:
@rbgarga would you guys consider a custom port of telegraf for pfSense that pulls in pre-compiled binaries so that telegraf can be compiled outside of poudriere and then pulled into a freebsd package?

I do this currently for an internal freebsd package we use, it's not ideal but it works, and the only workaround I'm currently aware of.

Essentially it would be a two step build. First run the native go cross compile to create binaries, and then run the poudriere jobs.

onedr0p:
If I have it correct we're just waiting on pfSense to update to 2.5.x to have telegraf supported on ARM?

victorhooi:
Oh, this is working on 2.5.x now? That is great news!

Where did you find this information?

onedr0p:
See this comment https://github.com/pfsense/FreeBSD-ports/pull/627#issuecomment-482140225

victorhooi:
Hmm - but I thought in the comment directly below that, he’s saying it’s not working again?

https://github.com/pfsense/FreeBSD-ports/pull/627#issuecomment-482185573

> FYI, lang/go is not building on poudriere armv7 jail. I'll re-enable Telegraf on arm when it's fixed upstream.

PR #630, opened 2019-03-19
Make -d and -v actually work, modify pfSense UI

zmanion:
Code was added to arpwatch to ignore CARP/VRRP MAC addresses when logging but the runtime option (-v) was not enabled. These changes also allow arpwatch to be run with -d for debug. Modify pfSense UI to support -v.

Here's the bug for upstream: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=235415. Not sure how pfSense feels about getting out of sync with FreeBSD.

I've tested this on one pfSense system running 2.4.4-RELEASE-p2 and the UI works:

```
$ pgrep -lf arpwatch
1762 /usr/local/sbin/arpwatch -v -z -f /usr/local/arpwatch/arp_igb3.dat -i igb3 -m foo@example.com
1533 /usr/local/sbin/arpwatch -v -z -f /usr/local/arpwatch/arp_igb2.dat -i igb2 -m foo@example.com
1155 /usr/local/sbin/arpwatch -v -z -f /usr/local/arpwatch/arp_igb0.dat -i igb0 -m foo@example.com
```

### DIFF
--- a/net-mgmt/arpwatch/Makefile
+++ b/net-mgmt/arpwatch/Makefile
@@ -3,7 +3,7 @@
 
 PORTNAME=	arpwatch
 DISTVERSION=	2.1a15
-PORTREVISION=	10
+PORTREVISION=	11
 CATEGORIES=	net-mgmt
 MASTER_SITES=	ftp://ftp.ee.lbl.gov/ \
 		GENTOO
--- a/net-mgmt/arpwatch/Makefile
+++ b/net-mgmt/arpwatch/Makefile
@@ -3,7 +3,7 @@
 
 PORTNAME=	arpwatch
 DISTVERSION=	2.1a15
-PORTREVISION=	11
+PORTREVISION=	10
 CATEGORIES=	net-mgmt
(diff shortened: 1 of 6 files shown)


zmanion:
Changes to the port would be necessary, see https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=235415. The port needs to actually accept the -v flag.
I understand not wanting to be out of sync with FreeBSD.

jim-p:
Looks like the changes to the port were just accepted, which is a good sign.

rbgarga:
@zmanion I've cherry picked `net-mgmt/arpwatch` changes from upstream. Could you please remove any change to this port from this Pull Request and keep only changes to `net-mgmt/pfSense-pkg-arpwatch` ?

zmanion:
@rbgarga: Done, I left the change to bump PORTVERSION in the Makefile, not sure if that is correct or not.

rbgarga:
I've manually merged it. Thanks!

PR #631, opened 2019-03-22
pfSense-pkg-suricata -- Update to version 4.1.3

bmeeks8:
### pfSense-pkg-suricata-4.1.3
This update for the Suricata package introduces one modified new feature, fixes three bugs and brings the stats logging parameter settings in _suricata.yaml_ inline with the newest Suricata release.

**New Features**
1. Provide separate enable checkboxes for controlling the logging of flow and netflow data.  Formerly a single setting (Traffic Flows) was provided that either enabled or disabled both flow and netflow data logging simultaneously.  These two parameters can now be set independently of each other.  Reference Issue #9403 on the pfSense Redmine site.

**Modified Features**
1. Adjust stats logging config parameters in _suricata.yaml_ to bring them in ilne with the format expected by the Suricata 4.1.3 binary.

**Bug Fixes**
1. The sample conf files for SID MGMT provided in the package we not being read and installed properly on new green field installations of Suricata.  Thus there were no sample conf files present on the SID MGMT tab.

2. On the CATEGORIES tab, status messages for the Snort GPLv2 Community Rules were showing the variable name used to hold the status message instead of the content of the variable.

3.  Update Snort Subscriber Rules tarball filename in hint text to _snortrules-snapshot-29120.tar.gz_ to reflect the most recent Snort 2.9.x rules release.

### DIFF
--- a/security/pfSense-pkg-suricata/files/arm/usr/local/www/suricata/suricata_rulesets.php
+++ b/security/pfSense-pkg-suricata/files/arm/usr/local/www/suricata/suricata_rulesets.php
@@ -428,9 +428,9 @@ else:
 						</td>
 						<td colspan="4">
 						<?php if ($no_community_files): ?>
-							<?php echo gettext("{$msg_community}"); ?>
+							<?php echo gettext($msg_community); ?>
 						<?php else: ?>
-							<a href='suricata_rules.php?id=<?=$id;?>&openruleset=<?=$community_rules_file;?>'><?=gettext('{$msg_community}');?></a>
+							<a href='suricata_rules.php?id=<?=$id;?>&openruleset=<?=$community_rules_file;?>'><?=gettext($msg_community);?></a>
 						<?php endif; ?>
 						</td>
 					</tr>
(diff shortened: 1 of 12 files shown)


security/pfSense-pkg-suricata/files/arm/usr/local/www/suricata/suricata_rulesets.php
```diff
@@ -428,9 +428,9 @@
 						</td>
 						<td colspan="4">
 						<?php if ($no_community_files): ?>
-							<?php echo gettext("{$msg_community}"); ?>
+							<?php echo gettext($msg_community); ?>
```
rbgarga: `$msg_community` declaration already passed through `gettext()` at https://github.com/pfsense/FreeBSD-ports/blob/0d68a01c5e7180e9afe9b06265391aa84103fc58/security/pfSense-pkg-suricata/files/arm/usr/local/www/suricata/suricata_rulesets.php#L407 so it could be removed on echo commands

bmeeks8: Changes made.  Sorry I missed that earlier.

PR #632, opened 2019-03-23
trying to fix missing CRL not written in CA file

paolomeraviglia:
From the forum

https://forum.netgate.com/topic/70749/freeradius-2-with-eap-tls/13

I still see problems merging CA and CRL in a single file for freeradius.

This patch works on freeradius3 with pfsense stable 2.4.4_2 setup and apply the same method used to export CRL from certmanager to create a CA+CRL file


jim-p:

See https://www.netgate.com/docs/pfsense/development/developer-style-guide.html#ports-packages-specific-rules

paolomeraviglia:
Ok, portrevision increased: Sorry for my noobness

PR #633, opened 2019-04-11
Convert Windows line feeds in squid ca certificate

Multiconecta:
Avoid FATAL error No valid signing SSL certificate configured load squid with a CA certificate imported from Windows.

Issue described in this [Netgate Forum thread](https://forum.netgate.com/topic/142453/ssl-intercept-with-windows-certificate-no-valid-signing-ssl-certificate-configured-for-http_port).

### DIFF
--- a/www/pfSense-pkg-squid/files/usr/local/pkg/squid.inc
+++ b/www/pfSense-pkg-squid/files/usr/local/pkg/squid.inc
@@ -1175,7 +1175,7 @@ function squid_resync_general() {
 					// Should never get here
 					$sslproxy_dhparams = "";
 				}
-				file_put_contents($crt_pk, str_replace("\r\n", "\n", base64_decode($srv_cert['prv']) . base64_decode($srv_cert['crt'])));
+				file_put_contents($crt_pk, unixnewlines(base64_decode($srv_cert['prv']) . "\n" . base64_decode($srv_cert['crt'])));
 				$sslcrtd_children = ($settings['sslcrtd_children'] ? $settings['sslcrtd_children'] : 5);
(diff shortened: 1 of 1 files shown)


www/pfSense-pkg-squid/files/usr/local/pkg/squid.inc
```diff
@@ -1175,7 +1175,7 @@ function squid_resync_general() {
 					// Should never get here
 					$sslproxy_dhparams = "";
 				}
-				file_put_contents($crt_pk, base64_decode($srv_cert['prv']) . base64_decode($srv_cert['crt']));
+				file_put_contents($crt_pk, str_replace("\r\n", "\n", base64_decode($srv_cert['prv']) . base64_decode($srv_cert['crt'])));
```
jim-p: We have a function for this, `unixnewlines()`.

So you'd use `unixnewlines( base64_decode($srv_cert['prv']) . base64_decode($srv_cert['crt']) )` here instead of  `str_replace()`.

Multiconecta: Perfect, I'll change it. But I'm getting the same error restarting squid without changing the certificate also, so, please, let this PR on hold until I can make more tests.

Multiconecta:
Ok. I'm not sure if it is because of Windows line feeds, but after de key there were no line feed, so writing both key and certificate on the file got us in only one line:
```
...
-----END PRIVATE KEY----------BEGIN CERTIFICATE-----
...
```
Let me try the patch with a newline between key and certificate.

jim-p:
We've seen that before with stunnel: https://github.com/pfsense/FreeBSD-ports/commit/09b030108caa8eaa23d102b50b82aa57275a06c9

www/pfSense-pkg-squid/files/usr/local/pkg/squid.inc
Multiconecta: Done

PR #640, opened 2019-05-15
pfSense-pkg-suricata - Update to version 4.1.4

bmeeks8:
### pfSense-pkg-suricata-4.1.4
This update to the Suricata GUI package contains one bug fix and no new features.

**New Features**
None

**Bug Fixes**
1.  On the SID MGMT tab, when the blocking mode for an interface does not meet all the criteria, the user can be prevented from deleting a previously assigned DROP SID or REJECT_SID list even if the list is no longer applicable for the selected blocking mode.

### DIFF
--- a/security/pfSense-pkg-suricata/Makefile
+++ b/security/pfSense-pkg-suricata/Makefile
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-suricata
-PORTVERSION=	4.1.3
+PORTVERSION=	4.1.4
 CATEGORIES=	security
 MASTER_SITES=	# empty
 DISTFILES=	# empty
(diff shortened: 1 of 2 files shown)


shonjir:
Can this be pulled into RELENG_2.4.4?  The release package tree is still showing 4.1.2.

rbgarga:
@bmeeks8 what do you think?

bmeeks8:
So far as I know, it should be fine to pull the Suricata 4.1.1 package into pfSense-2.4.x.  You will need to pull in both the suricata-4.1.1 binary and the pfSense-pkg-suricata-4.1.1 GUI package.  There are no pfSense-2.5 dependent features in the GUI code, and I'm not aware of any in the binary.  You might want to check the version of rust to be sure, though.

rbgarga:
I've sync'd suricata on 2.4.4 branch from devel

PR #643, opened 2019-05-29
snort-2.9.13_1 -- Add support for Inline IPS Mode and bug fix

bmeeks8:
### Snort-2.9.13_1
This update to the Snort binary adds support for the new Inline IPS Mode coming to the pfSense Snort GUI package.  Two existing bugs are also fixed.

**New Features:**
1.  Added two new optional field names that can be specified for the CSV log output plugin.  They are _action_ and _disposition_.  _Action_ will be one of "none, alert, drop, log, pass, reject or sdrop".  _Disposition_ will be one of "Allow, CDrop, WDrop, Drop or FDrop".  These additional optional fields, when configured for output by the GUI package, output the IPS inline status for the current packet event.

2.  Changed the BUILD_DEPENDS= line to include _daq-2.2.2_1_ or higher as a build requirement.  The _daq-2.2.2_1_ and higher versions contain the netmap inline IPS mode updates that allow use of the host OS stack as a netmap endpoint.

**Bug Fixes:**
1.  Added _bison_ to the USES= line in the Makefile to force use of _bison_ during the build process.  This fixes an error that creeps in and renders Snort unable to read in an XML Host Attribute Table file when _yacc_ happens to be the default on the build system.

2.  Edited the arguments parsing code in the initialization function for the custom alert_pf output blocking plugin used on pfSense. The old code could potentially segfault if an optional argument was NULL.


security/snort/Makefile
```diff
@@ -44,7 +45,7 @@ OPTIONS_GROUP=		ADDONS DEV
 OPTIONS_GROUP_ADDONS=	BARNYARD PULLEDPORK
 OPTIONS_GROUP_DEV=	DBGSNORT
 OPTIONS_SUB=		yes
-OPTIONS_DEFAULT=	GRE NORMALIZER BARNYARD \
+OPTIONS_DEFAULT=	GRE NORMALIZER TARGETBASED BARNYARD \
```
rbgarga: Instead of changing OPTIONS_DEFAULT, creating an extra diff from upstream, it's better just to set this option on make.conf used by poudriere to build packages at https://github.com/pfsense/pfsense/blob/master/tools/conf/pfPorts/make.conf#L147

security/snort/Makefile
```diff
@@ -15,13 +16,13 @@ COMMENT=	Lightweight network intrusion detection system
...
-BUILD_DEPENDS=	daq>=2.2.2:net/daq
+BUILD_DEPENDS=	daq>=2.2.2_1:net/daq
 LIB_DEPENDS=	libpcre.so:devel/pcre \
 		libdnet.so:net/libdnet \
 		libpcap.so:net/libpcap
-RUN_DEPENDS=	daq>=2.2.2:net/daq
+RUN_DEPENDS=	daq>=2.2.2_1:net/daq
 
-USES=			cpe libtool pathfix shebangfix ssl
+USES=			cpe libtool pathfix shebangfix ssl bison
```
rbgarga: Please keep the list sorted alphabetically

PR #653, opened 2019-06-16
Add pfSense-pkg-node_exporter

carlpett:
This PR adds a pfSense package for [node_exporter](https://github.com/prometheus/node_exporter). It also bumps the existing node_exporter port version.
Tested on 2.4.4-r3.

Note to the reviewer: I've never really touched FreeBSD itself outside of pfSense before, as well as not having any previous experience with the pfSense package, so probably makes sense to check for silly mistakes or misunderstandings.


carlpett:
@jim-p Sorry for direct ping, but it is not clear to me if I missed some step here to trigger someone to take a look, or if it just summertime and things are moving slow?

carlpett:
@jim-p Ping?

jim-p:
You didn't miss anything, as soon as we have some spare cycles to review it, we'll review it. Thanks!

sysutils/node_exporter/Makefile
```diff
@@ -2,7 +2,7 @@
 # $FreeBSD$
 
 PORTNAME=	node_exporter
-PORTVERSION=	0.17.0
+PORTVERSION=	0.18.1
```
rbgarga: You shouldn't touch `sysutils/node_exporter` port here. This is a port provided by FreeBSD, if you need to upgrade it, it must be done on FreeBSD ports tree.

sysutils/pfSense-pkg-node_exporter/Makefile
```diff
@@ -0,0 +1,39 @@
+# $FreeBSD$
+
+PORTNAME=	pfSense-pkg-node_exporter
+PORTVERSION=	0.18.1
+PORTEPOCH=	1
```
rbgarga: A new port cannot have PORTEPOCH set, please remove it

sysutils/pfSense-pkg-node_exporter/Makefile
```diff
@@ -0,0 +1,39 @@
...
+MASTER_SITES=	# empty
+DISTFILES=	# empty
+EXTRACT_ONLY=	# empty
+
+MAINTAINER=	coreteam@pfsense.org
+COMMENT=	pfSense package node_exporter
+
+LICENSE=	APACHE20
+
+RUN_DEPENDS=	${LOCALBASE}/bin/node_exporter:sysutils/node_exporter
```
rbgarga: This dependency format it deprecated, instead of pointing to a file just use the package. It will be like this:
```RUN_DEPENDS=	node_exporter>0:sysutils/node_exporter```

File: sysutils/pfSense-pkg-node_exporter/files/usr/local/etc/rc.conf.d/node_exporter [hunk 4]
```diff
@@ -0,0 +1 @@
+node_exporter_enable="NO"
```
rbgarga: This file will not be used by pfSense and can be removed

carlpett:
I've pushed fixes for the comments, thanks for taking a look @rbgarga!

carlpett:
@rbgarga @jim-p Any spare cycles coming up?

sysutils/pfSense-pkg-node_exporter/files/usr/local/pkg/node_exporter.inc
```diff
@@ -0,0 +1,133 @@
...
+
+function node_exporter_write_config($node_exporter_config) {
+	$txt = "# This file is generated by the pfSense node_exporter package.\n";
+	$txt .= "# Do not edit this file, it will be overwritten automatically.\n";
+	$txt .= "\n";
+	$txt .= "node_exporter_enable=\"YES\"\n"; // If it had been NO, this function isn't called, so fine to hardcode
+
+	$listen_addr = get_interface_ip($node_exporter_config["listen_iface"]);
+	$listen_port = $node_exporter_config["listen_port"] ?: "9100";
+	$txt .= "node_exporter_listen_address=\"" . $listen_addr . ":" . $listen_port . "\"\n";
```
jim-p: Use inline variable substitution here, for example:

```
$txt .= "node_exporter_listen_address=\"{$listen_addr}:{$listen_port}\"\n";
```

sysutils/pfSense-pkg-node_exporter/files/usr/local/pkg/node_exporter.inc
```diff
@@ -0,0 +1,133 @@
...
+
+	$collectorTxt = "";
+	foreach (explode(",", $node_exporter_config["collectors"]) as $collector) {
+		$collectorTxt .= " --collector.".$collector;
+	}
+
+	$txt .= "node_exporter_args=\"". $collectorTxt ." " . $node_exporter_config["extra_flags"] . "\"\n";
```
jim-p: Similar to the previous note, simplify this a little:
```
$txt .= "node_exporter_args=\"{$collectorTxt} {$node_exporter_config['extra_flags']}\"\n";
```
(Note the change to single quotes around `extra_flags`)

sysutils/pfSense-pkg-node_exporter/files/usr/local/pkg/node_exporter.inc
```diff
@@ -0,0 +1,133 @@
...
+		stop_service('node_exporter');
+	}
+
+	$node_exporter_config = &$config['installedpackages']['nodeexporter']['config'][0];
+	if (isset($node_exporter_config['enable'])) {
+		$enable = $node_exporter_config['enable'];
+	} else {
+		$enable = 'no';
+	}
+	if ($enable !== 'on') {
```
jim-p: There is no need to assign `$enable` here, you can test `$node_exporter_config['enable']` directly. For example:

```
if ($node_exporter_config['enable'] !== 'on') {
```

If it's empty/unset that will still take the same path.

sysutils/pfSense-pkg-node_exporter/files/usr/local/pkg/node_exporter.inc
```diff
@@ -0,0 +1,133 @@
...
+	unlink_if_exists(NODE_EXPORTER_RCFILE);
+	unlink_if_exists(NODE_EXPORTER_CONFIG_FILE);
+
+	// Set up textfile cron job
+	install_cron_job(NODE_EXPORTER_TEXTFILE_CRONTASK, false);
+}
+
+function node_exporter_validate_form($post, &$input_errors) {
+	if (empty($post['listen_iface'])) {
+		$input_errors[] = 'You must select an interface to listen to';
```
jim-p: Wrap this text in `gettext()`

sysutils/pfSense-pkg-node_exporter/files/usr/local/pkg/node_exporter.inc
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	if (!is_port($post['listen_port'])) {
+		$input_errors[] = 'You must enter a valid port number in the \'Listen Port\' field';
```
jim-p: Wrap this text in `gettext()`

sysutils/pfSense-pkg-node_exporter/files/usr/local/share/pfSense-pkg-node_exporter/interface-collector.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python
```
jim-p: Just a note that you'll need to try this python script on 2.5.0 snapshots (python 3.6) and 2.4.x (python 2.7) to ensure it works in both environments.

carlpett: Validated that the script works under both python versions. Updated with two suggestions from `2to3` (although they don't seem to be strictly needed, it worked as-was).
I don't have a spare pfsense environment to test pf2.5 on, however. Are there any changes to the config.xml? The script does some parsing of the `interfaces` tree.

carlpett:
Thanks @jim-p! Updated as per your suggestions

sysutils/pfSense-pkg-node_exporter/files/usr/local/share/pfSense-pkg-node_exporter/interface-collector.py
jim-p: Nothing changed around interfaces, so it should be OK

sysutils/pfSense-pkg-node_exporter/Makefile
```diff
@@ -0,0 +1,38 @@
...
+
+SUB_FILES=	pkg-install pkg-deinstall
+SUB_LIST=	PORTNAME=${PORTNAME}
+
+do-extract:
+	${MKDIR} ${WRKSRC}
+
+do-install:
+	${MKDIR} ${STAGEDIR}${PREFIX}/pkg
+	${MKDIR} ${STAGEDIR}${PREFIX}/etc/rc.conf.d
```
rbgarga: No files are going to be installed here, no need to create this directory.

sysutils/pfSense-pkg-node_exporter/Makefile
```diff
@@ -0,0 +1,38 @@
...
+	${MKDIR} ${STAGEDIR}${DATADIR}
+	${INSTALL_DATA} ${FILESDIR}${PREFIX}/pkg/node_exporter.xml ${STAGEDIR}${PREFIX}/pkg
+	${INSTALL_DATA} ${FILESDIR}${PREFIX}/pkg/node_exporter.inc ${STAGEDIR}${PREFIX}/pkg
+	${INSTALL_DATA} ${FILESDIR}${PREFIX}/etc/rc.conf.d/node_exporter ${STAGEDIR}${PREFIX}/etc/rc.conf.d/
```
rbgarga: This file doesn't exist, do not try to install it

sysutils/pfSense-pkg-node_exporter/files/usr/local/share/pfSense-pkg-node_exporter/node_exporter.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/sh
```
rbgarga: This file seems to be a copy of sysutils/node_exporter init script, we don't need to carry it here

sysutils/pfSense-pkg-node_exporter/Makefile
```diff
@@ -0,0 +1,38 @@
...
+	${INSTALL_DATA} ${FILESDIR}${DATADIR}/info.xml ${STAGEDIR}${DATADIR}
+	${INSTALL_SCRIPT} ${FILESDIR}${DATADIR}/node_exporter.sh ${STAGEDIR}${DATADIR}
```
rbgarga: If this file is really a copy of sysutils/node_exporter init script, we don't need to carry it here and it should be removed

sysutils/pfSense-pkg-node_exporter/Makefile
```diff
@@ -0,0 +1,38 @@
...
+	${INSTALL_SCRIPT} ${FILESDIR}${DATADIR}/interface-collector.py ${STAGEDIR}${DATADIR}
```
rbgarga: Since this package is installing a python script, we need to have `USES= python` to make sure python is installed as a dependency.  If it requires any extra python module available on ports, it also should be added to RUN_DEPENDS

carlpett:
Thanks @rbgarga! Pushed fixes now

rbgarga:
I've manually merged it to incorporate some ports specific changes.

Also opened a redmine ticket to track it - https://redmine.pfsense.org/issues/9974. Thanks!

PR #658, opened 2019-08-11
Patch Barnyard2 for SG-3100 ARMv6/7

rickyzhang82:
The Barnyard2 from 2.4.4.p3 is broken in my recent purchased SG-3100. There are two road blocks:

- Outdated syntax in SQL query.
- Use optimization `-O` CFLAGS that cause unaligned memory access in pointer casting. Barnyard2 crashed with bus error.

I have tested my patched Barnyard2 under SG-3100 in 2.4.4.p3 for days. The database is MariaDB 10.3.9. I haven't seen the crash any more.

See discussion details in [pfSense forum](https://forum.netgate.com/topic/143538/barnyard2-and-mariadb/6)

### DIFF
--- a/security/barnyard2/files/patch-configure.in
+++ b/security/barnyard2/files/patch-configure.in
@@ -1,0 +1,28 @@
+--- configure.in.orig	2013-05-27 04:04:34 UTC
++++ configure.in
+@@ -428,6 +428,16 @@ AC_DEFUN([FAIL_MESSAGE],[
+    exit 1
+ ])
+ 
++# any arm platform has to disable optimization
++AC_MSG_CHECKING(for arm avoid bus error)
++if eval "echo $host_cpu|grep -i 'armv[[6|7]]' >/dev/null"; then
++    AC_MSG_RESULT(yes)
++    NO_OPTIMIZE="yes"
++    CFLAGS="$CFLAGS -mno-unaligned-access "
++else
++    AC_MSG_RESULT(no)
++fi
++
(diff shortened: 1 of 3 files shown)


rickyzhang82:
See the core dump below. Using the optimization flag will let clang generate the optimized version multiple memory load and store opcode like `ldm` and `stm`, which doesn't support unaligned memory access in ARMv6 and ARMv7.

See [ARM technical doc here](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.faqs/ka15414.html).

```
[Ricky@t20 2019Aug11-O2-g-stripped]$ lldb -c barnyard2-26451-0.core
(lldb) target create --core "barnyard2-26451-0.core"
Core file '/home/Ricky/Documents/barnyard2/core/2019Aug11-O2-g-stripped/barnyard2-26451-0.core' (arm) was loaded.
(lldb) bt
* thread #1, name = 'barnyard2', stop reason = signal SIGBUS
  * frame #0: 0x0004005
```

```
[Ricky@t20 2019Aug11-O2-g-stripped]$ r2 barnyard2
 -- Use zoom.byte=entropy and press 'z' in visual mode to zoom out to see the entropy of the whole file
[0x0000b320]> pd@0x40058
            0x00040058      070099e8       ldm sb, {r0, r1, r2}
            0x0004005c      9c00a4e5       str r0, [r4, 0x9c]!
            0x00040060      0c0084e2       add r0, r4, 0xc

```

PR #671, opened 2019-09-10
update shodan search engine

fgrosswig:
shodan search query based on IP changed a while so keep this on track.


rbgarga:
@BBcan177 could you take a loot at this?

BBcan177:
The existing link is ok for IP, but this link can be added to the DNSBL Lookup threat sources. I will add this to the next version.

fgrosswig:
Well the original pointed to **https://www.shodan.io/host/** which gave a 404 on shodan.

rbgarga:
@BBcan177 should I close this ticket since you are adding it to next version?

rbgarga:
@BBcan177 ping

BBcan177:
This can be approved. But no need to bump the version until there are some other changes to make.

rbgarga:
I'll merge it on 2.5.0 and 2.4.5. When (and if we) touch 2.4.4 package we can include it together

PR #672, opened 2019-09-14
Change redir domain(s) to use regex

StarkJohan:
This makes the redirects much more flexible in general.
Also makes it possible to redirect a subdomain to a different mapping than the domain itself.
This is a more consistent approach as mappings and redirects are handled in a similar manner.

Redmine: https://redmine.pfsense.org/issues/9762
Ready for review

### DIFF
--- a/www/pfSense-pkg-squid/files/usr/local/pkg/squid_reverse.inc
+++ b/www/pfSense-pkg-squid/files/usr/local/pkg/squid_reverse.inc
@@ -244,7 +244,7 @@ function squid_resync_reverse() {
 
 				if (is_array($rdr['row'])) {
 					foreach ($rdr['row'] as $uri) {
-						$conf_rdr .= "acl rdr_dst_{$rdr['name']} dstdomain {$uri['uri']}\n";
+ 						$conf_rdr .= "acl rdr_dst_{$rdr['name']} dstdom_regex -i {$uri['uri']}\n";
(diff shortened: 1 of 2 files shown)


rbgarga:
It changes the behavior of a field already present on users config files and it would lead users to have a config they didn't chose. It would be better if you add a new field for that and let user use both options

rbgarga:
No answers from @StarkJohan since Sep 25 review. Closing it due to inactivity. If you are still interested to work on it, please re-open.

StarkJohan:
> No answers from @StarkJohan since Sep 25 review. Closing it due to inactivity. If you are still interested to work on it, please re-open.

I can't reopen myself as far as I know as it was closed by you but I intend to follow up on this.

My understanding of dstdomain vs dstdom_regex is not perfect but as far as I can tell dstdom_regex should be fully "backwards compatible" with dstdomain (I think). Consider "dstdomain www.example.com" from an existing config. This entry is a valid regex for dstdom_regex and the behaviour is exactly the same. I can't say I'm 100% sure but I cannot think of a scenario where my suggested change will break any existing configs.

The only scenario that I can think of that possibly could be a problem is two requests like this but I don't think dstdom_regex matches on the path. I am yet to test this but I'm fairly confident this will not be a problem.
http://example1.com
http://example2.com/path/to/example1.com

PR #673, opened 2019-09-19
Corrected Typo

StefanHeimberg:
Small typo. I noticed earlier in the web UI


StefanHeimberg:
i know, i did not made a redmine ticket for that. sorry. but i do not have a account to create such one. and only for this typo i do not want to create one. i hope this ist not a problem.

PR #675, opened 2019-09-23
Redmine #9752: Write certificate files before executing actions.

s-fiebig:
Execute actions only after the additional certificate files have been written.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/9752
- [x] Ready for review


s-fiebig:
Sorry, I had set the review status to early on ready.

PR #676, opened 2019-09-25
Suricata-4.1.5 update and fix Pass List search issue.

bmeeks8:
### Suricata-4.1.5
This updates the Suricata binary to version 4.1.5 to match the lastest upstream release. Upstream now incorporates the GeoIP2 patch formerly used exclusively on pfSense, thus that patch file is now no longer necessary and is removed (patch-geoip2.diff).

**New Features:**
None

**Bug Fixes:**
1. An issue within the custom blocking module (_alert-pf_) used on pfSense could result in the Destination IP address in a packet being treated as part of a configured Pass List if the Source IP address of the same packet is in the configured Pass List. This issue was only present when BOTH was selected for the "_Which IP to Block_" parameter in the Suricata GUI package.


rbgarga:
Manually merged patch adjustments after cherry-pick original commit

PR #679, opened 2019-09-28
Suricata - Update to version 4.1.5 and fix Pass List search issue.

bmeeks8:
### Suricata v4.1.5
This updates the Suricata binary to version 4.1.5 to match the latest upstream release. It also includes a fix for a Pass List search issue. Upstream now incorporates the former GeoIP2 custom patch used on pfSense, so that patch file is now removed.

**New Features:**
None

**Bug Fixes:**
1. Corrects an issue when the "Which IP to Block" parameter for the custom blocking plugin is set to BOTH that can result in the code assuming the Destination IP address is also covered by a Pass List IP range whenever the Source IP address in the same packet is covered by a Pass List IP range.


rbgarga:
I just cherry picked necessary commits from devel. Thanks!

PR #682, opened 2019-10-06
#9811: allow to set BATTERYLEVEL and MINUTES to -1

dbaldes:
Redmine issue: https://redmine.pfsense.org/issues/9811

### DIFF
--- a/sysutils/pfSense-pkg-apcupsd/files/usr/local/pkg/apcupsd.inc
+++ b/sysutils/pfSense-pkg-apcupsd/files/usr/local/pkg/apcupsd.inc
@@ -42,6 +42,11 @@ function php_deinstall_apcupsd() {
 
 }
 
+function is_int_with_minus_one($arg) {
+  return (is_int($arg) || is_string($arg) && preg_match('/^-?\d+$/', trim($arg)))
+         && intval($arg) >= -1;
+}
+
 function validate_input_apcupsd($post, &$input_errors) {
 
 	if (isset($post['apcupsdenabled'])) {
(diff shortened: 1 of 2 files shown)


sysutils/pfSense-pkg-apcupsd/files/usr/local/pkg/apcupsd.inc
```diff
@@ -42,6 +42,11 @@ function php_deinstall_apcupsd() {
 
 }
 
+function is_int_with_minus_one($arg) {
+  return (is_int($arg) || is_string($arg) && preg_match('/^-?\d+$/', trim($arg)))
+         && intval($arg) >= -1;
```
rbgarga: This function can be simplified doing
```php
    return (is_int($arg) || $arg == '-1');
```

and since the result is really simple conditional I believe the function should be removed and the check done on both cases below.

dbaldes: Your suggestion would

- return true for -2 (because it is an int)
- return false for "2" (because it is a string)

All existing validation seems to take strings and numbers, so I designed this function in this way too.
Also, -2 is not a valid value.

dbaldes: Bumped PORTREVISION

rbgarga: > Your suggestion would
> 
> * return true for -2 (because it is an int)
> * return false for "2" (because it is a string and not '-1')
> 
> All existing validation seems to take strings and numbers, so I designed this function in this way too.
> Also, -2 is not a valid value.

You can use `is_numericint()` from util.inc instead, it will make sure it's a positive int

dbaldes: Done

PR #685, opened 2019-10-10
patching logging

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9742
 Ready for review

Adds ability to log patch id to system log

### DIFF
--- a/sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
+++ b/sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
@@ -236,11 +236,11 @@ function patchlog($msg) {
 function patch_descr($id) {
 	global $config;
 	$a_patches = &$config['installedpackages']['patches']['item'];
-	$descr = gettext(" (ID: {$a_patches[$id]['uniqid']}");
+	$descr = " (" . gettext("ID") . ": {$a_patches[$id]['uniqid']}");
 	if (isset($a_patches[$id]['descr'])) {
-		$descr .= gettext(", DESCR: {$a_patches[$id]['descr']}");
+		$descr .= ", " . gettext("DESCR") . ": {$a_patches[$id]['descr']}");
 	}
-	$descr .= gettext(")");
+	$descr .= ")";
 	return $descr;
 }
 
--- a/sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
+++ b/sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
@@ -184,7 +184,9 @@ function bootup_apply_patches() {
 		   only attempt to apply if it can be applied;
 		   and	if it can be reverted it is presumably already applied, so skip it. */
 		if (isset($patch['autoapply']) && patch_test_apply($patch) && !patch_test_revert($patch)) {
-			patch_apply($patch);
+			$descr = $patch['descr'];
+			$savemsg = patch_apply($patch) ? gettext("Patch autoapplied succesfully") : gettext("Patch could NOT be autoapplied");
+	                patchlog($savemsg, $descr);
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


vktg:
+ code cleanup
+ description field is mandatory now
+ logging of **Auto Apply**

sysutils/pfSense-pkg-System_Patches/files/usr/local/www/system_patches.php
```diff
@@ -48,28 +48,38 @@
...
-	$savemsg .= " (<a href=\"system_patches.php?id={$_POST['id']}&amp;fulltest=revert\" usepost>" . gettext("detail") . "</a>)";
-}
-if (($_POST['fulltest']) && ($a_patches[$_POST['id']])) {
-	if ($_POST['fulltest'] == "apply") {
-		$fulldetail = patch_test_apply($a_patches[$_POST['id']], true);
-	} elseif ($_POST['fulltest'] == "revert") {
-		$fulldetail = patch_test_revert($a_patches[$_POST['id']], true);
+if ($a_patches[$_POST['id']]) {
+	$savemsg = gettext("Patch ");
+	$descr = $a_patches[$_POST['id']]['descr'];
```
jim-p: What about previously saved patches that have no description?

This should probably log the patch `uniqid` and then also print the description, if present. The patch ID is shown on the page for every patch already.

sysutils/pfSense-pkg-System_Patches/files/usr/local/www/system_patches.php
```diff
@@ -48,28 +48,38 @@
...
+	switch ($_POST['act']) {
+	case 'fetch':
```
jim-p: Everything in the `switch` needs indented one more tab

sysutils/pfSense-pkg-System_Patches/files/usr/local/www/system_patches_edit.php
```diff
@@ -71,8 +71,8 @@
 
 	/* input validation */
 	if (empty($_POST['location'])) {
-		$reqdfields = explode(" ", "patch");
-		$reqdfieldsn = array(gettext("Patch Contents"));
+		$reqdfields = explode(" ", "descr patch");
```
jim-p: I don't think description should be required. It wasn't in the past, and there are likely many patch entries already present without it. There is no compelling reason to force it.

vktg:
+ tabs to switch statement
+ mandatory description removed
+ uniqid in log, descr if present

sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
```diff
@@ -227,4 +229,19 @@ function patch_remove_shellcmd() {
...
 }
 
+function patchlog($msg) {
+        syslog(LOG_WARNING, gettext("System Patches: {$msg}"));
+}
+
+function patch_descr($id) {
+	global $config;
+	$a_patches = &$config['installedpackages']['patches']['item'];
+	$descr = gettext(" (ID: {$a_patches[$id]['uniqid']}");
```
jim-p: The only part of this string that should be run through `gettext()` is `ID`. The rest is formatting and variables.

sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
```diff
@@ -227,4 +229,19 @@ function patch_remove_shellcmd() {
...
+	if (isset($a_patches[$id]['descr'])) {
+		$descr .= gettext(", DESCR: {$a_patches[$id]['descr']}");
```
jim-p: The only part of this string that should be run through `gettext()` is `DESCR`. The rest is formatting and variables. And that label is probably not necessary.

sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
vktg: gettext() fixed

sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches.inc
vktg: gettext() fixed

vktg:
done

PORTREVISION +1

vktg:
$descr = " (" . gettext("ID") . ": {$a_patches[$id]['uniqid']}"**)**;
                                                                                             ^

vktg:
with fixes:
https://github.com/pfsense/FreeBSD-ports/pull/692

PR #687, opened 2019-10-14
Bump haproxy package versions

hoerup:
https://redmine.pfsense.org/issues/9521

### DIFF
--- a/net/pfSense-pkg-haproxy-devel/Makefile
+++ b/net/pfSense-pkg-haproxy-devel/Makefile
@@ -1,8 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-haproxy-devel
-PORTVERSION=	0.59
-PORTREVISION=	20
+PORTVERSION=	0.60
 CATEGORIES=	net
 MASTER_SITES=	# empty
 DISTFILES=	# empty
@@ -13,7 +12,7 @@ COMMENT=	pfSense package haproxy-devel
 
 LICENSE=	APACHE20
 
-RUN_DEPENDS=	haproxy18>0:net/haproxy18
+RUN_DEPENDS=	haproxy>0:net/haproxy
 
 CONFLICTS=	pfSense-pkg-haproxy-[0-9]*
 
--- a/net/pfSense-pkg-haproxy/Makefile
+++ b/net/pfSense-pkg-haproxy/Makefile
@@ -1,8 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-haproxy
-PORTVERSION=	0.59
-PORTREVISION=	19
+PORTVERSION=	0.60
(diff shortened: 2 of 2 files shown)


rbgarga:
@PiBa-NL could you please take a look at it?

PiBa-NL:
TLDR: this PR can be committed imho.

I think updating to 1.8 for the stable haproxy package and 2.0 for the haproxy-devel package is a good step to do. 1.8 has been stable for quite a while so no one should really be needing 1.7.x anymore, and 2.0.7 seems to be evolving to a pretty stable version as well, that wasn't quite so for the earlier 2.0.x releases.. (1.9 is a more 'short term' development release cycle, just as 2.1 will be.. so those should imho be skipped as a pfSense package..)

And yes it isn't really the latest/greatest haproxy-devel 2.1-devX port development branche as some people might think given the name of the pfSense package.. but to be honest that version gets updated to often to follow here and while offering 'bleeding edge' features they do also suffer from a variety of bugs that would need debugging. As such i believe that people wanting to use such a development version should be willing to compile their own version of haproxy binaries which then also them to bisect commits that caused a issue and applying/validating possible fixes to the the code that could be proposed by haproxy developers.

hoerup:
I agree with @PiBa-NL that this is the best scenario for now. 
But depending on how haproxy 2.0 evolves and stabilizes during pfsense 2.5 development cycle, we might want to re-evaluate if a bump to haproxy 2.0 for pkg-pfsense-haproxy is reasonable.

net/pfSense-pkg-haproxy-devel/Makefile
```diff
@@ -1,8 +1,8 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-haproxy-devel
-PORTVERSION=	0.59
-PORTREVISION=	20
+PORTVERSION=	0.60
+PORTREVISION=	1
```
rbgarga: PORTREVISION line must be deleted when PORTVERSION changes

net/pfSense-pkg-haproxy/Makefile
```diff
@@ -1,8 +1,8 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-haproxy
-PORTVERSION=	0.59
-PORTREVISION=	19
+PORTVERSION=	0.60
+PORTREVISION=	1
```
rbgarga: PORTREVISION line must be deleted when PORTVERSION changes

net/pfSense-pkg-haproxy-devel/Makefile
hoerup: Done

net/pfSense-pkg-haproxy/Makefile
hoerup: Done

PR #689, opened 2019-10-22
dns/pfSense-pkg-bind: freeze dynamic zones

overhacked:
Fixes [Bug #8258](https://redmine.pfsense.org/issues/8258)

DDNS-enabled zones are corrupted by any changes made
to their DB files during `bind_sync()`, because it causes
them to conflict with the .jnl files BIND maintains
for DDNS updates. The corruption causes BIND to refuse
to load the zone until the .jnl file is manually deleted.

This patch causes pfSense to run rndc freeze/thaw
before/after changing the zone files so that the
zones are no longer considered corrupt by BIND.

This patch quickly resolves the issue of pfSense causing
BIND to refuse to load the zones, but it does have the
effect that all DDNS changes to the zone are overwritten
with pfSense's saved XML zone configuration every time
`bind_sync()` runs.

This overwriting behavior is still an improvement on the
current situation in which the user must manually discard
all dynamic changes by deleting the .jnl file, so this
maintains the status quo behavior of dynamic zones and
fixes a bug.

A future improvement, but much more complex to implement,
will be to persist the dynamic changes to the zone either
by adding them to pfSense's custom zone records or
temporarily caching them while bind_sync() rewrites the
DB file and adding them back to the DB before thaw. Either
approach will require parsing the DB file with PHP
and diff-ing it with what pfSense expects it to be. Another
possibility would be to add use of the
[$INCLUDE directive](http://www.zytrax.com/books/dns/ch8/include.html)
to pfSense's zone generation if a zone is dynamic so that
BIND puts dynamic updates in one DB file and pfSense maintains
a separate DB file that corresponds to the XML configuration.

### DIFF
--- a/dns/pfSense-pkg-bind/files/usr/local/pkg/bind.inc
+++ b/dns/pfSense-pkg-bind/files/usr/local/pkg/bind.inc
@@ -586,11 +586,11 @@ EOD;
 						}
 
 						// Freeze dynamic zones to prevent journal corruption
-						$zone_is_dynamic = $zone['enable_updatepolicy'] == "on" || $zoneallowupdate != "none";
+						$zone_is_dynamic = (($zone['enable_updatepolicy'] == "on") || ($zoneallowupdate != "none"));
 						$rndc_conf_path = BIND_LOCALBASE . "/etc/rndc.conf";
-						$rndc = "/usr/local/sbin/rndc -q -c $rndc_conf_path";
+						$rndc = "/usr/local/sbin/rndc -q -c {$rndc_conf_path}";
 						if ($zone_is_dynamic) {
-							exec("$rndc freeze " . escapeshellarg($zonename) . " IN " . escapeshellarg($zoneview));
+							exec("{$rndc} freeze " . escapeshellarg($zonename) . " IN " . escapeshellarg($zoneview));
 							// TODO: diff frozen zone DB with pfSense's stored DB file
 							//       and (optionally?) add dynamic records to customzonerecords
 						}
@@ -600,7 +600,7 @@ EOD;
 
 						// Thaw frozen dynamic zone
 						if ($zone_is_dynamic) {
-							exec("$rndc thaw " . escapeshellarg($zonename) . " IN " . escapeshellarg($zoneview));
+							exec("{$rndc} thaw " . escapeshellarg($zonename) . " IN " . escapeshellarg($zoneview));
 						}
 
 						$config['installedpackages']['bindzone']['config'][$x]['resultconfig'] = base64_encode($zone_conf);
--- a/dns/pfSense-pkg-bind/files/usr/local/pkg/bind.inc
+++ b/dns/pfSense-pkg-bind/files/usr/local/pkg/bind.inc
@@ -590,7 +590,7 @@ EOD;
 						$rndc_conf_path = BIND_LOCALBASE . "/etc/rndc.conf";
 						$rndc = "/usr/local/sbin/rndc -q -c $rndc_conf_path";
 						if ($zone_is_dynamic) {
-							exec("$rndc freeze $zonename IN $zoneview");
+							exec("$rndc freeze " . escapeshellarg($zonename) . " IN " . escapeshellarg($zoneview));
 							// TODO: diff frozen zone DB with pfSense's stored DB file
 							//       and (optionally?) add dynamic records to customzonerecords
 						}
@@ -600,7 +600,7 @@ EOD;
 
 						// Thaw frozen dynamic zone
 						if ($zone_is_dynamic) {
-							exec("$rndc thaw $zonename IN $zoneview");
+							exec("$rndc thaw " . escapeshellarg($zonename) . " IN " . escapeshellarg($zoneview));
 						}
 
 						$config['installedpackages']['bindzone']['config'][$x]['resultconfig'] = base64_encode($zone_conf);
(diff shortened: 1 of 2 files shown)


dns/pfSense-pkg-bind/files/usr/local/pkg/bind.inc
```diff
@@ -586,9 +586,24 @@ EOD;
 							$zone_conf .= "\n\n;\n;custom zone records\n;\n".base64_decode($zone['customzonerecords'])."\n";
 						}
 
+						// Freeze dynamic zones to prevent journal corruption
+						$zone_is_dynamic = $zone['enable_updatepolicy'] == "on" || $zoneallowupdate != "none";
+						$rndc_conf_path = BIND_LOCALBASE . "/etc/rndc.conf";
+						$rndc = "/usr/local/sbin/rndc -q -c $rndc_conf_path";
+						if ($zone_is_dynamic) {
+							exec("$rndc freeze $zonename IN $zoneview");
```
jim-p: Since this is a shell exec using user input (zone name and view), each of these variable uses should be wrapped in `escapeshellarg()`, for example:

    exec("$rndc freeze " . escapeshellarg($zonename) . " IN " . escapeshellarg($zoneview));

dns/pfSense-pkg-bind/files/usr/local/pkg/bind.inc
```diff
@@ -586,9 +586,24 @@ EOD;
...
+							// TODO: diff frozen zone DB with pfSense's stored DB file
+							//       and (optionally?) add dynamic records to customzonerecords
+						}
+
 						// Save zone configuration DB file
 						file_put_contents(CHROOT_LOCALBASE."/etc/namedb/{$zonetype}/{$zoneview}/{$zonename}.DB", $zone_conf);
 
+						// Thaw frozen dynamic zone
+						if ($zone_is_dynamic) {
+							exec("$rndc thaw $zonename IN $zoneview");
```
jim-p: See previous `escapeshellarg()` note.

dns/pfSense-pkg-bind/files/usr/local/pkg/bind.inc
```diff
@@ -591,7 +591,7 @@ EOD;
 						$rndc_conf_path = BIND_LOCALBASE . "/etc/rndc.conf";
 						$rndc = "/usr/local/sbin/rndc -q -c $rndc_conf_path";
 						if ($zone_is_dynamic) {
-							exec("$rndc freeze $zonename IN $zoneview");
+							exec("$rndc freeze " . escapeshellargs($zonename) . " IN " . escapeshellargs($zoneview));
```
jim-p: These should all be `escapeshellarg()`, not `escapeshellargs()` (no `s` on the end)

overhacked: :flushed: I guess I get what I deserve for trying to "fix a small issue" when I'm not in front my my main workstation. Correct fixing commit pushed already.

dns/pfSense-pkg-bind/files/usr/local/pkg/bind.inc
```diff
@@ -586,9 +586,24 @@ EOD;
 							$zone_conf .= "\n\n;\n;custom zone records\n;\n".base64_decode($zone['customzonerecords'])."\n";
 						}
 
+						// Freeze dynamic zones to prevent journal corruption
+						$zone_is_dynamic = $zone['enable_updatepolicy'] == "on" || $zoneallowupdate != "none";
```
jim-p: This could use some parenthesis around the logic to make it more clear what is going on. For example:
`(($zone['enable_updatepolicy'] == "on") || ($zoneallowupdate != "none"))`

dns/pfSense-pkg-bind/files/usr/local/pkg/bind.inc
```diff
@@ -586,9 +586,24 @@ EOD;
...
+						$rndc_conf_path = BIND_LOCALBASE . "/etc/rndc.conf";
+						$rndc = "/usr/local/sbin/rndc -q -c $rndc_conf_path";
```
jim-p: Variables in strings should be enclosed in `{}`, for example `"/usr/local/sbin/rndc -q -c {$rndc_conf_path}"`

dns/pfSense-pkg-bind/files/usr/local/pkg/bind.inc
```diff
@@ -586,9 +586,24 @@ EOD;
...
+						if ($zone_is_dynamic) {
+							exec("$rndc freeze " . escapeshellarg($zonename) . " IN " . escapeshellarg($zoneview));
```
jim-p: Should be `{$rndc}`

dns/pfSense-pkg-bind/files/usr/local/pkg/bind.inc
```diff
@@ -586,9 +586,24 @@ EOD;
...
+							// TODO: diff frozen zone DB with pfSense's stored DB file
+							//       and (optionally?) add dynamic records to customzonerecords
+						}
+
 						// Save zone configuration DB file
 						file_put_contents(CHROOT_LOCALBASE."/etc/namedb/{$zonetype}/{$zoneview}/{$zonename}.DB", $zone_conf);
 
+						// Thaw frozen dynamic zone
+						if ($zone_is_dynamic) {
+							exec("$rndc thaw " . escapeshellarg($zonename) . " IN " . escapeshellarg($zoneview));
```
jim-p: Should be `{$rndc}`

overhacked:
This branch is based on RELENG_2_4_4 for testing purposes. Shall I go ahead and rebase onto devel or master?

jim-p:
I didn't catch that before, but yes, it must be against devel in this repo. Rebase might not be viable since it might attach all the commits to your PR, you'll probably need to make a new PR so it only contains your intended changes.

overhacked:
Successfully rebased to devel.

rbgarga:
It was manually merged. Thanks!

PR #690, opened 2019-10-22
- fix restart of bgpd

dinoex:
bugfix for https://redmine.pfsense.org/issues/9836


net/pfSense-pkg-OpenBGPD/files/usr/local/pkg/openbgpd.inc
```diff
@@ -266,7 +266,7 @@ EOF;
 }
 
 function openbgpd_restart() {
-	if (is_openbgpd_running() == true) {
+	if ((is_openbgpd_running() == true) && (file_exists("/var/run/bgpd.sock"))) {
```
rbgarga: The check for `/var/run/bgpd.sock` should be done inside `is_openbgpd_running()` so other places that call it will benefit of it

dinoex: For now is only one caller, the restart function.
If the service is runs without the socket, it may still be functional.
Suggested changes added.

PR #691, opened 2019-10-23
switch clog to cat in status_tinc.php

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9740
Ready for review

There is no /usr/local/sbin/clog in pfSense 2.5
using "cat" instead


security/pfSense-pkg-tinc/files/usr/local/www/status_tinc.php
```diff
@@ -26,8 +26,15 @@ function tinc_status_usr1() {
 	usleep(500000);
 	$clog_path = "/usr/local/sbin/clog";
 	$result = array();
+	$logfile = "/var/log/tinc.log";
+	$firmware = host_firmware_version();
 
-	exec("{$clog_path} /var/log/tinc.log | /usr/bin/sed -e 's/.*tinc\[.*\]: //'", $result);
+	if ($firmware['config_version'] >= 19.7) {
+		exec(system_log_get_cat() . ' ' . $logfile . "| /usr/bin/sed -e 's/.*tinc\[.*\]: //'", $result);
```
jim-p: Rather than `$logfile` here, it should use `sort_related_log_files()` like the other example.

security/pfSense-pkg-tinc/files/usr/local/www/status_tinc.php
```diff
@@ -26,8 +26,15 @@ function tinc_status_usr1() {
...
+	} else {
+		exec("{$clog_path} /var/log/tinc.log | /usr/bin/sed -e 's/.*tinc\[.*\]: //'", $result);
```
jim-p: Why use the full hardcoded path to the log here, rather than `{$logfile}` which is populated above?

security/pfSense-pkg-tinc/files/usr/local/www/status_tinc.php
```diff
@@ -26,8 +26,15 @@ function tinc_status_usr1() {
 	usleep(500000);
 	$clog_path = "/usr/local/sbin/clog";
 	$result = array();
+	$logfile = "/var/log/tinc.log";
+	$firmware = host_firmware_version();
 
-	exec("{$clog_path} /var/log/tinc.log | /usr/bin/sed -e 's/.*tinc\[.*\]: //'", $result);
+	if ($firmware['config_version'] >= 19.7) {
```
jim-p: This isn't a proper version test. The config revision is not a reliable indicator of the pfSense version. It should be based on `$g['product_version']` or similar to test the actual version directly.

security/pfSense-pkg-tinc/files/usr/local/www/status_tinc.php
```diff
@@ -56,8 +63,15 @@ function tinc_status_usr2() {
...
+	$logfile = "/var/log/tinc.log";
+	$firmware = host_firmware_version();
 
-	exec("{$clog_path} /var/log/tinc.log | sed -e 's/.*tinc\[.*\]: //'",$result);
+	if ($firmware['config_version'] >= 19.7) {
+		exec(system_log_get_cat() . ' ' . $logfile . "| /usr/bin/sed -e 's/.*tinc\[.*\]: //'", $result);
+	} else {
+		exec("{$clog_path} /var/log/tinc.log | /usr/bin/sed -e 's/.*tinc\[.*\]: //'", $result);
+	}
+
```
jim-p: See comments above which also apply to this similar block.
(And since they are practically identical, I have to wonder if this could be simplified significantly, though that's likely out of scope for this particular PR)

security/pfSense-pkg-tinc/files/usr/local/www/status_tinc.php
jim-p: No, it is not. That does not necessarily determine the pfSense version either. It must use product_version. If you need to compare, there are convenience functions like `pfs_version_compare()` that facilitate a proper comparison.

security/pfSense-pkg-tinc/files/usr/local/www/status_tinc.php
```diff
@@ -21,78 +21,54 @@
...
+		} else {
+		exec("{$clog_path} {$logfile} | /usr/bin/sed -e 's/.*tinc\[.*\]: //'", $result);
 	}
 	
 	$i = 0;
-	foreach ($result as $line) {
-		if (preg_match("/Connections:/", $line)) {
-			$begin = $i;
+	if ($usr == 'USR1') {
+		foreach ($result as $line) {
```
jim-p: The USR1 and else block are still pretty similar. Looks like you could use the `if ($usr == 'USR1') ... else` test to set the beginning and ending match strings only, and then just have one `foreach`.

security/pfSense-pkg-tinc/files/usr/local/www/status_tinc.php
```diff
@@ -22,40 +22,37 @@
...
-	} else {
-		foreach ($result as $line) {
-			if (preg_match("/Statistics for Generic BSD (tun|tap) device/",$line)) {
-				$begin = $i;
-			}
-			if (preg_match("/End of subnet list./",$line)) {
-				$end = $i;
-			}
-			$i++;
+	foreach ($result as $line) {
```
jim-p: I was thinking more like this:

```
	$matchbegin = ($usr == 'USR1') ? "Connections:" : "Statistics for Generic BSD (tun|tap) device";
	$matchend = ($usr == 'USR1') ? "End of connections." : "End of subnet list.";
	foreach ($result as $line) {
		if (preg_match("/{$matchbegin}/", $line)) {
			$begin = $i;
		}
		if (preg_match("/{$matchend}/", $line)) {
			$end = $i;
		}
		$i++;
	}
```

vktg: done

PR #693, opened 2019-10-27
show huperscan option only for x86 arch

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9850
 Ready for review

Hyyperscan will run on x86 processors in 64-bit (Intel® 64 Architecture) and 32-bit (IA-32 Architecture) modes.

hide Hyperscan Pattern Matcher Algorithm in case of ARM arch


rbgarga:
@bmeeks8 can you check this one please?

bmeeks8:
This is correct and was an oversight on my part. The Hyperscan library is only available for X86 architectures, so removing the option from ARM versions of the GUI is correct.

PR #699, opened 2019-11-06
pfSense-pkg-snort-3.2.9.10 -- Support 2.9.15 binary and bug fixes

bmeeks8:
### pfSense-pkg-snort-3.2.9.10
This update adds support for the latest 2.9.15 version of the snort binary, fixes two bugs and adds one new feature.

**New Features:**
1. Added sortable columns on the RULES tab to duplicate similar functionality available on the ALERTS tab.

**Bug Fixes:**
1. When adding a SID to a Suppress List or disabilng a rule on the ALERTS tab, Snort is stopped if running and logs a "_Snort Reload: Any change to the dynamic preprocessor configuration requires a restart._" message to the system log.

2. The SUPPRESS tab allows deletion of an assigned Suppress List when it should instead issue a warning and not delete lists currently assigned to an interface.

### DIFF
--- a/security/pfSense-pkg-snort/Makefile
+++ b/security/pfSense-pkg-snort/Makefile
@@ -1,8 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-snort
-PORTVERSION=	3.2.9.9
-PORTREVISION=	1
+PORTVERSION=	3.2.9.10
 CATEGORIES=	security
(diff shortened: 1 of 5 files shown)


bmeeks8:
Before merging this request, the Snort-2.9.15 binary update needs to be cherry-picked into RELENG_2_4_4 from the pfSense DEVEL branch.

PR #702, opened 2019-11-09
Make snort rules-update handle failed TAR extractions

DumboJetEngine:
When using RAM disks for `/tmp`, I get corruptions to `.so` files, apparently due to unavailable storage space for the extraction during the rules update.
I would understand failing to update, but corrupting the files and making snort not to function seemed a bit too much to me.
Here I try to check whether the TAR command has failed, to prevent file copy when this happens.

Disclaimer:
This code is not tested, since I have no idea how to test it.
If you can test it and then merge it (when all seems working great), it would be super.
Maybe some extra error logs could be added when the extraction fails, but probably that is a minor detail compared to the problem solved here.

If you have any feedback, shoot! :)
And if you don't like the changes, I won't be offended.


DumboJetEngine:
Hm...
On second review, maybe the `rmdir_recursive()` calls should be outside the IF blocks.
It would provide better cleanup.

DumboJetEngine:
Cleanup issue fixed! :)

security/pfSense-pkg-snort/files/usr/local/pkg/snort/snort_check_for_rule_updates.php
```diff
@@ -377,6 +377,20 @@ function snort_fetch_new_rules($file_url, $file_dst, $file_md5, $desc = "") {
...
+	if($extra != null) $cmd .= " {$extra}";
+	$ret = snort_exec($cmd);
+	return $ret === 0;
+}
+
+function snort_exec($cmd, &$output = null){
+	$return = -1;
+	exec($cmd, $output, $return);
+	return $return;
+}
```
rbgarga: This function is not necessary.  exec() call can happen inside snort_untar() without any harm

rbgarga:
@bmeeks8 can you check this PR please?

bmeeks8:
This will also need testing to ensure it actually performs as intended. The author stated on the pfSense forums that he had not tested this new code.

I will test it, but it will be another day before I can get to it. Busy on another task today.

bmeeks8:
I also see no feedback to the user via logging or any other means if the untar procedure fails. The only logged messages are for "success" operations. If the untar procedure fails, the user's rules will simply not be updated and no warning or error message will be logged.

The code needs to acknowledge and handle the situation where the untar procedure fails.

security/pfSense-pkg-snort/files/usr/local/pkg/snort/snort_check_for_rule_updates.php
DumboJetEngine: I have created it because I saw that there were many other calls to `exec()` in there, that did not check for success/failure. Maybe you can reuse it in the future. I thought of replacing those calls myself but then the chances of breaking something would be more, since I cannot test the code.

DumboJetEngine: But if you are sure I should keep it as one function, I can do that. :)

DumboJetEngine:
@bmeeks8 
Yes, I have mentioned this in the PR description.
I tried to keep the fix simple and not have to involve you into text translations as well.
If I log a general error with this code, from inside the `snort_untar()` function, will it be OK? :
```
error_log(gettext("\tFailed to extract a TAR archive. Some snort rules might still be out-of-date. Make sure there is enough free disk space and try again...\n"), 3, SNORT_RULES_UPD_LOGFILE);
```

bmeeks8:
> @bmeeks8
> Yes, I have mentioned this in the PR description.
> I tried to keep the fix simple and not have to involve you into text translations as well.
> If I log a general error with this code, from inside the `snort_untar()` function, will it be OK? :
> 
> ```
> error_log(gettext("\tFailed to extract a TAR archive. Some snort rules might still be out-of-date. Make sure there is enough free disk space and try again...\n"), 3, SNORT_RULES_UPD_LOGFILE);
> ```

Logging to the Update Log File is good, but I would also consider logging an extra message to the system log. You can do that by also calling log_error() with the message string. That is a standard PHP function as well. You will see some other places in that rules update PHP module where messages are sent to both locations (the rules update log and to the system log).

DumboJetEngine:
@rbgarga, @bmeeks8 
The changes are done. :)
I have also used `syslog(LOG_ERR, ....` to log the error in the system error logs.

bmeeks8:
> @rbgarga, @bmeeks8
> The changes are done. :)
> I have also used `syslog(LOG_ERR, ....` to log the error in the system error logs.

Thanks! I switched over to using _syslog()_ calls in the new 4.0 version of the Snort GUI used with pfSense-2.5. I just have not gone back and backported all of those changes to the Snort 3.2.9.x branch. Using _syslog()_ in the 3.2.9.x branch is fine.

bmeeks8:
The last step will be a test of the new code. I don't know of a method for producing a corrupted tar gzip archive, but I suppose at least the out-of-disk-space issue could be produced by purposefully configuring a too-small **/tmp** RAM disk and then attempting a rules update. Making **/tmp** 25 MB or less should result in a rules update failure.

@DumboJetEngine, can you try this on your end? Maybe in a virtual machine running pfSense?

DumboJetEngine:
@bmeeks8 
I think I can try.
Is this what I should do? :

* Switch the "System Update" to the development branch (Experimental 2.5.x DEVEL)
* Update the Snort package to the latest version (which will contain the fix)
* Do some "Force Updates" on the snort rules (while using a RAM disk for `/tmp`)

Can you tell me where on disk is the code for the snort package stored, so that I can verify that my code changes are indeed applied, when I upgrade Snort?

I am not sure if I should also upgrade pfSense itself using the development branch, so that I can also test with the development version. But I guess that the stable branch is the main focus, right?

bmeeks8:
> @bmeeks8
> I think I can try.
> Is this what I should do? :
> 
> * Switch the "System Update" to the development branch (Experimental 2.5.x DEVEL)
> * Update the Snort package to the latest version (which will contain the fix)
> * Do some "Force Updates" on the snort rules (while using a RAM disk for `/tmp`)
> 
> Can you tell me where on disk is the code for the snort package stored, so that I can verify that my code changes are indeed applied, when I upgrade Snort?
> 
> I am not sure if I should also upgrade pfSense itself using the development branch, so that I can also test with the development version. But I guess that the stable branch is the main focus, right?

You can do this two ways. You can swap your pfSense setup to use 2.5-DEVEL or you can stay on 2.4.4-RELEASE. Either way you have to do the following step to manually copy over an updated (changed) PHP source file, so I would suggest just staying on RELEASE and doing this:

1. Configure a RAM disk for /tmp that you know will be too small for a successful update. I'm guessing 25 MB, but depending on number of enabled rules archives you may need to make it smaller.

2. Copy over your "modified" _snort_check_for_rules_update.php_ file to **/usr/local/pkg/snort**. This will overwrite the existing file, so you might want to make a copy of it first.

3. Test a forced rules update by clicking the **Force** button on the UPDATES tab. You should get an error/warning message and the update not happen and Snort keep running if things went well.

4. Either get rid of the RAM disk or greatly increase its size and repeat the forced update to be sure things still work like they are supposed to.

DumboJetEngine:
OK. :)
I have just tested it and it seems working nice!

I started with a RAM disk of 60MiB and it was failing with the most frequent failure being `Failed writing body`:
Fix 1

Then I increased it to 170MiB and after a couple retries, I managed to make the error appear :
Fix 3

After that failure (and all failures in general), Snort seemed to be running fine after its automatic restart. I even tried rebooting the entire OS, just to make sure all was OK with the new files.
I have tried several times and it seems that I can't break it, so it looks like a success!

One thing that I noticed (and that existed before my changes too), was that on the failed rulesets it was showing a `Not Downloaded`, which (I think) is a bit misleading, because it suggests that Snort runs without those rulesets :
Fix 4

But I think that is not the case, since I haven't yet spotted a place in the code deleting the old files. Anyway, maybe that is something to improve in the future.

Once I increased the RAM disk size, it was kind of hard to successfully complete a "Force Update", because I was frequently getting server errors (500, 503) on download, but I finally managed to get some successes too:
Fix 2

The server errors seemed to happen in the past too, according to my recollection. Probably the update server is under heavy load.

Normal update (without force) works fine too.
So, I think all is good!
:) :) :)

bmeeks8:
You are getting the "Not Downloaded" message because the way the "force update" code works is by simply deleting the md5 checksum file maintained for each set of enabled rule vendors. The Update page code sees that you have say Snort GPLv2 rules enabled by there is no md5 checksum file to test against the current version posted on the web site, so the code assumes nothing has been downloaded yet.

If you had a typical update situation from the periodic cron task, the md5 file is not deleted so the Updates tab would still show the old file's timestamp and md5 checksum.

bmeeks8:
Renato:
I am okay with this change now if you are satisfied with the final result compared to the changes you requested.

DumboJetEngine:
@bmeeks8 
No problem!  :)

rbgarga:
> Renato:
> I am okay with this change now if you are satisfied with the final result compared to the changes you requested.

@bmeeks8 should I cherry-pick it to 2.4.x?

DumboJetEngine:
@rbgarga Done. :)

PR #709, opened 2019-11-15
show ECDSA CAs only with correct curves

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9901
 Ready for review

Do not show incompatible ECDSA CAs for Squid HTTPS/SSL Interception
same as https://redmine.pfsense.org/issues/9897


vktg:
same func as in https://github.com/pfsense/FreeBSD-ports/pull/710 and https://github.com/pfsense/FreeBSD-ports/pull/712

I think we need to add this func to cert.inc:
`function pkg_get_ca_or_certs($type, $none=False) {
	$c_arr = array();
	if ($none != False) {
	        $c_arr[] = array('refid' => 'none', 'descr' => 'none (auto)');
	}
	$ecdsagood = cert_build_list($type, 'IPsec');
	foreach ($ecdsagood as $refid => $descr) {
		$c_arr[] = array('refid' => $refid, 'descr' => $descr);
	}
	return $c_arr;
}`

jim-p:
> I think we need to add this func to cert.inc

I'll try to come up with a good generic option this week, then it can be used on all packages.

rbgarga:
> > I think we need to add this func to cert.inc
> 
> I'll try to come up with a good generic option this week, then it can be used on all packages.

In the same minute I pushed #712 I thought about it.  Let me know when the generic function is available and I take care of using it on stunnel and freeradius

jim-p:
> In the same minute I pushed #712 I thought about it. Let me know when the generic function is available and I take care of using it on stunnel and freeradius

I added an extra parameter to `cert_build_list()` which will generate this output format. https://redmine.pfsense.org/issues/9923  https://github.com/pfsense/pfsense/commit/59fac81f316b0616e0c50ec47ffa9cfa97a10ebb
Additionally, if you add a `function_exists` test that looks for `cert_build_list()`, packages could fall back to the old behavior, so that we can keep them in sync between 2.4.4 and 2.5.0.

PR #710, opened 2019-11-16
freeradius: show ECDSA CAs and certs only with correct curves

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9906
Ready for review

Do not show incompatible ECDSA CAs or certs for FreeRADIUS
same as https://redmine.pfsense.org/issues/9897


vktg:
Yes, it seems that using the IPsec list is safe

See  https://www.sevecek.com/EnglishPages/Lists/Posts/Post.aspx?ID=57 about Windows Suite-B support
and https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-platform-common-criteria , where you can see only “NIST curves” **P-256**, **P-384** and **P-521** for latest Windows versions

net/pfSense-pkg-freeradius3/files/usr/local/pkg/freeradius.inc
```diff
@@ -1601,18 +1601,22 @@ EOD;
...
-		foreach ($config['ca'] as $ca) {
-			$ca_arr[] = array('refid' => $ca['refid'], 'descr' => $ca['descr']);
-		}
+function freeradius_get_ca_or_certs($type) {
+	$c_arr = array();
+	if ($type == 'ca') {
+		$ecdsagood = cert_build_list('ca', 'IPsec');
+	} else {
+		$ecdsagood = cert_build_list('cert', 'IPsec');
+		$c_arr[] = array('refid' => 'none', 'descr' => 'none (auto)');
```
jim-p: Why is this only added to the certificate array? Before it was added to both.

net/pfSense-pkg-freeradius3/files/usr/local/pkg/freeradius.inc
```diff
@@ -1601,18 +1601,22 @@ EOD;
...
-	$ca_arr = array();
-	$ca_arr[] = array('refid' => 'none', 'descr' => 'none (auto)');
-
-	if (is_array($config['ca'])) {
-		foreach ($config['ca'] as $ca) {
-			$ca_arr[] = array('refid' => $ca['refid'], 'descr' => $ca['descr']);
-		}
+function freeradius_get_ca_or_certs($type) {
+	$c_arr = array();
+	if ($type == 'ca') {
```
jim-p: This can be simplified a bit since you are passing the same string as `$type` on to `cert_build_list()`, so the whole if block can be changed to something like:

```
$c_arr = array();
$c_arr[] = array('refid' => 'none', 'descr' => 'none (auto)');
$ecdsagood = cert_build_list($type, 'IPsec');
```

net/pfSense-pkg-freeradius3/files/usr/local/pkg/freeradius.inc
vktg: yeap, fixed

net/pfSense-pkg-freeradius3/files/usr/local/pkg/freeradius.inc
vktg: done

PR #712, opened 2019-11-23
stunnel: check ECDSA certs curves

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9919
Ready for review

stunnel client can use cert with any ECDSA curve,
but if stunnel server use incorrect (not prime256v1, secp384r1, secp521r1) curve, an error occurs:
`SSL_connect: /build/ce-crossbuild-master/pfSense/tmp/FreeBSD-src/crypto/openssl/ssl/record/rec_layer_s3.c:1528: error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure`

same func as https://github.com/pfsense/FreeBSD-ports/pull/710
maybe we can create system-wide pkg_get_ca_or_certs($type) function?

### DIFF
--- a/security/pfSense-pkg-stunnel/Makefile
+++ b/security/pfSense-pkg-stunnel/Makefile
@@ -2,6 +2,7 @@
 
 PORTNAME=	pfSense-pkg-stunnel
 PORTVERSION=	5.50
+PORTREVISION=	1
 CATEGORIES=	security
 MASTER_SITES=	# empty
 DISTFILES=	# empty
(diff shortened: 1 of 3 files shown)


jim-p:
I added an extra parameter to `cert_build_list()` which will generate this output format. https://redmine.pfsense.org/issues/9923  https://github.com/pfsense/pfsense/commit/59fac81f316b0616e0c50ec47ffa9cfa97a10ebb
Additionally, if you add a `function_exists` test that looks for `cert_build_list()`, packages could fall back to the old behavior, so that we can keep them in sync between 2.4.4 and 2.5.0.

PR #714, opened 2019-11-27
filterdns - always wake action thread from hostname thread

gav-:
Redmine Issue: https://redmine.pfsense.org/issues/9296

This change to filterdns.c causes the action thread to always be woken from the hostname thread, regardless of whether the resolved address has changed or not. 

This change is necessary for when a new alias is added with the same hostname in a previous alias, and the hostname thread never wakes the new action thread to populate the table since the resolved address hasn't changed.

### DIFF
--- a/net/filterdns/files/filterdns.c
+++ b/net/filterdns/files/filterdns.c
@@ -575,16 +575,14 @@ check_hostname(void *arg)
 			}
 		}
 
-		if (update > 0) {
-			if (debug >= 4)
-				syslog(LOG_WARNING,
-				    "Change detected on host: %s",
-				    thr->hostname);
-			TAILQ_FOREACH(act, &thr->actions, next_actions) {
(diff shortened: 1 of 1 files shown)


loos-br:
A fix based on this PR was committed.

Please test again with the new version and let me know if you still have problems.

Thanks!

gav-:
Committed fix tested, and confirmed to be working.

luckman212:
Just want to say **thank you** to @gav- for this important fix!

bbrendon:
@gav-  where is your tip-jar link?

gav-:
While I greatly appreciate the sentiment for identifying the cause of the bug, it was really @loos-br that crafted a far more elegant patch than my attempt.

@bbrendon I know nothing of a "tip-jar" in relation to github or pfsense, do you have a link ?

PR #715, opened 2019-11-27
openvpn client export ECDSA safe list check

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9929
Ready for review

show only ECDSA-safe exports packages on OpenVPN \ Client Export Utility page
i.e. certs with prime256v1, secp384r1, secp521r1 curves


security/pfSense-pkg-openvpn-client-export/Makefile
```diff
@@ -2,6 +2,7 @@
 
 PORTNAME=	pfSense-pkg-openvpn-client-export
 PORTVERSION=	1.4.19
+PORTREVISION=   1
```
rbgarga: You must use TAB after `=`

PR #719, opened 2019-12-14
security_file_certgen

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9652
Ready for review

squid 4 uses security_file_certgen instead of ssl_crtd
see http://www.squid-cache.org/Doc/config/sslcrtd_program/


www/pfSense-pkg-squid/files/usr/local/pkg/squid.inc
```diff
@@ -1140,7 +1140,11 @@ function squid_resync_general() {
 					if (is_dir(SQUID_SSL_DB)) {
 						mwexec("/bin/rm -rf " . SQUID_SSL_DB);
 					}
-					mwexec(SQUID_LOCALBASE . "/libexec/squid/ssl_crtd -c -s " . SQUID_SSL_DB);
+					if (pfs_version_compare(false, 2.4, $g['product_version'])) {
```
jim-p: This is not a good test, as 2.4.5 will likely have squid 4 due to changes in the ports tree. It will need to test the actual squid package version or test for another files that is only present on squid 4.x.

vktg: now it check security_file_certgen file (squid 4 only)

www/pfSense-pkg-squid/Makefile
```diff
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-squid
 PORTVERSION=	0.4.44
-PORTREVISION=	9
+PORTREVISION=	10
```
rbgarga: After #721 was reverted I bumped revision to 11. Please rebase your fork and bump it again. While there, please squash all commits

vktg: done

PR #722, opened 2019-12-18
nrpe: allow ipv6 nagios servers addresses

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9973
Ready for review

Nagios NRPE package lacks IPv6 capabilities.
1. binding IP: I can only give one IP. For Dual Stack I need to be able to give more than one IP - need additional changes, xinetd
2. binding IP: I can only give IPv4 addresses, IPv6 address not accepted - it works
3. nagios Servers: IPv6 addresses not accepted - **fixed by this PR**

### DIFF
--- a/net-mgmt/pfSense-pkg-nrpe/files/usr/local/pkg/nrpe.inc
+++ b/net-mgmt/pfSense-pkg-nrpe/files/usr/local/pkg/nrpe.inc
@@ -247,7 +247,7 @@ function nrpe_custom_php_validation_command($post, &$input_errors) {
 		}
 	}
 	foreach (explode(",", $post['allowed_hosts']) as $host) {
-		if (!empty($host) && !is_ipaddr($host)) {
+		if (!empty($host) && !is_ipaddr($host) && !is_ipaddrv6($host)) {
 			$input_errors[] = gettext("'Nagios Server(s)' entry '{$host}' is not a valid IP address.");
 		}
 	}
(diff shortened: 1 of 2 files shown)


net-mgmt/pfSense-pkg-nrpe/files/usr/local/pkg/nrpe.inc
```diff
@@ -247,7 +247,7 @@ function nrpe_custom_php_validation_command($post, &$input_errors) {
 		}
 	}
 	foreach (explode(",", $post['allowed_hosts']) as $host) {
-		if (!empty($host) && !is_ipaddr($host)) {
+		if (!empty($host) && !is_ipaddr($host) && !is_ipaddrv6($host)) {
```
jim-p: `is_ipaddr()` matches both IPv4 and IPv6, so this change would not alter any behavior.

```
pfSense shell: $host = '10.0.0.1';
pfSense shell: var_dump( (!empty($host) && !is_ipaddr($host)) );
pfSense shell: exec
bool(false)
pfSense shell: var_dump( (!empty($host) && !is_ipaddr($host) && !is_ipaddrv6($host)) );
pfSense shell: exec
bool(false)
pfSense shell: $host = '2001:db8::1';
pfSense shell: var_dump( (!empty($host) && !is_ipaddr($host)) );
pfSense shell: exec
bool(false)
pfSense shell: var_dump( (!empty($host) && !is_ipaddr($host) && !is_ipaddrv6($host)) );
pfSense shell: exec
bool(false)
```

vktg: in this case we don't need this PR at all

PR #723, opened 2019-12-20
Cellular Package Upgrade for 2.4.5

svenauhagen:
Add another modem and compatibility with python3

### DIFF
--- a/net/pfSense-pkg-cellular/Makefile
+++ b/net/pfSense-pkg-cellular/Makefile
@@ -2,7 +2,6 @@
 
 PORTNAME=	pfSense-pkg-cellular
 PORTVERSION=	1.2.2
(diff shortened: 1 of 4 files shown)


net/pfSense-pkg-cellular/Makefile
```diff
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-cellular
-PORTVERSION=	1.2.1
+PORTVERSION=	1.2.2
 PORTREVISION=	1
```
rbgarga: When PORTVERSION changes, PORTREVISION line must be deleted

svenauhagen: I removed the PORTREVISION

PR #725, opened 2019-12-20
squid: ssl/starttls with trust store CA cert storage

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9982
Ready for review

Allow to use SSL/STARTTLS for LDAP connection


www/pfSense-pkg-squid/files/usr/local/pkg/squid.inc
```diff
@@ -1092,11 +1092,19 @@ function squid_validate_auth($post, &$input_errors) {
...
 				if (empty($user)) {
 					$input_errors[] = "'LDAP server user DN' is required.";
 				} elseif (!$user) {
 					$input_errors[] = "'LDAP server user DN' must be a valid DN.";
 				}
+				if ($ldaptype == 'starttls' && $ldapversion == '2') {
+					$input_errors[] = "'STARTTLS supported only by LDAP version 3.";
+				}
+				if (pfs_version_compare(false, 2.4, $g['product_version']) && $settings['ldap_urltype'] != 'standard') {
+					$input_errors[] = "'SSL/STARTTLS feature supported only by pfSense >= 2.5.";
```
jim-p: Not sure we should completely exclude this. It only doesn't work if they need to trust a self-signed CA. If their LDAP server is using a globally trusted certificate, like one from Let's Encrypt, then it would work even on 2.4.x.
It's probably enough to just warn about that in the option text.

vktg: done

www/pfSense-pkg-squid/files/usr/local/pkg/squid_auth.xml
```diff
@@ -180,6 +180,18 @@
...
+                        <fielddescr>Transport</fielddescr>
+			<fieldname>ldap_urltype</fieldname>
+			<type>select</type>
+			<default_value>standard</default_value>
+			<options>
+				<option><name>TCP - Standard</name><value>standard</value></option>
+				<option><name>TCP - STARTTLS</name><value>starttls</value></option>
+				<option><name>SSL - Encrypted</name><value>ssl</value></option>
+			</options>
+			<description>If option 'SSL Encrypted' or 'TCP - STARTTLS' is used, you need to add the CA certificate of your AD to the Operating System Trust Store. pfSense 2.5 only feature.</description>
```
jim-p: This still needs updated:
> If option 'SSL Encrypted' or 'TCP - STARTTLS' is used, you need to add the CA certificate of your AD to the Operating System Trust Store. pfSense 2.5 only feature.

Change to:
> If 'SSL Encrypted' or 'TCP - STARTTLS' is selected, the CA certificate of the LDAP server must be trusted by the Operating System Trust Store. This is automatic for certificates signed by globally trusted CAs such as Let's Encrypt; self-signed CAs can optionally be added to the Trust Store on pfSense 2.5.

vktg: done

rbgarga:
@vktg please squash all commits into a single one with a meaning commit log message before we merge it

rbgarga:
Merge together with #725

vktg:
done, squashed

PR #732, opened 2019-12-28
allow to use lo0 interface by pfblocker for web/vip

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/10002
Ready for review

By default pfBlocker uses LAN interface for internal web-server and VIP alias, and do not allow to use Loopback for it
This will lead to issues with other packages, such as FRR OSPF, when it is running on the same interface:
https://forum.netgate.com/topic/145533/frr-0-6-3-ospf-seems-to-break-with-ip-aliases/6

This PR allows you to use Loopback interface for the web-server/VIP and sets it by default.

It also hide enc0 and openvpn-server interfaces in DNSBL page configuration,
because you can't run web-server on these interfaces (use VTI and OpenVPN interfaces for this)


rbgarga:
@BBcan177 please take a look

vktg:
BBcan177 said that the same function is already in the upstream

PR #736, opened 2019-12-31
pfblocker: allow to select webserver certificate 

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/10140
Ready for review

This feature allow user to select certificate for internal webserver of pfBlocker (lighttpd)
It saves the old /var/unbound/dnsbl_cert.pem file for backward compatibility (Default (auto) cert) and creates a new /var/unbound/dnsbl_web_cert.pem

On each 'update' it compares the fingerprint of cert from the config and cert from this file, and if it's different, updates the dnsbl_web_cert.pem file.
(maybe there is a more elegant way to verify that the certificate is changed)


rbgarga:
@BBcan177 please take a look

rbgarga:
@BBcan177 any news about this one?

vktg:
we discussed this with BBcan177 - this is unnecessary
I think this can be useful only if the company uses a special internal policy that allows only specific (int CA) certificates to be used in any software.

Closed

PR #746, opened 2020-01-04
Fix for pfsense linode dns

sreyemnayr:
Here's the PR for the main acme project: https://github.com/Neilpang/acme.sh/pull/2670
These changes allowed the Linode v4 API to work on my pfSense device. There seems to be an issue with \s needing to be hard-coded to spaces and opening brackets requiring escapes on the pfSense (FreeBSD) version of grep, I think.


jim-p:
When the fix is incorporated into acme.sh we'll pull it in when we sync with upstream. No need to maintain it or patch it custom here.

sreyemnayr:
Except that it doesn't work at all without it... but right on.

PR #747, opened 2020-01-04
Remove conditional assignments in pfSense-repo makefile

rschell:
In non-pfsense builds, these four PFSENSE variables are already defined, so renaming logic in builder_common.sh does not get applied.  Changing "?=" to "=" allows builder_common.sh logic to work as expected.

See 2.5 Development Snapshot "Minor issues with non-pfSense rename logic"  discussion and redmine #10161

### DIFF
--- a/sysutils/pfSense-repo/Makefile
+++ b/sysutils/pfSense-repo/Makefile
@@ -25,10 +25,10 @@ DATADIR=	${PREFIX}/share/pfSense
 
 .include <bsd.port.pre.mk>
 
-PFSENSE_REPOS?=	pfSense-repo pfSense-repo-devel
-PFSENSE_REPOS_${ARCH}?=	${PFSENSE_REPOS}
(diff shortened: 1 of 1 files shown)


rbgarga:
This change is wrong. These variables can be changed through pfSense main repo at `tools/conf/pfPorts/make.conf`

rschell:
Then is the issue that PFSENSE_REPOS in make.conf is hard coded to pfSense and not changed to $(PRODUCT_NAME) anywhere?

rbgarga:
You can change this line to whatever you want

https://github.com/pfsense/pfsense/blob/master/tools/conf/pfPorts/make.conf#L2

rschell:
Understood, but that requires changing another file.  By changing the port repo as suggested, the builder_common script would do it automatically.

rbgarga:
This is not the way it was intended to work. We want to have the hability to define on make.conf which repositories we will have for that version generally and per arch.

PR #751, opened 2020-01-14
Ticket #5413

n3bul4:
https://redmine.pfsense.org/issues/5413

This is a proposal for dhcpleases to use unbound-control for updating dhcp lease entries.

The way this works is by making a diff (with the standard diff command) between the current dhcp.leases file and the corresponding unbound view (called dhcpleases).
Obsolete DNS entries get removed and new ones are added (via unbound-control).

For this to work, a few changes to unbound.conf are needed:

1.) unbound needs to be configured with a view named "dhcpleases"

```
view:
    name: dhcpleases
```

2.) every interface, where clients should "see" the hostnames needs to have an access-control-view directive:

```
server:
    access-control-view: 192.168.12.0/24 dhcpleases
    access-control-view: 192.168.13.0/24 dhcpleases
    access-control-view: 192.168.14.0/24 dhcpleases
    .
    .
    .
```

The above would add access for all clients in the networks 192.168.12.0-192.168.14.0 with a netmask 255.255.255.0.

3.) The include directive for dhcpleases_entries.conf in unbound.conf has to be removed / commented out
```
# dhcp lease entries
#include: /var/unbound/dhcpleases_entries.conf
```

For a quick shot, points 1 and 2 can be added to the "Custom options" section in the pfsense DNS Resolver configuration webgui.
For point 3 to work, one would have to edit /etc/inc/unbound.inc. Function unbound_generate_config_text.
Simply search for "# dhcp lease entries" and comment out the line below as shown in point 3.
Compile dhcpleases, copy it to your pfsense box.
Make a backup of /usr/local/sbin/dhcpleases.
Kill running dhcpleases process.
Move compiled dhcpleases version to /usr/local/sbin/dhcpleases.
Restart unbound.

Final notes:

The code creates 2 files in /tmp:
/tmp/dhcpleases.sort
/tmp/dhcpleases.diff

New DNS entries are added with a bulk operation (view_local_datas), whereas delete operations are done one by one.
This is due to lack of a bulk remove operation from views in unbound-control, which would be a nice feature request.
So keep in mind, that DNS delete operations will take longer than adding DNS entries.

dhcpleases logs information about added / removed entries to system.log

### DIFF
--- a/sysutils/dhcpleases/files/dhcpleases.c
+++ b/sysutils/dhcpleases/files/dhcpleases.c
@@ -40,6 +40,10 @@
 */
 /* Code in this file is based on contributions by John Volpe. */
 
+/* Unbound synchronization via unbound-control
+	by Alexander Berkes <office@metasoft.at>
+*/
+
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <sys/event.h>
@@ -80,8 +84,7 @@ static char *HOSTS = NULL;
 static char *domain_suffix = NULL;
 static int hostssize = 0;
 static int foreground = 0;
-static int unbound = 0;
-static char *unbound_conf = NULL;
+static int bulk_remove = 0;
 
 static int fexist(char *);
 static int fsize(char *);
@@ -93,10 +96,18 @@ static time_t convert_time(struct tm);
 static int load_dhcp(FILE *, char *, char *, time_t);
 
 static int write_status(void);
-static int write_unbound_conf(void);
+static int write_unbound_diff(void);
 static void cleanup(void);
-static void signal_process(char *);
 static void handle_signal(int);
+
+static int exec_and_cb(char *cmd, int (*callback)(FILE *fp));
+static int print_cb(FILE *fp);
+static int bulk_cb(FILE *fp);
+static int del_cb(FILE *fp);
+static char* reverse_ip(char* ip);
+static int sync_unbound(void);
+static void log_kqueue_errno(void);
+static void log_kevent_errno(void);
 
 /* Check if file exists */
 static int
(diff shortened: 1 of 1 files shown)


rbgarga:
@n3bul4 before I start reviewing it please revert all changes you did to README file.  It's a copy from upstream describing the FreeBSD-ports and we should never change it.

n3bul4:
@rbgarga sorry for that, I reverted the README updates.

n3bul4:
Hi there, I added some more code.

See some explanation below:

1.) Added kevent to listen to unbound exits
dhcpleases automatically exits now when unbound is killed or shuts down. This should prevent the situation where unbound-control could be trying to update dns records while unbound has already exited, resulting in a quite long unbound-control termination delay.
The pid number is passed with the -u parameter. Further, if the -u parameter is not present unbound dns records are not updated. The -u parameter is mandatory in background.

2.) Added dns bulk remove functionality
As current unbound releases lack the functionality of removing more than one dns entry from views at a time I submitted a pull request to the unbound project which has already been merged into the master branch. So it is just a matter of time til future releases of unbound-control will contain that command. To use this feature I introduced the -b parameter which - when present - tells dhcpleases to use the bulk remove command.

3.) Removed obsolete parameter -p (pidfile)
Did this actually ever do something? I don't think so from what I can see.

4.) Added usage info

rbgarga:
@n3bul4 could you please squash all these changes into a single commit with a meaning commit log message?

n3bul4:
@rbgarga I hope this is better.

rbgarga:
> @rbgarga I hope this is better.

much better! Thanks!

jkleckner:
+1 for this improvement.  It took me quite a bit of sleuthing to figure out that this was the underlying cause of what appeared to be outages.

loos-br:
The idea is good, but the implementation needs to be polished, please do not commit this code as it stands.

Also, I don't think this PR has to be tied to #751, while it contains related work, it doesn't address or fix any of the Ticket issues.

jabbera:
This issue really sucks for large networks with short lease times.

n3bul4:
As @loos-br already statet this is for sure not a finished implementation. I just had this idea and thought it might be a good proposal for this already really long lasting issue. 

Personally, I had really bad issues with unbound starting again and again when a joining client was getting a new dhcp lease. Imagine this situation with running pfblockerng with thousands of entries, and a reload taking place whenever a dhcp lease accures. It will unavoidably take you to annoying dns autages with the current solution.

The point from my personal view is, that it is an absolute no-go (kind of design error) to let unbound restart at any given point of time, except the one when the admin is requesting a restart manually through the web interface / console. This is actually one of the main points of the ticket as the author clearly says.

@loos-br what I don't really get: 

> Also, I don't think this PR has to be tied to #751...

Am I getting something wrong, because the ticket I am referring to is ticket #5413
See the issue here: http://redmine.pfsense.org/issues/5413 

@loos-br 

> it doesn't address or fix any of the Ticket issues.

I am running on this "patch" for about 3/4 of a year. It is working for me quite well. Did not have any time to improve it further so far. How about having a discussion about the whole spectrum of problems I might not see.

Let's talk about this issue to get things better.

Cheers!

sokai:
@n3bul4 Have you seen + tried this: 
"[pfBlockerNG-devel v3.0.0 - No longer bound by Unbound!](https://forum.netgate.com/topic/158592/pfblockerng-devel-v3-0-0-no-longer-bound-by-unbound)"
For me I think that works, after a couple of hours of testing …

kaysond:
Any update on this? It's a real bummer to have DNS go out every time a dhcp lease needs to be added

luckman212:
@n3bul4 Have you heard from anyone or made any further progress?

n3bul4:
> @n3bul4 Have you heard from anyone or made any further progress?

I am sorry but haven't heard from anyone nor made any further progress. If someone could tell me what needs to be done I might be able to help.

rcmcdonald91:
I'm looking at this

kaysond:
@rcmcdonald91 any updates?

kaysond:
Bumping. @rcmcdonald91

marcos-ng:
This issue has been addressed with a different implementation using Kea. Details on the associated redmine issue.

PR #753, opened 2020-01-23
FreeRADIUS conditional modules loading if sql auth is disabled. Issue #10197

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/10197
Ready for review

When using freeRADIUS 3 package with mysql enabled for accounting only and using pfsense gui for users/authorization, the generated virtual-server-default file (/usr/local/etc/raddb/sites-enabled/default) authorize section looks like:
```
authorize {
...
    ### sql DISABLED ###
    daily
    weekly
    monthly
    forever
...
```
I believe since I am not using mysql for authorization, the daily, weekly, monthly, and forever modules are failing to load and causing the virtual server to fail to load. 

Prepending the '-' causing the modules to only be loaded if they are configured,
see https://fossies.org/linux/freeradius-server/raddb/mods-available/README.rst:
> Conditional Modules
> Version 3 allows modules to be conditionally loaded. This is useful when you want to have a virtual server which references a module, but does not require it. Instead of editing the virtual server file, you can just conditionally enable the module.
> 
> Modules are conditionally enabled by adding a "-" before their name in a virtual server.
>


rbgarga:
@vktg ticket 10165 is related to Avahi HA

vktg:
> @vktg ticket 10165 is related to Avahi HA

oh, fixed

PR #755, opened 2020-01-25
Remove rogue TABs from the pimd status page

stephenw10:


jim-p:
I made this change manually in a single commit, along with some other fixes. So this is no longer needed

PR #760, opened 2020-02-01
Set Snort autoflowbit checkbox. Issue #8830

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/8830
Ready for review

Snort (version 4.0_10), visit Global Settings, enable some rulesets, visit Update and run an update. Once that's done visit Interfaces, add an interface, go to the Categories tab and see that the Automatic flowbit resolution setting is unchecked. Below that, notice the message that reads:

> If checked, Snort will auto-enable rules required for checked flowbits. Default is Checked.

This PR simply sets the default value for this checkbox to on when adding a new interface.


bmeeks8:
I am fine with this change. Thank you Viktor for the contribution. That setting should default to "ON".

rbgarga:
@vktg this PR only applies on 2.4.5.  If desired to see this change on 2.4.5 please submit a fresh PR against RELENG_2_4_5 branch.  Thanks!

PR #770, opened 2020-02-15
Set Snort autoflowbit checkbox (2.4.5). Issue #8830

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/8830
Ready for review

same as https://github.com/pfsense/FreeBSD-ports/pull/760 but for pfSense 2.4.5

Snort (version 4.0_10), visit Global Settings, enable some rulesets, visit Update and run an update. Once that's done visit Interfaces, add an interface, go to the Categories tab and see that the Automatic flowbit resolution setting is unchecked. Below that, notice the message that reads:

> If checked, Snort will auto-enable rules required for checked flowbits. Default is Checked.

This PR simply sets the default value for this checkbox to on when adding a new interface.

### DIFF
--- a/security/pfSense-pkg-snort/Makefile
+++ b/security/pfSense-pkg-snort/Makefile
@@ -2,6 +2,7 @@
 
 PORTNAME=	pfSense-pkg-snort
 PORTVERSION=	3.2.9.10
+PORTREVISION=	1	
 CATEGORIES=	security
 MASTER_SITES=	# empty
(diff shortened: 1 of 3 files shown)


rbgarga:
@bmeeks8 can you please take a look at this?

bmeeks8:
I'm okay with this change. It corrects an oversight error in the code. That value should default to "ON" with new interfaces.

PR #774, opened 2020-02-22
Fix FRR BGP redistribute config creation. Issue #8729

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/8729
Ready for review

Bug:
Selecting Yes to "Redistribute connected network" is not picking up those IPv6 networks.
If the networks are added manually to Networks to Distribute, everything works fine.

Looks like FRR assumes the globally declared option to be IPv4 only.

In bgpd.conf:
```
router bgp 65506
  redistribute connected
  address-family ipv4 unicast
   network 10.6.0.0/16
```

But when viewing the running configuration from vtysh:
```
router bgp 65506
 ...
 address-family ipv4 unicast
  network 10.6.0.0/16
  redistribute connected
 exit-address-family
```

This PR allow to select No/IPv4/IPv6/IPv4+IPv6 in the Redistribute drop-down menu (where 'IPv4+IPv6' is 'yes' for backward compatibility with previous configs)

It also fix creating 'address-family' entries in bgpd.conf, by doing this in more accurate way, i.e.
```
address-family ipv4 unicast
   redistribute connected
   redistribute ospf route-map testmap
   network 10.123.123.0/24
 exit-address-family
```

instead of:
```
address-family ipv4 unicast
   redistribute connected
exit-address-family
address-family ipv4 unicast
   redistribute ospf route-map testmap
exit-address-family
address-family ipv4 unicast
   network 10.123.123.0/24
exit-address-family
```


vktg:
rebased

PR #776, opened 2020-02-22
Squid listen on IPv6. Issue #8887

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/8887
Ready for review

original issue:

> How to reproduce:
> 1. Configure LAN interface to have IPv4 and IPv6
> 2. Install and configure Squid Proxy package
> 3. Go to /usr/local/etc/squid/squid.conf and there not exist:
> 4.1. For simple proxy:
> http_port [IPv6]:3128
> https_port [IPv6]:3129
> 4.2. If in addition transparent proxy configured:
> http_port [::1]:3128
> https_port [::1]:3129

This fix allows you to select the IP protocol over which Squid will listen for connections

Transparent mode is not supported in IPv6 mode, see:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=193568


vktg:
TODO:

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=193568#c2:

> Your PF rule redirects a packet to ::1, but doesn't change the receiving interface.  Thus, it violates scoping rules.  You can tell by running 'netstat -s -f inet6 | grep "violated scope"' before and after generating the traffic that you want to redirect.  The check is in in6_setscope().
> 
> The simple workaround is to change your rdr rule to redirect to your actual link-local, site-local, or global IPv6 address instead of ::1.

www/pfSense-pkg-squid/files/usr/local/pkg/squid.inc
```diff
@@ -788,6 +804,9 @@ function squid_validate_general($post, &$input_errors) {
 	unset($port, $webgui_port);
 
 	if ($post['transparent_proxy'] == 'on') {
+		if ($post['listenproto'] == 'inet6') {
+			$input_errors[] = "Transparent Proxy is not working for IPv6 only mode.";
```
jim-p: Reword this to `"Transparent Proxy does not support IPv6."`

www/pfSense-pkg-squid/files/usr/local/pkg/squid.xml
```diff
@@ -131,6 +131,19 @@
 			<type>checkbox</type>
 			<default_value>on</default_value>
 		</field>
+		<field>
+			<fielddescr>Listen IP Version</fielddescr>
+			<fieldname>listenproto</fieldname>
+			<description>Select listen IP version. This controls what addresses Squid listens for client connections.</description>
```
jim-p: Reword this to `Select the IP version Squid will use to select addresses for accepting client connections.`

www/pfSense-pkg-squid/files/usr/local/pkg/squid.inc
vktg: fixed

www/pfSense-pkg-squid/files/usr/local/pkg/squid.xml
vktg: fixed

vktg:
now it avoid to use ::1
I'll create a new issue for IPv6 transparent mode

JonathanDLee24:
> now it avoid to use ::1 I'll create a new issue for IPv6 transparent mode

Can you please list the other ticket for this? I am having this issues also with transparent mode, Squid Email support said I do need to use the [::] I am just testing out stuff with a HE tunnel in use. Everything works except if I enable IPV6 only and disable IPV4 it will not allow connections to the proxy from IPV6 local sources I get 409 errors for all of those connections.

PR #777, opened 2020-02-23
FRR Copy Running to Saved option for the raw config page. Issue #9003

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9003
Ready for review

> Would be possible to add these buttons to the raw config page as the Quagga package does, i'm slowly moving over to using FRR over Quagga for the new features being added but having to cat the /var/etc/frr/*.conf file and paste is into the UI for each daemon can be error prone.
> As an improvement over the quagga a 'copy all' button would be excellent.

I do not understand why quagga-way (saving configuration in base64 format in config.xml) may be error prone. Both *.conf and config.xml contains same data.
This is the port from Quagga with a 'copy all' button.


net/pfSense-pkg-frr/files/usr/local/pkg/frr/inc/frr_zebra.inc
```diff
@@ -620,6 +620,21 @@ function frr_zebra_generate_acceptfilters() {
...
+
+	$moduleRunningFile = "{$frr_config_base}/{$module}.conf";
+
+	if ( file_exists( $moduleRunningFile ) && ( filesize( $moduleRunningFile ) > 0 ) ) {
+		$moduleRunning = fopen( "$moduleRunningFile", "r" );
+		$config['installedpackages']['frrglobalraw']['config'][0][$module . "running"] = base64_encode( fread( $moduleRunning, filesize($moduleRunningFile) ));
+ 	} else {
+		$config['installedpackages']['frrglobalraw']['config'][0][$module . "running"] = base64_encode("!!!!! {$module}.conf does not exist or is empty.");
+	}
+	write_config( $desc = gettext("FRR: Wrote {$module}.conf startup-config to pfSense config file."), $backup = false, $write_config_only = true );
```
jim-p: Why is this setting variables by name here in the function call? Seems like it could be problematic. If this was copied directly over from quagga I suspect this may also be why on quagga I see lots of config writes which update the running config with a broken description of `(system): 1`.

This will need tested and maybe fixed both here and in quagga.

quadrinary: When I wrote that function for Quagga, the idea of the variable being part of the function call was to reduce/simplify the code, allowing the function to be used to write the config for any module.

vktg: fixed & rebased

vktg:
Now it uses 'Update RUNNING' button on the frr_global_raw.xml page to put base64-encoded configs to config.xml

PR #778, opened 2020-02-23
pfBlockerNG-devel, certificate options and :80 port on vip for DNSBL feature

PiBa-NL:
pfBlockerNG-devel, certificate options and :80 port on vip for DNSBL feature

- use a server-certificate with that purpose(x509_extensions) so Chrome doesn't complain about wrong cert usage (without option to skip..)
- use a random serial to avoid possible conflict with Firefox remembering the old CA-cert with same serial..
- Also listen on :80 for dnsBL.

(p.s. the old cert /var/unbound/dnsbl_cert.pem needs to be removed manually..)

### DIFF
--- a/net/pfSense-pkg-pfBlockerNG-devel/files/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/net/pfSense-pkg-pfBlockerNG-devel/files/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -1017,6 +1017,9 @@ accesslog.filename		= "|/usr/local/bin/php -f /usr/local/pkg/pfblockerng/pfblock
 		ssl.ec-curve	= "secp384r1"
 	}
 
+	\$SERVER["socket"] == "{$pfb['dnsbl_vip']}:80" {
+	}
+	
 	\$SERVER["socket"] == "{$pfb['dnsbl_vip']}:443" {
 		ssl.engine	= "enable"
 		ssl.pemfile	= "/var/unbound/dnsbl_cert.pem"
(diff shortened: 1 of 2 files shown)


net/pfSense-pkg-pfBlockerNG-devel/files/usr/local/pkg/pfblockerng/pfblockerng.inc
```diff
@@ -1055,7 +1058,8 @@ function pfb_create_dnsbl_cert() {
 
 	$pkey	= openssl_pkey_new($pkey_config);
 	$csr	= openssl_csr_new($dn, $pkey);
-	$cert	= openssl_csr_sign($csr, NULL, $pkey, 3650);
+	$cert	= openssl_csr_sign($csr, NULL, $pkey, 3650, array('digest_alg'=>'sha256',
+				'x509_extensions' => 'server'), cert_generate_serial());
```
jim-p: You probably also want to set `"req_extensions" => "req_server"` in this. Or, even better, use the functions we already have like `cert_create`/`csr_generate`/`csr_sign` and so on rather than using OpenSSL functions directly here.

Is there something you are doing which isn't supported by the built-in functions? If you need self-signed, you can still follow the example set by the GUI cert in `system.inc` inside `system_webgui_create_certificate()` and do something like this:

    cert_create($cert, null, 2048, 389, $dn, "self-signed", "sha256");

Note that you shouldn't use 3650 for server cert lifetimes anymore. The current max is 825 and this fall it will be lowered to 398. We've already updated 2.5.0 to recommend 389. See https://redmine.pfsense.org/issues/9825 for details.

PiBa-NL: Thanks, new commit pushed that uses cert_create as suggested.

rbgarga:
@PiBa-NL it still miss a bump on PORTVERSION/PORTREVISION to trigger a new build

PiBa-NL:
Currently discussing a bit about this PR with @BBcan177 .

BBcan177:
Looks good

rbgarga:
Anthony added these changes to another PR.  Thanks

PR #779, opened 2020-02-26
Propagate description field into QR code for FreeRADIUS. Issue #8878

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/8878
Ready for review

it is often desirable (esp. when you have more than one identity added in your Google Auth mobile app) to distinguish between different identities. Having all of them to be descripted as "*FreeRADIUS*" is not the most convenient way.

original code by Juraj Lutter (on the redmine page)


jurajlutter:
Thanks.

jurajlutter:
Thanks.

PR #780, opened 2020-02-26
SIP Trunk plugin for siproxd. Issue #9249

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9249
Ready for review

from siproxd.conf.example:

> Plugin_siptrunk
> 
>  Plugin to handle SIP Trunks where using *one* single SIP account a
>  whole number block is routed. This means asn incoming INVITE does carry
>  the target number (in SIP URI or To: header field) but does not really
>  carry any clear indications to which account it belongs to.
>  Thus, we need some help - a mapping of the number blocks used in a SIP
>  trunk and the corresponding SIP account (as used during REGISTER)
> 
>  ..._name:          any name, pure documentation
>  ..._account:       SIP account in the form of 'sip:user@host',
>                     identical as used for registration.
>  ..._numbers_regex: REGEX that matches the whole number block associated
>                     with this account
> 
>  NOTE: plugin_siptrunk is mutually exclusive with plugin_defaulttarget!
>        if the defaulttartet 'plugin is also active, then all incoming calls 
>        on SIP trunks will be redirected to the default target.
> 
> plugin_siptrunk_name = Example Trunk, 555-123100 ... 555-123112
> plugin_siptrunk_account = sip:user@sip.example.org
> plugin_siptrunk_numbers_regex = ^555123(10[0-9]|11[012])$

This PR adds siptrunk configuration to WebGUI

### DIFF
--- a/net/pfSense-pkg-siproxd/Makefile
+++ b/net/pfSense-pkg-siproxd/Makefile
@@ -1,8 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-siproxd
-PORTVERSION=	1.1.3
-PORTREVISION=	3
+PORTVERSION=	1.1.4
 CATEGORIES=	net
 MASTER_SITES=	# empty
 DISTFILES=	# empty
--- a/net/pfSense-pkg-siproxd/files/usr/local/pkg/siproxd.xml
+++ b/net/pfSense-pkg-siproxd/files/usr/local/pkg/siproxd.xml
@@ -517,6 +517,29 @@
 			<default_value>192.168.0.0/16,10.0.0.0/8,172.16.0.0/20</default_value>
 		</field>
 		<field>
+			<name>Plugin Settings - SIP Trunk</name>
+			<type>listtopic</type>
+		</field>
+		<field>
+			<fielddescr>SIP Trunk</fielddescr>
+			<fieldname>plugin_siptrunk</fieldname>
+			<description>Enable SIP Trunk Plugin</description>
+			<sethelp>Plugin to handle SIP Trunks where using *one* single SIP account to which a whole number block is routed. This means an incoming INVITE does carry the target number (in SIP URI or To: header field) but does not really carry any clear indications to which account it belongs to.</sethelp>
+			<type>checkbox</type>
+		</field>
+		<field>
+			<fielddescr>SIP account</fielddescr>
+			<fieldname>plugin_siptrunk_account</fieldname>
+			<description>SIP account in the form of 'sip:user@host', identical as used for registration.</description>
+			<type>input</type>
+		</field>
+		<field>
+			<fielddescr>REGEX number</fielddescr>
+			<fieldname>plugin_siptrunk_numbers_regex</fieldname>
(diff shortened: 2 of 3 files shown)


net/pfSense-pkg-siproxd/files/usr/local/pkg/siproxd.xml
```diff
@@ -516,6 +516,29 @@
...
 		</field>
+		<field>
+			<name>Plugin Settings - SIP Trunk</name>
+			<type>listtopic</type>
+		</field>
+		<field>
+			<fielddescr>SIP Trunk</fielddescr>
+			<fieldname>plugin_siptrunk</fieldname>
+			<description>Enable SIP Trunk Plugin</description>
+			<sethelp>Plugin to handle SIP Trunks where using *one* single SIP account a whole number block is routed. This means asn incoming INVITE does carry the target number (in SIP URI or To: header field) but does not really carry any clear indications to which account it belongs to.</sethelp>
```
jim-p: Seems to be some missing words and typos in this line. Like `asn` instead of `an`, and around `account a whole number block` is missing some words like `to which a whole number block is routed` or `where a whole number block is routed`.

net/pfSense-pkg-siproxd/files/usr/local/pkg/siproxd.xml
```diff
@@ -516,6 +516,29 @@
...
+		<field>
+			<fielddescr>SIP account</fielddescr>
+			<fieldname>plugin_siptrunk_account</fieldname>
+			<description>SIP account in the form of 'sip:user@host', identical as used for registration.</description>
+			<type>input</type>
+		</field>
+		<field>
+			<fielddescr>REGEX number</fielddescr>
+			<fieldname>plugin_siptrunk_numbers_regex</fieldname>
+			<description>Regular expression that matches the whole number block associated with this account</description>
```
jim-p: Would be nice to have an example here to show the expected format of the regex (since parsers vary)

net/pfSense-pkg-siproxd/Makefile
```diff
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-siproxd
 PORTVERSION=	1.1.3
-PORTREVISION=	3
+PORTREVISION=	4
```
jim-p: This is a large enough change to warrant bumping `PORTVERSION` (which also means removing `PORTREVISION`)

net/pfSense-pkg-siproxd/files/usr/local/pkg/siproxd.xml
vktg: fixed

net/pfSense-pkg-siproxd/files/usr/local/pkg/siproxd.xml
vktg: done

net/pfSense-pkg-siproxd/Makefile
vktg: done

PR #376, opened 2017-07-16
pfSense-pkg-LCDproc: add a shutdown/reboot control menu

c-schwamborn:
add a control menu with the name 'pfsense' next to the 'Options' menu of LCDproc.
Currently there are only two options implemented: Reboot and shutdown, but I think more usefull direct accessible features could be made avalable (i.e. switch wlan on/off for example)
Current structure:
```
Options (original LCDproc)
pfsense
  ->Reboot
    ->No
    ->Yes
  ->Shutdown
    ->No
    ->Yes
```

PR #784, opened 2020-02-28
LCDproc: add a shutdown/reboot control menu. Issue #8196

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/8196
Ready for review

add a control menu with the name 'pfsense' next to the 'Options' menu of LCDproc.
Currently there are only two options implemented: Reboot and shutdown.
Structure:
```
Options (original LCDproc)
pfsense
  ->Reboot
    ->No
    ->Yes
  ->Shutdown
    ->No
    ->Yes
```

This is fixed/resolved copy of the original PR https://github.com/pfsense/FreeBSD-ports/pull/376 by @c-schwamborn


vktg:
Why we need to use printf() instead of '<br />' as everywhere in this code https://github.com/pfsense/FreeBSD-ports/pull/376/files#r130587889 ?

rbgarga:
> Why we need to use printf() instead of `<br />` as everywhere in this code https://github.com/pfsense/FreeBSD-ports/pull/376/files#r130587889 ?

setHelt() is one of the functions considered by our gettext config to be translated and we don't want to have HTML tags mixed with strings for translation.  setHelp accepts printf() similar parameters so you can replace <br /> by `%1$s` and pass `'<br />'` as a parameter

vktg:
> > Why we need to use printf() instead of `<br />` as everywhere in this code https://github.com/pfsense/FreeBSD-ports/pull/376/files#r130587889 ?
> 
> setHelt() is one of the functions considered by our gettext config to be translated and we don't want to have HTML tags mixed with strings for translation. setHelp accepts printf() similar parameters so you can replace
> by `%1$s` and pass `'<br />'` as a parameter

ok, fixed in all lcdproc_client.php

sysutils/pfSense-pkg-LCDproc/files/usr/local/pkg/lcdproc_client.php
```diff
@@ -425,6 +440,14 @@ function outputled_enabled_CFontz633() {
 	}
 }
 
+function cmenu_enabled() {
+	global $config;
+	$lcdproc_config = $config['installedpackages']['lcdproc']['config'][0];
+	if(!isset($lcdproc_config['controlmenu']))
+		return false;
```
jim-p: `if` body needs braces

vktg: fixed

sysutils/pfSense-pkg-LCDproc/files/usr/local/pkg/lcdproc_client.php
```diff
@@ -425,6 +440,15 @@ function outputled_enabled_CFontz633() {
 	}
 }
 
+function cmenu_enabled() {
+	global $config;
+	$lcdproc_config = $config['installedpackages']['lcdproc']['config'][0];
+	if(!isset($lcdproc_config['controlmenu'])) {
```
jim-p: Needs a space between `if` and `(`

vktg: fixed

PR #789, opened 2020-03-05
Send notifications to all platforms.

Kapmeister:
At the moment, the service watchdog will only notify via email. 
Pfsense has the ability to notify via growl as well (although I read it will be removed). 
This change calls the subroutine that will call all notification methods, not just email.


jim-p:
Since growl is out, I don't see the advantage of changing this at this time. Maybe in the future if other methods are added.

Kapmeister:
> 
> 
> Since growl is out, I don't see the advantage of changing this at this time. Maybe in the future if other methods are added.

Hi Jim, 
I've got push notifications to Telegram working successfully on 2 installs. I've obviously edited one line in this file in order to get them to work properly. I'm in the process of learning github so that if / when I think the Telegram modification is ready for production I'm familiar with the process. 

regards

John


Telegram

PR #791, opened 2020-03-09
Allow multiple IP in ListenIP for Zabbix Agent. Issue #9272

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9272
Ready for review

This PR allow to enter multiple comma-separated Listen IPs,
And also allow to use :: and ::/1 IPv6 addresses

### DIFF
--- a/net-mgmt/pfSense-pkg-zabbix-agent/files/usr/local/pkg/zabbix-agent.inc
+++ b/net-mgmt/pfSense-pkg-zabbix-agent/files/usr/local/pkg/zabbix-agent.inc
@@ -46,24 +46,45 @@ function php_deinstall_zabbix_agent() {
 function validate_input_zabbix_agent($post, &$input_errors) {
 
 	if (isset($post['agentenabled'])) {
-		if (!preg_match("/\w+/", $post['server'])) {
+		if ($post['server'] != '') {
+			foreach (explode(",", $post['server']) as $srv) {
+				$srv = trim($srv);
+				if (!is_ipaddr($srv) && !is_hostname($srv)) {
+					$input_errors[] = htmlspecialchars($srv) .
+					" is not a valid IP address or hostname for 'Server'.";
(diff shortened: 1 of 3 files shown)


net-mgmt/pfSense-pkg-zabbix-agent/files/usr/local/pkg/zabbix-agent.inc
```diff
@@ -55,15 +55,17 @@ function validate_input_zabbix_agent($post, &$input_errors) {
 		}
 
 		if ($post['listenip'] != '') {
-			if (!is_ipaddr_configured($post['listenip']) && !preg_match("/(127.0.0.1|0.0.0.0)/", $post['listenip'])) {
-				$input_errors[] = "'Listen IP' is not a configured IP address.";
+			$listenips = explode(",", $post['listenip']);	
+			foreach ($listenips as $listip) {
+				$listip = trim($listip);
+				if (!is_ipaddr_configured($listip) && !preg_match("/(127.0.0.1|0.0.0.0|::1|::)/", $listip)) {
+					$input_errors[] = "'Listen IP' is not a configured list of IP addresses.";
```
jim-p: Doing the test this way will repeat the same error multiple times if more than one entry fails the test. It should probably  Set a flag variable then break out of the loop on the first error, then print the error once.

Alternately, it could print the failed value in the message and print an error for each failed value. Encode the value in output for safety (e.g. `htmlspecialchars()`).

vktg: + printing error for each field
+ extra validation checks

rbgarga:
@dbaio can you please review this change?

dbaio:
> @dbaio can you please review this change?

I've tested it here and looks good, thanks

PR #798, opened 2020-03-11
FRR OSPFv3 router-id fix. Issue #10338

bmhughes:
## FRR

- The current package configures the ospf6 router-id as `router-id A.B.C.D`, it should be `ospf6 router-id A.B.C.D`.

[Issue #10338](https://redmine.pfsense.org/issues/10338)


bmhughes:
> 
> 
> 
> See https://www.netgate.com/docs/pfsense/development/developer-style-guide.html#ports-packages-specific-rules

Done

PR #810, opened 2020-03-28
FreeRADIUS user IPv6 Prefix fix. Issue #10297

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/10297
Ready for review

https://forum.netgate.com/topic/151725/freeradius-ipv6-framed-ip-issue-with-2-4-5:
> Thought I'd have a play with framed IPv6 now im running 2.4.5.
> 
> /usr/local/etc/raddb/mods-config/files/authorize[31]: Parse error (reply) for entry andy-iphone: Failed resolving "64" to IPv6 address: hostname nor servname provided, or not known

Correct format is Prefix/Prefix-length, i.e. 2001:db8:abab::/64

See https://tools.ietf.org/html/rfc3162#section-2.3


vktg:
Done

also includes a fix for https://redmine.pfsense.org/issues/10447:
> Description
> 
> if you use Framed-IP-Address with plus sign, i.e.
> 
> Framed-IP-Address = 192.0.2.32+
> there is an error:
> 
> Error: /usr/local/etc/raddb/mods-config/files/authorize[159]: Parse error (reply) for entry DEFAULT: 
> Failed resolving "192.0.2.32+" to IPv4 address: Name or service not known
>
> The plus sign is no longer supported in version 3.0.x, see:
> https://github.com/FreeRADIUS/freeradius-server/issues/3243

vktg:
done
now it uses `varusersframedip6address` value for Framed-IP-Address/Framed-IP-Prefix

PR #815, opened 2020-04-01
ACME package version 0.6.6

danielztolnai:
I've updated the ACME package to include the latest acme.sh version. I needed the CloudFlare Zone ID authentication, but I've included and added support for every other change as well.

### DIFF
--- a/security/pfSense-pkg-acme/Makefile
+++ b/security/pfSense-pkg-acme/Makefile
@@ -1,7 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-acme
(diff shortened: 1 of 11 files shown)


jim-p:
I have a special repository that I use for this. Thanks, but I'll handle it internally.

danielztolnai:
Sorry, I did not know about this. Could you please update the package so we have this new feature in production?

jim-p:
I had a couple other bugs that popped up which needed fixing. I updated everything just now.

https://forum.netgate.com/topic/151994/acme-package-version-0-6-6

PR #822, opened 2020-04-03
Implements #10415 Adds prehashed NT-Password and MD5-Password to FreeRadius config

twlee79:
Implements [Feature #10415](https://redmine.pfsense.org/issues/10415).

These changes expose the option to use 'NT-Password' in FreeRADIUS. This provides the user an option to use hashed passwords with more Authentication protocols including the commonly used EAP-MSCHAPv2. Currently FreeRADIUS in pfSense only exposes the option to use 'MD5-Password' (MD5 hashes) which has very limited protocol support. EAP-MSCHAPv2 users cannot use MD5-Password and must use 'Cleartext-Password' which exposes passwords to casual observation.

This feature is implemented with very few changes to the codebase, as I chose to add this option with prehashed NT Passwords, i.e. a user calculates the hash themselves (easily done with freely available tools) and enters it in the configuration. Thus this feature simply adds the "NT-Password (pre-hashed)" option to the 'Password Encryption' field in the FreeRADIUS users configuration page.

I also added a corresponding "MD5-Password (pre-hashed)" option to the 'Password Encryption' field.

Correct function of this code and be checked by examining the 'users' radius configuration file under 'View config', it should store the user/password as: "USERNAME" NT-Password := "NTHASHEDPASSWORD" or "USERNAME" MD5-Password := "MD5HASHEDPASSWORD" 

I have tested use of this feature to allow hashed NT Passwords in my own pfSense firewall with EAP-MSCHAPv2 and this has been running without issue since November 2019.


twlee79:
Done. I've incremented the PORTREVISION in the Makefile as this is a minor change.

PR #830, opened 2020-04-04
Squid IPv6 transparent mode. Issue #10335

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/10335
Ready for review

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=193568#c2:
> Your PF rule redirects a packet to ::1, but doesn't change the receiving interface. Thus, it violates scoping rules. You can tell by running 'netstat -s -f inet6 | grep "violated scope"' before and after generating the traffic that you want to redirect. The check is in in6_setscope().
> 
> The simple workaround is to change your rdr rule to redirect to your actual link-local, site-local, or global IPv6 address instead of ::1.

In transparent IPv6 mode, redirection to the IPv6 address of the interface is used.


vktg:
Other fixes:
- fixes IPv6 duplicate addresses in localnet ACL
- allows to use IPv6 addresses in ACL
- extra input validations

PR #833, opened 2020-04-06
Feature #10428

kiokoman:
Feature #10428
https://redmine.pfsense.org/issues/10428


stephenw10:
I have tested this and it looks good. Conf file is generated as expected. LCD/keypad functions.

PR #835, opened 2020-04-07
Integration of bfd daemon (Ticket #10441)

dsiunicaen:
https://redmine.pfsense.org/issues/10441


vktg:
Works as expected

Tested on pfSense 2.5.0.a.20200601.1136 with frr pkg 0.6.5 +
Cisco IOS 15.5(3)M

OSPF and BGP works fine with BFD

PR #837, opened 2020-04-10
ACME IDN description. Issue #10442

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/10442
Ready for review

this fix uses `descr` field name instead of `desc`,
it's included in the `$cdata_fields` of xmlparser.inc

### DIFF
--- a/security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc
+++ b/security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc
@@ -1202,9 +1202,26 @@ function acme_custom_php_install_command() {
 	/* Remove old cron job (if present), see #9279 */
 	install_cron_job('/usr/local/pkg/acme/acme_command.sh "renewall"', false);
 	set_cronjob();
+	/* convert <desc> to <descr> to allow IDN characters, #10442 */
+	acme_convert_desc_to_descr();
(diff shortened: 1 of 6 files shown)


vktg:
done
now it converts <desc> to <descr> in `acme_custom_php_install_command()`

security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc
```diff
@@ -1202,11 +1202,33 @@ function acme_custom_php_install_command() {
...
+		foreach ($config['installedpackages']['acme'][$a_array]['item'] as & $item) {
+			if (isset($item['desc'])) {
+				$item['descr'] = $item['desc'];
+				unset($item['desc']);
+				$writecf = true;
+			}
+		}
+	}
+	if ($writecf) {
+		write_config("Converting ACME desc to descr to allow IDN characters in description");
```
jim-p, line 1228: Does it actually need to do its own `write_config()` here? Usually the pkg system would do a `write_config()` of its own after running the installer code.

vktg: hm,didn't think about it.
fixed!

vktg:
can be merged with https://github.com/pfsense/FreeBSD-ports/pull/847

PR #839, opened 2020-04-13
Squid reverse proxy switching peers fix. Issue 10450

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/10450
Ready for review

https://forum.netgate.com/topic/118255/squid-reverse-proxy-switching-peers:
> I want Squid to direct incoming requests to a specific peer based on URI, but what I'm getting instead is it's sending the requests to each peer in a round-robin fashion. So, once in a while, I get the content, or sometimes I get a 404 when it's hitting the wrong peer. I only have one peer selected per mapping.
> ...
> I have a working solution that might have to be re-implemented if and when I update the squid package. The trick is to add cache_peer_access directives that specifically deny the peers from the other mappings. For some reason, the "deny allsrc" that is automatically generated is not cutting it.


vktg:
it also adds extra input validation for Squid reverse proxy pages 
to avoid 'Squid Reverse Proxy alternating between destinations' bug:
https://redmine.pfsense.org/issues/7797
caused by space symbols in Peer Alias

PR #840, opened 2020-04-15
Squid Reverse Proxy dstdom_regex support. Issue #9762

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9762
Ready for review

This PR allow to use "dstdom_rexex" for domain redirection.
This makes the redirects much more flexible in general.
Also makes it possible to redirect a subdomain to a different mapping than the domain itself.
This is a more consistent approach as mappings and redirects are handled in a similar manner using regular expressions all the way.


StarkJohan:
Nice addition, should solve any possible issues with backward compatibility.

PR #842, opened 2020-04-16
OpenVPN export legacy remote proto. Issue #10369

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/10369
Ready for review

OpenVPN < 2.4 doesn't support remote IPv4/IPv6 protocol definition (udp4/udp6/tcp4/tcp6),
If checkbox **Legacy Client** is set, it must generate compatible config (udp/tcp).

see https://forum.netgate.com/topic/152577/pfsense-openvpn-client-export-problem

### DIFF
--- a/security/pfSense-pkg-openvpn-client-export/files/usr/local/pkg/openvpn-client-export.inc
+++ b/security/pfSense-pkg-openvpn-client-export/files/usr/local/pkg/openvpn-client-export.inc
@@ -181,7 +181,7 @@ EOF;
 	}
 	
 	// determine basic variables
-	$remotes = openvpn_client_export_build_remote_lines($settings, $useaddr, $interface, $expformat, $nl);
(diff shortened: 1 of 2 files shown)


security/pfSense-pkg-openvpn-client-export/files/usr/local/pkg/openvpn-client-export.inc
```diff
@@ -944,7 +944,7 @@ function openvpn_client_export_sharedkey_config($srvid, $useaddr, $proxy, $nokey
 	}
 }
 
-function openvpn_client_export_build_remote_lines($settings, $useaddr, $interface, $expformat, $nl) {
+function openvpn_client_export_build_remote_lines($settings, $useaddr, $interface, $expformat, $nl, $legacy) {
```
jim-p, line 947: Would be best to make `$legacy` default to `false` since it's being added to an existing function.

vktg: ok, fixed

PR #843, opened 2020-04-18
ARPwatch: Suppress notifications for specific MACs. Issue #10474

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/10474
Ready for review

Allow to suppress specific MACs from ARPwatch notifications

### DIFF
--- a/net-mgmt/pfSense-pkg-arpwatch/files/usr/local/pkg/arpwatch.inc
+++ b/net-mgmt/pfSense-pkg-arpwatch/files/usr/local/pkg/arpwatch.inc
@@ -111,6 +111,12 @@ function arpwatch_validate_input($post, &$input_errors) {
 	if (filter_var($post['notifications_recipient'], FILTER_VALIDATE_EMAIL) === false) {
 		$input_errors[] = gettext("{$post['notifications_recipient']} is not a valid email address.");
 	}
+
+	for ($i = 0; $i < 30; $i++) {
+		if (!empty($post["mac{$i}"]) && !is_macaddr($post["mac{$i}"])) {
+			$input_errors[] = "{$post["mac{$i}"]} is not valid MAC address.";
+		}
+	}
 }
 
 function arpwatch_resync_config_command() {
@@ -215,6 +221,7 @@ function arpwatch_install_sendmail_proxy() {
 #!/usr/bin/env php
 <?php
 
+global $config;
 require_once("notices.inc");
 
 $fd = fopen('php://stdin','r');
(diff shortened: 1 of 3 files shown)


net-mgmt/pfSense-pkg-arpwatch/files/usr/local/pkg/arpwatch.xml
```diff
@@ -117,6 +117,20 @@
 			<type>checkbox</type>
 			<description>Reset the database of collected mac/ip addresses before starting Arpwatch.</description>
 		</field>
+		<field>
+			<fielddescr>Suppress MAC</fielddescr>
+			<fieldname>none</fieldname>
+			<type>rowhelper</type>
+			<rowhelper>
+				<rowhelperfield>
```
dbwiddis, line 125: It would be useful to only suppress certain types of notifications per-MAC.  For example, I might want to suppress "flip flop" but keep "new activity", "new station", and "changed ethernet address".

dbwiddis: I've implemented and tested my suggested improvement and submitted a PR to your fork.

vktg: Thank you!
I did some code cleanup and switched mac&type fields (type is more optional)

dbwiddis: Great!  Looks good. I'm not that experienced with php so I did what I knew would work.  I thought of doing multi-select on the options rather than "all" but decided the added complexity probably wasn't worth the effort.  Really appreciate you noting my report on the forum and getting this started!

yuriw:
@vktg This is great, thx !  Pls merge ASAP

net-mgmt/pfSense-pkg-arpwatch/files/usr/local/pkg/arpwatch.inc
```diff
@@ -224,6 +230,16 @@ fclose($fd);
 if (false !== $message) {
 	$subject = array();
 	preg_match('/^Subject: (.*)$/m', $message, $subject);
+	if (is_array($config['installedpackages']['arpwatch']['config'][0]['row'])) {
```
jim-p, line 233: Rather than doing `is_array()`, You should just run that whole bit through `init_config_arr()`. Otherwise if any of the intermediate parts are missing/unset, this will result in a PHP error.

Also the code being generated that this is inside of needs a `global $config;` line.

vktg: Done, fixed

vktg: + added `global $config;`

PR #844, opened 2020-04-18
net-mgmt/pfSense-pkg-arpwatch: Never clobber database on install

leres:
and only clobber on deinstall if clear_database is enabled.

### DIFF
--- a/net-mgmt/pfSense-pkg-arpwatch/Makefile
+++ b/net-mgmt/pfSense-pkg-arpwatch/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-arpwatch
 PORTVERSION=	0.2.0
-PORTREVISION=	1
+PORTREVISION=	2
 CATEGORIES=	net-mgmt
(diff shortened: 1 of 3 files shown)


dbwiddis:
Love this fix! An alternative/other option might be to pre-fill/reset the database from the DHCP Server static IP table ($config['dhcpd']['staticmap']).

leres:
> Love this fix! An alternative/other option might be to pre-fill/reset the database from the DHCP Server static IP table ($config['dhcpd']['staticmap']).

I prefer that my arpwatch database only contain entries for mac addresses actually seen on the network. Also I have entries in that are more than 10 years old, I don't see any reason to ever remove these.

(Fun fact: I started writing arpwatch in November, 1990, almost 30 years ago!)

dbwiddis:
Fair enough. Those of us just now setting up with 86 devices we laboriously moved over from our old router don't have 10 years of incrementally adding them.  But 86 one time "new station" alerts are not that bad, especially if we can keep our old DB!

PR #860, opened 2020-05-14
LLDPd integrated switch workaround. Issue #9635

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/9635
Ready for review

Add a workaround for appliances with integrated switches to allow run LLDPd on it,
as by default LLDPd uses parent physical interface to run, which doesn't work with integrated switches

### DIFF
--- a/net-mgmt/pfSense-pkg-lldpd/files/usr/local/pkg/lldpd/lldpd.inc
+++ b/net-mgmt/pfSense-pkg-lldpd/files/usr/local/pkg/lldpd/lldpd.inc
@@ -120,11 +120,24 @@ function lldpd_sync_config() {
 	
 	/* Interface list */
 	$interfaces = array();
-	foreach (explode(',', $lldpd_config['interfaces']) as $i) {
-		foreach (get_parent_physical_interface($i) as $p) {
-			$interfaces[$p] = $p;
+	$int_switch = array();
+	if (file_exists("/etc/inc/switch.inc")) {
+		require_once('switch.inc');
+		if (!empty(switch_get_devices())) {
+			$int_switch = switch_get_uplink_ports();
 		}
 	}
+	foreach (explode(',', $lldpd_config['interfaces']) as $i) {
+		foreach (get_parent_physical_interface($i) as $p) {
+			if (!empty($int_switch) && in_array($p, $int_switch)) {
+				$int = get_real_interface($i);
+				$interfaces[$int] = $int;
+			} else {
+				$interfaces[$p] = $p;
+			}
+		}
+	}
+	$interfaces = array_unique($interfaces);
 	if (count($interfaces)) {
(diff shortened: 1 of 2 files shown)


net-mgmt/pfSense-pkg-lldpd/files/usr/local/pkg/lldpd/lldpd.inc
```diff
@@ -120,11 +120,21 @@ function lldpd_sync_config() {
 	
 	/* Interface list */
 	$interfaces = array();
+	$platform = system_identify_specific_platform();
+	$int_switch = array('SG-1100', 'SG-3100');
```
jim-p, line 124: There are more platforms that just these with integrated switches. You should check for the presence of the function `switch_get_devices()`, if it's found, and has a non-empty result, then you know the device has a switch.

Might need to also check for the presence of `/etc/inc/switch.inc` and `require_once('switch.inc');` before that function is available, but I think it gets included early enough in other files that may not be necessary.

dragoangel:
I think this not enough check. As I was written in bug report: people can use their own Laggs/bonding. As I use lagg on ix0-1 on XG7100 for spf+. There 2 ways to fix thix this:
1. Relly on pfsense administrator: provide him list of interfaces that lldp will use, so he can  modify it to his needs.
2. Figure out what interaces are part of bonding or lagg and exclude them from lldp

net-mgmt/pfSense-pkg-lldpd/files/usr/local/pkg/lldpd/lldpd.inc
vktg: got it! fixed

tested on SG-1100, XG-7100 and KVM

vktg:
> I think this check isn't correct. As I was written in bug report: people can use their own Laggs/bonding. As I use lagg on ix0-1 on XG7100 for spf+. There 2 ways to fix this:
ix0-1 is not an integrated switch, do you have issues running LLDPd on it?

> 2. Figure out what interaces are part of bonding or lagg and exclude them from lldp
in this case, it will not allow you to run LLDPd on the XG-7100 ETH2-ETH8 ports or run LLDPd on any SG-1100 ports

dragoangel:
Sorry for late reply,
Yep, I have lagg on SPF+ ix0 & ix1 for reductant connection.
Sorry for unclear second option, I mean exclude parent, not child interfaces. Doesn't understand why they selected in end as in lldp GUI I selected LAN(lagg1.3080), MGMNT(lagg1.3081) and so on. They are vlans on lagg1 which has child ix0 & ix1.

vktg:
Please report it on the redmine issue page

dragoangel:
You speak about p.s. offtopic or all comment in general?

dragoangel:
Additionally maybe even lldpd on VLANs will generate such errors but I can't check this as all my pfsense servers have VLANs on LAGGs. Could you please check it?

PR #861, opened 2020-05-15
Build HAProxy Package with buildin Prometheus exporter. Implement #10500

vktg:
Redmine Issue: https://redmine.pfsense.org/issues/10500
Ready for review

However, PROMEX is not built by default with HAProxy. It is provided as an extra
component for everyone want to use it. So you need to explicitly build HAProxy
with the PROMEX service, using the Makefile variable "EXTRA_OBJS". For instance:
`    make TARGET=linux-glibc EXTRA_OBJS="contrib/prometheus-exporter/service-prometheus.o"`

if HAProxy provides the PROMEX service, the following build option will be
reported by the command "haproxy -vv":

>     Built with the Prometheus exporter as a service

https://github.com/haproxy/haproxy/blob/master/contrib/prometheus-exporter/README

needs testing on 2.5 first

### DIFF
--- a/net/pfSense-pkg-haproxy-devel/Makefile
+++ b/net/pfSense-pkg-haproxy-devel/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-haproxy-devel
 PORTVERSION=	0.60
-PORTREVISION=	4
+PORTREVISION=	5
 CATEGORIES=	net
 MASTER_SITES=	# empty
 DISTFILES=	# empty


vktg:
> Seems to me this should be added as a configurable option for the FreeBSD port and that should be submitted upstream to FreeBSD first.

it is already there
https://github.com/pfsense/FreeBSD-ports/blob/a198c1a7b6294bf74f90f93b074849acd878dd74/net/haproxy/Makefile#L39

PR updated

net/haproxy/Makefile
```diff
@@ -28,7 +28,7 @@ DPCRE_DESC=		Link dynamically
 SPCRE_DESC=		Link statically
 DEVICEATLAS_DESC=	DeviceAtlas Device Detection support
 PROMEX_DESC=		Enable Prometheus exporter
-OPTIONS_DEFAULT=	SPCRE OPENSSL
+OPTIONS_DEFAULT=	SPCRE OPENSSL PROMEX
```
jim-p, line 31: Don't change the default options. Add/change the relevant lines in the pfSense repository in https://github.com/pfsense/pfsense/blob/master/tools/conf/pfPorts/make.conf

vktg: done: https://github.com/pfsense/pfsense/pull/4320

PR #878, opened 2020-06-12
Snort VPN pass list improvements. Fixes #8688

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/8688
- [X] Ready for review

`filter_get_vpns_list()` returns only:
- IPsec Mobile IPv4 subnet
- IPsec site-to-site networks
- OpenVPN client/server Tunnel Network / Remote Network IPv4
- PPPoE server networks

but not:
- IPsec Mobile IPv6 subnet
- IPsec Mobile warriors IPs (VPN / IPsec / Pre-Shared Keys / Edit)
- OpenVPN client/server Tunnel Network / Remote Network IPv6
- L2TP VPN network

new `snort_get_vpns_list()` returns all

TODO: same for snort 3.x and suricata

### DIFF
--- a/security/pfSense-pkg-snort/Makefile
+++ b/security/pfSense-pkg-snort/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-snort
 PORTVERSION=	4.1
-PORTREVISION=	1
+PORTREVISION=	2
 CATEGORIES=	security
 MASTER_SITES=	# empty
 DISTFILES=	# empty
--- a/security/pfSense-pkg-snort/files/usr/local/pkg/snort/snort.inc
+++ b/security/pfSense-pkg-snort/files/usr/local/pkg/snort/snort.inc
@@ -472,7 +472,7 @@ function snort_build_list($snortcfg, $listname = "", $whitelist = false, $extern
 
 	// Grab a list of vpns enabled - these come back as CIDR mask networks
 	if ($vpns == 'yes') {
-		$vpns_list = filter_get_vpns_list();
+		$vpns_list = snort_get_vpns_list();
 		if (!empty($vpns_list)) {
 			/* Convert the returned space-delimited string to an array */
 			/* and then add each VPN address to our HOME_NET array.    */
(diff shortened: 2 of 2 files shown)


bmeeks8:
I'm fine with this change, but I have a couple of comments to be considered. 

1. Suricata will be impacted as well since it uses the exact same system call to get VPNs. So we need to port this fix over to Suricata, too.

2. Should we consider modifying the pfSense system call to return all the VPN addresses, or will that potentially break other things unrelated to Snort and Suricata?

vktg:
> 2\. Should we consider modifying the pfSense system call to return all the VPN addresses, or will that potentially break other things unrelated to Snort and Suricata?

This is not so trivial, see https://redmine.pfsense.org/issues/10493
This may be changed in the future, but not so fast.

rbgarga:
@vktg today snort package differs a lot between 2.4.5 and 2.5.0, so I can't cherry-pick it to 2.4.5.  If this fix is important a new PR should be submitted against RELENG_2_4_5 branch.

@bmeeks8 just confirm it's approved and I will go ahead and merge it

bmeeks8:
> @vktg today snort package differs a lot between 2.4.5 and 2.5.0, so I can't cherry-pick it to 2.4.5. If this fix is important a new PR should be submitted against RELENG_2_4_5 branch.
> 
> @bmeeks8 just confirm it's approved and I will go ahead and merge it

Yes, I am fine with this change and appreciate the code fix from Viktor. Renato is correct that there are some substantial changes between Snort 3.x and Snort 4.x due to the inclusion of an inline IPS mode with netmap in the Snort 4.x branch for pfSense-2.5. I agree that the approach for the Snort 3.x package on pfSense-2.4.5 is to submit based on the RELENG_2_4_5 branch.

This change also needs to get pulled into the Suricata package as it currently uses the same pfSense system call for obtaining VPN addresses and thus will be missing the addresses not returned by the current system call. I will be happy to do that by copying Viktor's code over. This particular function call is almost identicall in both Snort and Suricata, so porting over should be easy.

PR #880, opened 2020-06-12
Squid ACL extra check. Fixes #10654

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10654
- [X] Ready for review

`idn_to_ascii()` returns false if domain name started with dot (`.example.com`)
avoid it, and any other non-domain strings (regexp)


jim-p:
Correct Redmine link is: https://redmine.pfsense.org/issues/10654

vktg:
> Correct Redmine link is: https://redmine.pfsense.org/issues/10654

fixed

PR #881, opened 2020-06-12
ACME key type change on renew. Fixes #10656

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10656
- [X] Ready for review

Fix for issue:
> changing the private key configuration item from ECDSDA to RSA appears to succeed, but when re-issuing or renewing the key it silently continues to use the old private key type. At best this is very confusing when the configuration now shows one type of private key, but they system is actually receiving and using a previously configured type.

### DIFF
--- a/security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc
+++ b/security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc
@@ -1370,10 +1370,31 @@ function & get_certificate($name) {
 			echo "\n{$cert['prv']}";
 			$desc = "Acme: Add new certificate & key.";
 			write_config($desc);
-		} elseif (   $certificate['keylength'] == 'custom'
-				  && $cert['prv'] != $certificate['keypaste']) {
+		} elseif (($certificate['keylength'] == 'custom') && ($cert['prv'] != $certificate['keypaste'])) {
 			echo "\n getCertificatePSK updating custom key";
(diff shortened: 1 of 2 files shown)


security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc
```diff
@@ -1370,10 +1370,13 @@ function & get_certificate($name) {
 			echo "\n{$cert['prv']}";
 			$desc = "Acme: Add new certificate & key.";
 			write_config($desc);
-		} elseif (   $certificate['keylength'] == 'custom'
-				  && $cert['prv'] != $certificate['keypaste']) {
+		} elseif (($certificate['keylength'] == 'custom') && ($cert['prv'] != $certificate['keypaste'])) {
 			echo "\n getCertificatePSK updating custom key";
 			$cert['prv'] = $certificate['keypaste']; // (already base64-encoded)
+		} elseif ($certificate['keylength'] != 'custom') {
```
jim-p, line 1376: What's the logic behind this test? It would appear to make a new key any time the function is run on an existing certificate rather than keeping the existing key.

It should only generate a new key if there is a change in the `keylength` value.

vktg: Updated
now it compares the `keylength` with the `$cert['prv']`

PR #882, opened 2020-06-16
Fix small typo

dagwieers:
I noticed this typo in the pfSense dashboard.

### DIFF
--- a/www/pfSense-pkg-squid/files/usr/local/pkg/squid.xml
+++ b/www/pfSense-pkg-squid/files/usr/local/pkg/squid.xml
@@ -57,7 +57,7 @@
 		<name>c-icap</name>
 		<rcfile>c-icap.sh</rcfile>
 		<executable>c-icap</executable>
-		<description>ICAP Inteface for Squid and ClamAV integration</description>
+		<description>ICAP Interface for Squid and ClamAV integration</description>
 	</service>
 	<plugins>
 		<item>


dagwieers:
And there are 6 more like this, but in pkg-descr and code comments.

jim-p:
Can you point to the other typos specifically? I only see this one single instance of the typo in the PR, and I didn't see any in the `www/pfSense-pkg-squid/pkg-descr`.

I do see one in an LCDProc page, however.

```shell
$ find . -type d -name 'pfSense-pkg-*' | xargs grep -nri -H 'Inteface'
./www/pfSense-pkg-squid/files/usr/local/pkg/squid.xml:60:		<description>ICAP Inteface for Squid and ClamAV integration</description>
./sysutils/pfSense-pkg-LCDproc/files/usr/local/pkg/lcdproc_client.php:454:	Assumes ALL CARP status are the same for all the intefaces.
```

dagwieers:
@jim-p I used GitHub search for this, so maybe that is less accurate than I expected :roll_eyes: 
https://github.com/pfsense/FreeBSD-ports/search?q=inteface&unscoped_q=inteface

jim-p:
Ah! The rest of those are actually in FreeBSD ports we do not maintain.

I committed a fix manually for this one in squid and in LCDProc, and another in the base system. This PR can be closed.

dagwieers:
Thanks !

PR #891, opened 2020-07-04
Squid Auth maximum user IP configuration. Implements #10725

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10725
- [X] Ready for review

Allows to set `max_user_ip` in the WebGUI

from http://www.squid-cache.org/Doc/config/acl/:
```
acl aclname max_user_ip [-s] number
# This will be matched when the user attempts to log in from more
# than <number> different ip addresses. The authenticate_ip_ttl
# parameter controls the timeout on the ip entries. [fast]
# If -s is specified the limit is strict, denying browsing
# from any further IP addresses until the ttl has expired. Without
# -s Squid will just annoy the user by "randomly" denying requests.
# (the counter is reset each time the limit is reached and a
# request is denied)
# NOTE: in acceleration mode or where there is mesh of child proxies,
# clients may appear to come from multiple addresses if they are
# going through proxy farms, so a limit of 1 may cause user problems.
```


www/pfSense-pkg-squid/files/usr/local/pkg/squid_auth.xml
```diff
@@ -141,6 +141,22 @@
 			<size>5</size>
 			<default_value>5</default_value>
 		</field>
+		<field>
+			<fielddescr>Authentication Max User IP</fielddescr>
```
jim-p, line 145: Change "IP" to "IP Addresses"

www/pfSense-pkg-squid/files/usr/local/pkg/squid_auth.xml
```diff
@@ -141,6 +141,22 @@
...
+			<fieldname>max_user_ip</fieldname>
+			<description>
+				<![CDATA[
+				This will be matched when the user attempts to log in from more than <number> different IP addresses.<br/>
```
jim-p, line 149: "Enforces a limit to the number of unique IP addresses from which a single user can login."

www/pfSense-pkg-squid/files/usr/local/pkg/squid_auth.xml
```diff
@@ -141,6 +141,22 @@
...
+				Browsing from any further IP addresses is denying until the Authentication TTL has expired.<br/>
```
jim-p, line 150: "Attempts to login from additional IP addresses are denied until the Authentication TTL has expired"

www/pfSense-pkg-squid/files/usr/local/pkg/squid_auth.xml
```diff
@@ -141,6 +141,22 @@
...
+				<span class="text-info">Default: none</span>
+				<div class="infoblock">
+				<span class="text-info">Note:</span> In acceleration mode or where there is mesh of child proxies, clients may appear to come from multiple addresses if they are going through proxy farms, so a limit of 1 may cause user problems.
```
jim-p, line 153: This seems a bit confusing in the wording. Perhaps:

> Acceleration mode or multiple child proxies in a mesh or farm may cause clients to appear as though they are using multiple IP addresses. A limit of 1 may cause problems for users in these cases.

vktg: Ok, fixed

PR #900, opened 2020-07-10
Suricata 5.0.3 -- Update Suricata binary to version 5.0.3 to keep pace with upstream.

bmeeks8:
This updates the Suricata binary on pfSense to the latest 5.0.3 version from upstream. Release Notes can be found here:  [https://suricata-ids.org/2020/04/28/suricata-5-0-3-released/](https://suricata-ids.org/2020/04/28/suricata-5-0-3-released/).

This binary update is compatible with both pfSense-2.4.5 and pfSense-2.5 AMD64 versions. This version is not compatible with ARM 32-bit hardware platforms!

### DIFF
--- a/security/suricata/Makefile
+++ b/security/suricata/Makefile
@@ -2,8 +2,7 @@
 # $FreeBSD$
 
 PORTNAME=	suricata
-DISTVERSION=	5.0.2
-PORTREVISION=	3
+DISTVERSION=	5.0.3
 CATEGORIES=	security
 MASTER_SITES=	https://www.openinfosecfoundation.org/download/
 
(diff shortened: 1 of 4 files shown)


rbgarga:
@bmeeks8 are you going to submit another PR changing version of pfSense-pkg-suricata or should we do it together with binary upgrade?

bmeeks8:
> @bmeeks8 are you going to submit another PR changing version of pfSense-pkg-suricata or should we do it together with binary upgrade?

Yes, I am submitting a separate PR for the GUI package. See email I sent you earlier. So this PR for binary and a separate PR for GUI. All are for 2.5-DEVEL branch for now and AMD64 or 64-bit ARM platforms only.

PR #908, opened 2020-07-17
Prevent users from creating new ACMEv1 keys. Implements #10769

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10769
- [X] Ready for review

https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430:
In June of 2020 we will stop allowing new domains to validate via ACMEv1.
In June of 2021 we will entirely disable ACMEv1 as a viable way to get a Let’s Encrypt certificate.


vktg:
Done,
it will now migrate from v1 to v2 during pkg installation.

PR #912, opened 2020-07-22
ARPwatch improvements. Issue #10770

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10770
- [X] Ready for review

- Allow to disable email notifications by removing recipient email  https://redmine.pfsense.org/issues/10770 
- Do not send cron email notifications https://redmine.pfsense.org/issues/10771
- Do not show PHP error if `ethercodes.dat` file is not found:
`[22-Jul-2020 10:11:21 Europe/Moscow] PHP Warning:  file_get_contents(/usr/local/arpwatch/ethercodes.dat): failed to open stream: No such file or directory in /usr/local/pkg/arpwatch.inc on line 170`

### DIFF
--- a/net-mgmt/pfSense-pkg-arpwatch/files/usr/local/pkg/arpwatch.inc
+++ b/net-mgmt/pfSense-pkg-arpwatch/files/usr/local/pkg/arpwatch.inc
@@ -63,9 +63,11 @@ function arpwatch_install_command() {
 			    .' '.$disable_zero
 			    .' '.$enable_zeropad
 			    .' -f '.escapeshellarg($arp_file)
-			    .' -i '.escapeshellarg($ifname)
-			    .' -w '.escapeshellarg($notifications_recipient)
-			    ."\n";
+			    .' -i '.escapeshellarg($ifname);
+			if (!empty($notifications_recipient)) {
+			    $rc['start'] .= ' -w '.escapeshellarg($notifications_recipient);
+			}
+			$rc['start'] .= "\n";
 		}
 	}
 
@@ -108,7 +110,8 @@ function arpwatch_validate_input($post, &$input_errors) {
 		}
 	}
 
-	if (filter_var($post['notifications_recipient'], FILTER_VALIDATE_EMAIL) === false) {
+	if (!empty($post['notifications_recipient']) &&
+	    (filter_var($post['notifications_recipient'], FILTER_VALIDATE_EMAIL) === false)) {
 		$input_errors[] = gettext("{$post['notifications_recipient']} is not a valid email address.");
 	}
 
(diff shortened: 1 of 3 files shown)


net-mgmt/pfSense-pkg-arpwatch/files/usr/local/pkg/arpwatch.inc
```diff
@@ -63,9 +63,11 @@ function arpwatch_install_command() {
...
 			    .' '.$enable_zeropad
 			    .' -f '.escapeshellarg($arp_file)
-			    .' -i '.escapeshellarg($ifname)
-			    .' -w '.escapeshellarg($notifications_recipient)
-			    ."\n";
+			    .' -i '.escapeshellarg($ifname);
+			    if (!empty($notifications_recipient)) {
+				    $rc['start'] .= ' -w '.escapeshellarg($notifications_recipient);
+			    }
+			    $rc['start'] .= "\n";
```
jim-p, line 70: These lines do not need to be indented the extra spaces here. They should be aligned with line 60. Lines 61-66 are indented more as they are continuations of the statement on line 60.

net-mgmt/pfSense-pkg-arpwatch/files/usr/local/pkg/arpwatch.inc
```diff
@@ -228,7 +232,7 @@ $fd = fopen('php://stdin','r');
 $message = stream_get_contents($fd);
 fclose($fd);
 
-if (false !== $message) {
+if ((false !== $message) && (false === strpos($message, ': Cron '))) {
```
jim-p, line 235: Should this be done unconditionally in this way? The feature request mentioned adding a checkbox/option for it.

net-mgmt/pfSense-pkg-arpwatch/files/usr/local/pkg/arpwatch.inc
vktg: Fixed

net-mgmt/pfSense-pkg-arpwatch/files/usr/local/pkg/arpwatch.inc
vktg: Checkbox added, but this is weird behavior anyway
https://redmine.pfsense.org/issues/8454

jim-p: Unfortunately, since arpwatch can only use `/usr/sbin/sendmail` to send mail, anything else that wants `/usr/sbin/sendmail` will also take advantage of it being there (Notably Cron). There isn't really a good way around that as far as I'm aware. Unless the script can check something in its environment to tell it's being called by arpwatch specifically and nothing else.

vktg: it should be OK with this `Disable Cron emails` checkbox for now

PR #913, opened 2020-07-24
Add ldapcachetime setting to squidguard. Implements #10785

coffeelover:
Allow setting the ldapcachetime for squidguard.

This caches the search results of ldap based group filters and allow scaling for large environments.

Redmine Issue: https://redmine.pfsense.org/issues/10785


coffeelover:
I increased the PORTREVISION, thanks for the hint

PR #915, opened 2020-08-01
Change UPS Type field to optional

adamtbeames:
UPS Type field needs to be optional not required as usb UPS devices need that field to be set to "BLANK".  Its not possible to use a usb UPS with that field set to required.

### DIFF
--- a/sysutils/pfSense-pkg-apcupsd/files/usr/local/pkg/apcupsd.xml
+++ b/sysutils/pfSense-pkg-apcupsd/files/usr/local/pkg/apcupsd.xml
@@ -148,7 +148,7 @@
 			</description>
 			<type>input</type>
 			<size>60</size>
-			<required>true</required>
+			<required>false</required>
 		</field>
 		<field>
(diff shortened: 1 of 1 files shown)


jim-p:
To effectively track changes and bug/feature requests, we require an entry on the [pfSense Redmine issue tracker](https://redmine.pfsense.org) associated with every pull request, and likewise the Redmine entry should also have a link back to the pull request.

Please associate this PR to a Redmine issue either by locating an existing issue at https://redmine.pfsense.org or by creating a new issue. Add a link to the Redmine issue that points to this PR, and also add a link on the PR to the Redmine issue.

For more information, see https://www.netgate.com/docs/pfsense/development/submitting-a-pull-request-via-github.html

Thanks!

dennypage:
I think most people using this package over the past few years are using USB based devices, so I’m a bit surprised by this. Did something change recently?

rbgarga:
@adamtbeames please open a redmine ticket with details about the issue you are having and then we can re-open this PR if needed after discussion

PR #917, opened 2020-08-11
BIND graceful shutdown. Fixes #10824

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10824
- [X] Ready for review

Bind is killed quite TERMinally in the /usr/local/etc/rc.d/named.sh script - with a SIGTERM. This causes the server to stop uncleanly.

Bug is that this causes journal files with dynamic zone updates (e.g. DHCP server updating named zones) to not be synchronized with zones and fail to reload with errors such as:

> journal rollforward failed: journal out of sync with zone

Could be fixed with an "rndc sync -clean" before shutdown.

https://www.freebsd.org/cgi/man.cgi?query=rndc&apropos=0&sektion=0&manpath=FreeBSD+12.1-RELEASE+and+Ports&arch=default&format=html:
```
sync [-clean] [zone [class [view]]]
	   Sync	changes	in the journal file for	a dynamic zone to the master
	   file. If the	"-clean" option	is specified, the journal file is also
	   removed. If no zone is specified, then all zones are	synced.
```


vktg:
Updated,
see https://redmine.pfsense.org/issues/10824#note-5

PR #925, opened 2020-09-07
Remove bandwidthd pgsql options. Issue #9809

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/9809
- [X] Ready for review

Remove PostgreSQL feature from package

current bandwidthd port don't support postgresql:
https://github.com/pfsense/FreeBSD-ports/blob/e8b54d83912d6390289df50bd8a1806447eabacb/net-mgmt/bandwidthd/Makefile#L19-L22

removed by port maintainer? (not found in history)

### DIFF
--- a/net-mgmt/pfSense-pkg-bandwidthd/Makefile
+++ b/net-mgmt/pfSense-pkg-bandwidthd/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-bandwidthd
 PORTVERSION=	0.7.4
-PORTREVISION=	1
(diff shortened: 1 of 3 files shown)


jim-p (approved):
I think at some point there were custom build options that had allowed it to work, but were not in the FreeBSD port. Then those were removed (likely years ago when things switched to pkg) and then so few people used it that it was lost/forgotten.

Probably safe to remove at this point, though if someone does need it, we could ask the port maintainer to add a port option to enable it and setup the dependency properly.

PR #936, opened 2020-09-18
Feature #10915 ; tweak sudo.inc for better support of NRPE

infraweavers:
Change to `sudo.inc` in `security/pfSense-pkg-sudo` for better support of `NRPE`:
Add the 'User: nagios' to the drop down list of available users when the `net-mgmt/pfSense-pkg-nrpe` package is installed.

https://redmine.pfsense.org/issues/10915


security/pfSense-pkg-sudo/files/usr/local/pkg/sudo.inc
```diff
@@ -126,6 +127,15 @@ function sudo_get_users($list_all_user = false) {
...
 	$users[] = $tmpuser;
+	
+	/* Make an entry for nagios, if the nrpe package is installed */
+        $nrpePackage = get_pkg_info('net-mgmt/pfSense-pkg-nrpe', true, true);
+        if (!empty($nrpePackage)) {
+                $tmpuser = array();
+                $tmpuser["name"] = "user:nagios";
+                $tmpuser["descr"] = "User: nagios";
+                $users[] = $tmpuser;
+        }
```
rbgarga, line 138: Please use hard tabs to indent instead of spaces

rbgarga (changes requested):
In order to get a new package built you must bump PORTVERSION or PORTREVISION on security/pfSense-pkg-sudo/Makefile

security/pfSense-pkg-sudo/files/usr/local/pkg/sudo.inc
infraweavers: Apologies - change applied

rbgarga (approved):

jim-p (changes requested):
I don't like hardcoding things in this manner. It solves a problem for this one package and one package only. I'd much prefer to see a more generalized "plugin" style setup where the sudo package checks to see if other packages want to register users with sudo. Look at similar functions in the base system for `plugin_carp`, `plugin_certificates` or the hook system used for `ipsec_reload_function`.

infraweavers:
> 
> 
> I don't like hardcoding things in this manner. It solves a problem for this one package and one package only. I'd much prefer to see a more generalized "plugin" style setup where the sudo package checks to see if other packages want to register users with sudo. Look at similar functions in the base system for `plugin_carp`, `plugin_certificates` or the hook system used for `ipsec_reload_function`.

Will have a look and assess how much work is involved.

infraweavers:
> Will have a look and assess how much work is involved.

Hi @jim-p, @rbgarga,

Based the last comments, this diagram is our attempt to demonstrate our understanding of what jim-p is suggesting. I think if you are happy, we are happy to have a stab at implementing it that way.

An alternative idea, is embedding the list of users associated to a package in the package.xml, but we are guessing that would mean a change to the XML Document Type Definition - we've not been able to find the '../schema/packages.dtd' definition anywhere? It would also probably mean writing a bespoke function in the pfSense to return said list of users for sudo, which seems a bit specialised.

If there anyway of enumerating system accounts so that, the NRPE package for example, doesn't have to have a function which is hard coded to return 'User: nagios'?

image

jim-p:
You should be able to dynamically get the name of the user(s) required by a package with something like `/usr/local/sbin/pkg-static query '%U' nrpe3`.

That may be an alternative to the plugin system, actually, now that I think about it. Just have sudo build a list from the output of `/usr/local/sbin/pkg-static query '%U'` since without a package name it returns all users required by all packages.

infraweavers:
Apologies,

Due to sheer incompetence using GitHub we are going to close this PR and create a new one

PR #940, opened 2020-09-23
Allow to disable Squid caching completly. Implements #10913

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10913
- [X] Ready for review


apollo13:
@vktg Many thanks. I have downloaded the patch and applied the relevant parts via "system patches" and it is working like a charm.

jim-p (approved):

PR #947, opened 2020-09-30
Feature #10915 ; tweak sudo.inc for better support of NRPE

infraweavers:
This PR supersedes https://github.com/pfsense/FreeBSD-ports/pull/936.
Redmine Feature Request: https://redmine.pfsense.org/issues/10915

We have a requirement to permit NRPE to run custom commands as root so that we can, for instance, monitor VPN connected clients and Snort alerts on given interfaces. The NRPE plugin already has a mechanism to support using sudo. Right now the only way to add definitions to sudoers for the NRPE user (nagios) is by manually editing configuration.

This change includes users and groups created by packages in lists available for configuring sudoers via the sudo plugin.


security/pfSense-pkg-sudo/files/usr/local/pkg/sudo.inc
```diff
@@ -115,17 +116,37 @@ function sudo_get_users($list_all_user = false) {
...
+	/* Make an entry for root, even though admin is essentially the same as root, they are distinct. */
+	$a_user[]['name'] = 'root';
+
+	/* Include users created by other packages */
+	$stdout = '';
+	$stderr = '';
+
+	pkg_exec('query %U', $stdout, $stderr);
+	foreach (explode(PHP_EOL, $stdout) as $user) {
+		if ($user == ''){
```
jim-p, line 129: Needs a space between `)` and `{`

security/pfSense-pkg-sudo/files/usr/local/pkg/sudo.inc
```diff
@@ -115,17 +116,37 @@ function sudo_get_users($list_all_user = false) {
...
 	if (!is_array($config['system']['group'])) {
 		$config['system']['group'] = array();
 	}
 	$a_group = &$config['system']['group'];
-	$users = array();
+	
+	/* Include groups created by other packages */
+	pkg_exec('query %G', $stdout, $stderr);
+	foreach (explode(PHP_EOL, $stdout) as $group) {
+		if ($group == ''){
```
jim-p, line 143: Needs a space between `)` and `{`

jim-p (approved):
That looks much better, thanks!

jim-p (changes requested):
Accidentally marked it approved. There are two very minor syntax formatting changes to make but that's all.

jim-p (approved):

PR #950, opened 2020-10-01
FRR - Fix typo in frr_bgp.xml

bmhughes:
As title.

Continuation of: https://redmine.pfsense.org/issues/10789


jim-p (changes requested):

bmhughes:
> 
> 
> 

Ah missed that, done.

jim-p (approved):

PR #952, opened 2020-10-03
OpenVPN Client Export privilege to get user config. Issue #6176

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/6176
- [X] Ready for review

limit access for users to their own client configs. users don't need to have access to the configs of other users

### DIFF
--- a/security/pfSense-pkg-openvpn-client-export/Makefile
+++ b/security/pfSense-pkg-openvpn-client-export/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-openvpn-client-export
 PORTVERSION=	1.4.23
-PORTREVISION=	2
+PORTREVISION=	3
 CATEGORIES=	security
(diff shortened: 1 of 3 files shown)


jim-p:
I'm not inclined to start down this slippery slope, since it will likely lead to people significantly weakening the security of their infrastructure in the name of convenience. It's likely that people will see this and grant access to the GUI for people who normally shouldn't/wouldn't have access just so they can download their own VPN clients.

andreicaba:
hm... wouldn't - still - a feature in pfSense code that enables plugin code writers to easily give/take permissions for specific parts of their plugin's page only for specific users?... Wouldn't this be - anyway - a security *enhacement?... :)
Just my two cents on this, as an IT Architect and a Security Engineer...

have a nice day there!...

jim-p:
There are already mechanisms for packages to add their own permissions for pages, that's not even close to the same issue as this.

PR #953, opened 2020-10-03
Snort netmap support check. Implements #10950

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10950
- [X] Ready for review

input validation to prevent users from choosing an unsupported card

### DIFF
--- a/security/pfSense-pkg-snort/files/usr/local/www/snort/snort_interfaces_edit.php
+++ b/security/pfSense-pkg-snort/files/usr/local/www/snort/snort_interfaces_edit.php
@@ -29,6 +29,13 @@ global $g, $config, $rebuild_rules;
 $snortdir = SNORTDIR;
 $snortlogdir = SNORTLOGDIR;
 
+$netmapifs = array('cc', 'cxl', 'cxgbe', 'em', 'igb', 'em', 'lem', 'ix', 'ixgbe', 'ixl', 're', 'vtnet');
+if (pfs_version_compare(false, 2.4, $g['product_version'])) {
+	/* add FreeBSD 12 iflib(4) supported devices */
+	$netmapifs = array_merge($netmapifs, array('ice', 'bnxt', 'vmx'));
+	sort($netmapifs);
+}
+
 if (!is_array($config['installedpackages']['snortglobal']['rule'])) {
 	$config['installedpackages']['snortglobal']['rule'] = array();
 }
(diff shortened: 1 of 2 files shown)


bmeeks8:
I am okay with this change for Snort. I do suggest adding the same to Suricata, though. However, a Suricata binary release update is due on October 8th, so we could hold off on the Suricata change and I can push it along with the binary update (to 5.0.4).

security/pfSense-pkg-snort/files/usr/local/www/snort/snort_interfaces_edit.php
```diff
@@ -243,6 +243,21 @@
 		}
 	}
 
+	if ($_POST['ips_mode'] == 'ips_mode_inline') {
+		$netmapifs = array('cxgbe', 'em', 'igb', 'em', 'lem', 'ixgbe', 'ixl', 're', 'vtnet');
```
jim-p, line 247: The `ixgbe` driver also covers `ix` interfaces.
The `cxgbe` driver also covers `cxl` and `cc`

jim-p (changes requested):

bmeeks8:
After thinking this through some more, things get a little murkier when you have to navigate between FreeBSD-11 and FreeBSD-12. My understanding is that in FreeBSD-12 the netmap functionality is now entirely within the new _iflib_ framework, so that means many drivers can take advantage of netmap compatibility just from using the _iflib_ API. In FreeBSD-11, each NIC driver has to individually support netmap operation. So I can see a scenario where a given NIC does not work on FreeBSD-11 but would work on FreeBSD-12 with netmap (based on the _iflib_ integration).

So the check proposed by Viktor may have to go a bit deeper and examine the current OS version (FreeBSD-11 or FreeBSD-12) in order to adequately determine which NIC driver supports netmap and which does not.

vktg:
Ok, updated
Now it adds extra FreeBSD 12 iflib supported devices to the `$netmapifs` list if pfSense > 2.4 is detected

bmeeks8:
I think the new changes will work. Thank you Viktor.

I will pull this change into Suricata as well. Upstream released both 5.0.4 and 6.0.0 yesterday. Still deciding which way to go on pfSense. I'm leaning towards 6.0.0. I will create a Suricata update pull request in the near future.

And just FYI, upstream also released Suricata 4.1.9 yesterday, but that will be all for the 4.x branch. It goes EOL on December 31, 2020 (so this year). That means the 32-bit ARM hardware will wind up frozen at 4.1.9 for Suricata.

jim-p (approved):
Looks good now. Will need a version bump, unless it's going to wait until the next update for snort to be included.

vktg:
Version bump

PR #954, opened 2020-10-03
pfBlockerNG L2TP VPN support. Implements #8332

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/8332
- [X] Ready for review


gogglespisano:
The code looks ok to me, but I don't know that code well.

I think the comment for the function is out of date now that it's "-devel" and settings have been moved.

// Collect all Interfaces for General Tab and DNSBL Firewall Permit Rule
function pfb_build_if_list($show_wan=FALSE, $show_groups=FALSE)

jim-p (approved):

PR #955, opened 2020-10-04
FRR Add support for OSPFv3 areas

bmhughes:
Contrary to the FRR documentation (http://docs.frrouting.org/en/latest/ospf6d.html#ospf6-area)
OSPFv3 areas are fully supported in FRR (https://github.com/FRRouting/frr/issues/2453)

- Add area tab to OSPFv3 configuration.
- Split OSPFv3 configuration generation out into Area/Interface/Main functions as OSPFv2.
- Update validation for OSPFv3 to support multiple areas.

Continuation of Redmine issue: https://redmine.pfsense.org/issues/10789


jim-p (approved):
As long as it's working now, it should be OK. Last time I tried it in FRR directly it failed in various ways but that was a couple FRR versions ago (Around 4.x or 5.x I think).

bmhughes:
> 
> 
> As long as it's working now, it should be OK. Last time I tried it in FRR directly it failed in various ways but that was a couple FRR versions ago (Around 4.x or 5.x I think).

I've quickly tested it with a few VMs and it seemed ok without pushing it, VyOS has support as well so i'm presuming it works as designed at this point.

PR #956, opened 2020-10-08
FRR OSPF default area fix. Issue #10964

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10964
- [X] Ready for review

The definition of `$ospfd_conf` in `frr_generate_config_ospf_interfaces()` is missed after https://github.com/pfsense/FreeBSD-ports/pull/955 

used in https://github.com/pfsense/FreeBSD-ports/blob/94a6e4df3b9887f3fe7835804bc4192b1355066d/net/pfSense-pkg-frr/files/usr/local/pkg/frr/inc/frr_ospf.inc#L244:


net/pfSense-pkg-frr/files/usr/local/pkg/frr/inc/frr_ospf.inc
```diff
@@ -169,6 +169,9 @@ function frr_generate_config_ospf_areas() {
 
 function frr_generate_config_ospf_interfaces() {
 	global $config, $passive_interfaces, $frr_integrated_config;
+	if (is_array($config['installedpackages']['frrospfd']['config'])) {
+		$ospfd_conf = &$config['installedpackages']['frrospfd']['config'][0];
+	}
```
rbgarga, line 174: It's better to initialize this config item with `init_config_arr()` as done few lines below

rbgarga (changes requested):

net/pfSense-pkg-frr/files/usr/local/pkg/frr/inc/frr_ospf.inc
vktg: done

rbgarga (approved):

PR #962, opened 2020-10-14
pfBlockerNG pfb_aliastables() fix. Issue #10927

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10927
- [X] Ready for review

Put `parse_config()` before `&$config` references,
otherwise it would not put `$earlyshellcmd` to config.xml


BBcan177:
Thanks for the patch, looks good.

PR #965, opened 2020-10-18
Wrap net-snmp IPv6 listen address in square brackets

bmhughes:
With net-snmp v5.7 (current 2.4.5) snmpd will accept an IPv6 listen address without square brackets, with v5.9 (2.5.0 dev) it requires them to be wrapped in square brackets and will silently fail to start without them. Running in debug to stdout will show a failed to bind error for the v6 address.

Both versions will accept the address in square brackets so do it regardless.

Redmine: https://redmine.pfsense.org/issues/10990

### DIFF
--- a/net-mgmt/pfSense-pkg-net-snmp/Makefile
+++ b/net-mgmt/pfSense-pkg-net-snmp/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-net-snmp
 PORTVERSION=	0.1.5
-PORTREVISION=	4
+PORTREVISION=	5
 CATEGORIES=	net-mgmt
 MASTER_SITES=	# empty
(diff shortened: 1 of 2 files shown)


jim-p (changes requested):
My only question is if NET-SNMP 5.7 will accept the addresses in brackets. If so, it's OK as-is. If not, it will need a version test to make sure this change only happens on 2.5.0.

bmhughes:
> 
> 
> My only question is if NET-SNMP 5.7 will accept the addresses in brackets. If so, it's OK as-is. If not, it will need a version test to make sure this change only happens on 2.5.0.

Yes 5.7 will take both formats. I thought i'd put that in the commit message but apparently not.

jim-p (approved):

PR #967, opened 2020-10-22
Added xmlrpc support to the sudo package

infraweavers:
Will only work on Pfsense > 2.4 however as this is adding a feature, we think it's reasonable to not have backwards compatibility here.


security/pfSense-pkg-sudo/files/usr/local/pkg/sudo.inc
```diff
@@ -203,4 +206,73 @@ function sudo_validate_commands(&$input_errors) {
...
+		$rpc_client = new pfsense_xmlrpc_client();
+		$rpc_client->setConnectionData($sync_to_ip, $port, $username, $password, $protocol);
+		$resp = $rpc_client->xmlrpc_method('merge_installedpackages_section', $xml, $synctimeout);
+		$resp = $rpc_client->xmlrpc_exec_php($execcmd, $synctimeout);
+		$resp = $rpc_client->xmlrpc_exec_php($execcmd2, $synctimeout);
+	}
+}
+
+
+if(!function_exists('pf_version')) {
```
rbgarga, line 272: This function can be removed

security/pfSense-pkg-sudo/files/usr/local/pkg/sudo.inc
```diff
@@ -203,4 +206,73 @@ function sudo_validate_commands(&$input_errors) {
...
+	/*************************************************/
+	$execcmd = "file_put_contents('/tmp/remote_sudo_config_apply.php', '{$remote_sudo_config_apply}');";
+
+	/*************************************************/
+	/* Now assemble a command to execute the         */
+	/* previously sent PHP file in the background.   */
+	/*************************************************/
+	$execcmd2 = "exec(\"/usr/local/bin/php-cgi -f '/tmp/remote_sudo_config_apply.php' > /dev/null 2>&1 &\");";
+
+	if (pf_version() >= "2.4") {
```
rbgarga, line 261: No need to check if pfSense is >= 2.4 since the smallest supported version is 2.4.5-p1

security/pfSense-pkg-sudo/files/usr/local/pkg/sudo.inc
```diff
@@ -203,4 +206,73 @@ function sudo_validate_commands(&$input_errors) {
...
+		$system_carp = $config['hasync'];
+		// XMLRPC sync is currently only supported over connections using the same protocol and port as this system
+		if ($config['system']['webgui']['protocol'] == "http") {
+			$varsyncprotocol = "http";
+			$varsyncport = $config['system']['webgui']['port'] ?: '80';
+		} else {
+			$varsyncprotocol = "https";
+			$varsyncport = $config['system']['webgui']['port'] ?: '443';
+		}
+		sudo_do_xmlrpc_sync($system_carp['synchronizetoip'], $varsyncport, $varsyncprotocol, $system_carp['username'], $system_carp['password'], 100);
```
rbgarga, line 222: Please break this really long line.  pfSense style says we should try to keep lines in ~80 columns.  Multiline statements must be indented with 4 spaces after previous line indentation.

security/pfSense-pkg-sudo/files/usr/local/pkg/sudo.inc
```diff
@@ -203,4 +206,73 @@ function sudo_validate_commands(&$input_errors) {
 		}
 	}
 }
+
+function sudo_sync_config() {
+	global $config;
+	if (is_array($config['hasync'])) {
```
rbgarga, line 212: You can reverse this check and just return when it's false, then next block will be one level less indented

rbgarga (changes requested):
All these comments are also valid for PR #968

infraweavers:
@rbgarga all changed on this one

rbgarga (approved):

security/pfSense-pkg-sudo/files/usr/local/pkg/sudo.inc
```diff
@@ -203,4 +206,70 @@ function sudo_validate_commands(&$input_errors) {
...
+	/* Now assemble a command to execute the         */
+	/* previously sent PHP file in the background.   */
+	/*************************************************/
+	$execcmd2 = "exec(\"/usr/local/bin/php-cgi -f '/tmp/remote_sudo_config_apply.php' > /dev/null 2>&1 &\");";
+	$synctimeout = intval($synctimeout);
+	$rpc_client = new pfsense_xmlrpc_client();
+	$rpc_client->setConnectionData($sync_to_ip, $port, $username, $password, $protocol);
+	$resp = $rpc_client->xmlrpc_method('merge_installedpackages_section', $xml, $synctimeout);
+	$resp = $rpc_client->xmlrpc_exec_php($execcmd, $synctimeout);
+	$resp = $rpc_client->xmlrpc_exec_php($execcmd2, $synctimeout);
```
jim-p, line 272: Why are you using a function which executes PHP to execute a shell command which then runs PHP as a script?

Just execute the PHP directly without jumping through all the extra hoops.

jim-p (changes requested):

infraweavers:
@jim-p changed, we basically copied snort's implementation hence the complexity

jim-p:
That makes sense, Snort is a lot more complex and may need to run that process in the background since it will take a lot longer. Not needed in these cases.

jim-p (approved):

rbgarga:
Manually merged in order to squash all commits into a single one.  Thanks!

PR #968, opened 2020-10-23
Added xmlrpc support to the nrpe package

infraweavers:
Will only work on Pfsense > 2.4 however as this is adding a feature, we think it's reasonable to not have backwards compatibility here.


infraweavers:
@rbgarga All done here too

rbgarga (approved):

net-mgmt/pfSense-pkg-nrpe/files/usr/local/pkg/nrpe.inc
```diff
@@ -253,4 +254,69 @@ function nrpe_custom_php_validation_command($post, &$input_errors) {
...
+	/* Now assemble a command to execute the         */
+	/* previously sent PHP file in the background.   */
+	/*************************************************/
+	$execcmd2 = "exec(\"/usr/local/bin/php-cgi -f '/tmp/remote_nrpe_config_apply.php' > /dev/null 2>&1 &\");";
+	$synctimeout = intval($synctimeout);
+	$rpc_client = new pfsense_xmlrpc_client();
+	$rpc_client->setConnectionData($sync_to_ip, $port, $username, $password, $protocol);
+	$resp = $rpc_client->xmlrpc_method('merge_installedpackages_section', $xml, $synctimeout);
+	$resp = $rpc_client->xmlrpc_exec_php($execcmd, $synctimeout);
+	$resp = $rpc_client->xmlrpc_exec_php($execcmd2, $synctimeout);
```
jim-p, line 319: Same concern here as with sudo. This looks unnecessary when the PHP can be run directly.

jim-p (changes requested):

net-mgmt/pfSense-pkg-nrpe/files/usr/local/pkg/nrpe.inc
infraweavers: Changed

jim-p (approved):

rbgarga:
Manually merged to get commits squashed.  Thanks!

PR #970, opened 2020-10-26
FRR fix BGP migrations and frrctl with password encryption

bmhughes:
- **Update BGP UI redistribution field name and add migration**
    BGP connected redistribution has been renamed from 'connectedsubnets' to 'connected' in the configuration builder function. Update the UI to match and add a migration action to 'frr_package_config_upgrade'.

-  **Fix missed address family global use**

-  **Refactor frrctl script to use vtysh**
    Allows FRR status pages to function with password encryption enabled

Redmine: https://redmine.pfsense.org/issues/10789


jim-p (approved):

net/pfSense-pkg-frr/files/usr/local/pkg/frr/inc/frr_bgp.inc
```diff
@@ -967,4 +967,3 @@ function frr_generate_config_bgp() {
 		$frr_integrated_config['rpki'] = frr_bgp_generate_rpki_servers();
 	}
 }
- 
```
rbgarga, line 970: The removal of EOL from last line created a conflict.  Please revert this

rbgarga (changes requested):

net/pfSense-pkg-frr/files/usr/local/pkg/frr/inc/frr_bgp.inc
bmhughes: That's strange I've got a newline in my local (up to date with github) branch and I certainly don't remember removing it consciously. I'll sort it out.

bmhughes: Hmm that's officially strange, looks like git(hub) is removing the newlines (from display at least) even though they're present in my local copy. Rebase fixed it.

jim-p: There wasn't a newline that got removed, there was an odd character on a line at the end of that file that broke the package. I removed it yesterday in f704edf8024c3ed86d8f642d4c3618930ebd6495

jim-p: See also: https://redmine.pfsense.org/issues/11011

PR #971, opened 2020-10-27
mailreport: Add flag to allow skipping email send if there is no report content. Implements #11008

Gosherm:
Adds a flag to email report configs that will allow for the email to not be sent if there was no output from the commands or content in the log files.

https://redmine.pfsense.org/issues/11008


jim-p (changes requested):

jim-p (approved):

rbgarga:
Manually merged to squash commits.  Thanks!

PR #979, opened 2020-10-30
Fix XMLRPC for standalone and secondaries - NRPE

infraweavers:
Package was not correctly detecting whether HA was configured or no. It now checks the configuration in more detail

In some configurations the `haconf` is set, however it contains missing values so we now defensively check this. It was erroring like so:

```
Exception calling XMLRPC method exec_php #1 : Unhandled XML_RPC2_InvalidUriException exception:Client URI 'https://:@:443/xmlrpc.php' is not valid#0 /usr/local/share/pear/XML/RPC2/Backend/Php/Client.php(80): XML_RPC2_Client-&gt;__construct('https://:@:443/...', Array)
#1 /usr/local/share/pear/XML/RPC2/Client.php(238): XML_RPC2_Backend_Php_Client-&gt;__construct('https://:@:443/...', Array)
#2 /etc/inc/xmlrpc_client.inc(92): XML_RPC2_Client::create('https://:@:443/...', Array)
#3 /etc/inc/xmlrpc_client.inc(153): pfsense_xmlrpc_client-&gt;xmlrpc_internal('merge_installed...', Array, 100)
#4 /usr/local/pkg/nrpe.inc(320): pfsense_xmlrpc_client-&gt;xmlrpc_method('merge_installed...', Array, 100)
#5 /usr/local/pkg/nrpe.inc(290): nrpe_do_xmlrpc_sync('', '443', 'https', '', '', 100)
#6 /usr/local/pkg/nrpe.inc(215): nrpe_sync_config()
#7 /usr/local/pkg/nrpe.inc(160): nrpe_custom_php_write_config()
#8 /usr/local/www/xmlrpc.php(141) : eval()'d code(2): nrpe_resync_package()
#9 /usr/local/www/xmlrpc.php(141): eval()
#10 /usr/local/share/pear/XML/RPC2/Server/CallHandler/Instance.php(141): pfsense_xmlrpc_server-&gt;exec_php('require_once('/...')
#11 /usr/local/share/pear/XML/RPC2/Backend/Php/Server.php(135): XML_RPC2_Server_Callhandler_Instance-&gt;__call('pfsense.exec_ph...', Array)
#12 /usr/local/share/pear/XML/RPC2/Backend/Php/Server.php(99): XML_RPC2_Backend_Php_Server-&gt;getResponse()
#13 /usr/local/www/xmlrpc.php(771): XML_RPC2_Backend_Php_Server-&gt;handleCall()
#14 {main} @ 2020-10-30 09:39:54 
```

PR #980, opened 2020-10-30
Fixing XMLRPCSync - Stunnel

infraweavers:
Same problem as: https://github.com/pfsense/FreeBSD-ports/pull/979


infraweavers:
added https://github.com/pfsense/FreeBSD-ports/pull/979#discussion_r515094156

PR #981, opened 2020-10-30
Fixing XMLRPCSync - Sudo

infraweavers:
Same problem as https://github.com/pfsense/FreeBSD-ports/pull/979


infraweavers:
Added https://github.com/pfsense/FreeBSD-ports/pull/979#discussion_r515094156

PR #993, opened 2020-11-15
pfSense-pfBlockerNG-devel 3.0.0

BBcan177:
# pfBlockerNG-devel v3.0.0 #

## CHANGELOG: ##

* DNSBL - Python Mode Integration
* TOP1M - Add Tranco as the default TOP1M Whitelist [ https://tranco-list.eu/ ]
* DNSBL VIP Type = CARP (High Availability Sync) improvements
* XMLRPC Sync Tab - Remove the CARP HA Skew from being sync'd
* Add Localhost at the default DNSBL Listening interface (Suggest all existing users use change to this interface)
* DNSBL Web Server is now bound to localhost instead of 0.0.0.0
* Widget - Packet counters are more accurate, however, the Clear counter option will now clear all pfSense counters.
* Widget - Tooltip Improvements
* Widget - Add IDN/Regex/TLD Allow entries
* DNSBL SQLite statistic databases are now located in the /var/unbound folder
* DNSBL IPv6 blocking
* Wizard - Add IPv6 checkbox
* Wizard - Add DNSBL Whitelist checkbox to allow users to disable the default DNSBL Whitelist.
* Alerts/Reports tab - Timeline Bar graph charts (via d3Pie)
* Alerts/Reports tab - Pause Icon added to halt the page refresh on each table.
* Alerts/Reports tab - Several new Settings to refine the Reports page output
* Alerts/Reports tab - Page refresh timer now starts after page has been fully loaded.
* Alerts/Reports tab - Add port lookup tool
* DNSBL Block page - Screen resolution improvements for device screen sizes
* ASN - Database updated
* DNSBL TLD database updated
* Add the DNSBL SafeSearch menu option to the User Privileges page
* CURL - Download speed improvements (CURLOPT_TCP_FASTOPEN, CURLOPT_TCP_NODELAY)
* DNSBL Web Server certificate improvements (by PiBa)
* DNSBL Restore Resolver DNS cache option on Cron/Force Update/Force Reloads
* DNSBL improvements to the Stop and Start of Unbound
* DNSBL Null Blocking mode with or without logging
* Add Threat Page lookups - GreyNoise, Shodan and Stop Forum Spam
* Improvement to pfblockerng.log messages
* Feeds added/removed/edit
* Several improvements to the Alerts/Reports Tab
* Several other under-the-hood improvements

## CAUTIONS: ##

1)

The DNS Resolver (Unbound) DHCP Registration option is not compatible with DNSBL Python mode. The pfSense devs are aware and changes are required to be made to the dhcpleases binary to stop/start Unbound instead of sending a SIGHUP. The use of this option and the Unbound Python mode will cause an Unbound crash.

If DHCP Registration is enabled in Unbound Python mode, or DHCP Registration enabled after Unbound Python mode is enabled, Unbound Python mode will be downgraded to Unbound mode to prevent Unbound from crashing.

2)

The DNS Resolver Unbound "Views" feaure it not currently compatable with DNSBL Python mode.


gogglespisano:
Is -devel the new NG?  I was told this would merge into NG a year or two ago.  Is it ever going to, or do we need to use -devel from now on?

BBcan177:
Continue to use devel, this will be pushed into the Release version once this has baked.

net/pfSense-pkg-pfBlockerNG-devel/pkg-plist
```diff
@@ -35,6 +35,9 @@ www/widgets/javascript/pfblockerng.js
 www/widgets/widgets/pfblockerng.widget.php
 www/wizards/pfblockerng_wizard.xml
 www/wizards/pfblockerng_wizard.inc
+var/unbound/pfb_unbound.py
+var/unbound/pfb_unbound_include.inc
+var/unbound/pfb_py_hsts.txt
 %%DATADIR%%/info.xml
```
rbgarga, line 41: These 3 lines must start with `/`, otherwise it considers them to be installed under `${PREFIX}`.  It will require to add `@dir /var/unbound` after them so the directory itself it removed if empty on pkg uninstall

rbgarga (changes requested):

net/pfSense-pkg-pfBlockerNG-devel/pkg-plist
BBcan177: ok

PR #994, opened 2020-11-18
Fix slow loading of HAProxy widget on dashboard, buy avoiding reapeated costly permissions checks.

pruiz:
In our current setup we have a few dozen frontends with a few backends under the hood. So for each server this is calling userHasPrivilage, which loads an parses config file every time.. adding ~2s to each loop iteration. This patch changes the code so privilegies are checked just once (before entering the foreach loop).


jim-p (changes requested):

pruiz:
@jim-p updated PORTVERSION as requested. ;)

jim-p (approved):

rbgarga:
Manually merged.  Thanks!

PR #998, opened 2020-12-02
pfBlocker phishing feeds update. Implements #11113

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/11113
- [X] Ready for review


rbgarga:
@BBcan177 please review

BBcan177:
Mailborder Phishing Updates - This feed uses the phishtank data, so there is no need for this feed.

The other two are compilation type feeds, and I typically recommend that users use Feeds from the original source. I have been asked to add these many times, so if we have to add these, then its best to create a new DNSBL Group "**Compilation**" and add these two feeds to this new Group. And add a Title that these are Compilation type feeds, and that they are not the original source of the data.

Don't bump the makefile, as this will be added when I submit my next version update.

vktg:
Updated

PR #1004, opened 2020-12-14
pfBlockerNG v3.0.0_6

BBcan177:
- Fix incorrect function name call
- Add safety belt for DNS Python mode and the DNS Resolver OpenVPN Client Registration option.
- Add a Phishing Army alternative feed.
- Remove any empty < config >< /config > config.xml entries
- DNSBL - NAT / Floating rule modifications when Localhost interface is selected
- Preliminary DNSBL Group Policy configuration


BBcan177:
@rbgarga Please merge this first: https://github.com/pfsense/FreeBSD-ports/pull/1003

PR #1006, opened 2020-12-15
HAProxy OCSP crontab fix. Issue #11135

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/11135
- [X] Ready for review

use correct `openssl ocsp -header` format,
see https://www.openssl.org/docs/man1.1.1/man1/openssl-ocsp.html:
```
-header name=value
Adds the header name with the specified value to the OCSP request that is sent to the responder. This may be repeated.
```
https://www.openssl.org/docs/man1.0.2/man1/ocsp.html:
```
-header name value
If sending a request to an OCSP server, then the specified header name and value are added to the HTTP request. Note that the name and value must be specified as two separate parameters, not as a single quoted string, and that the header name does not have the trailing colon. Some OCSP responders require a Host header; use this flag to provide it.
```

### DIFF
--- a/net/pfSense-pkg-haproxy-devel/files/usr/local/pkg/haproxy/haproxy.inc
+++ b/net/pfSense-pkg-haproxy-devel/files/usr/local/pkg/haproxy/haproxy.inc
@@ -41,6 +41,9 @@ $a_frontendmode['tcp'] = array('name' => "tcp", 'shortname' => "tcp");
 $a_frontendmode['health'] = array('name' => "health", 'shortname' => "health");
 */
 
+global $openssl111;
+$openssl111 = (OPENSSL_VERSION_NUMBER >= 269488128);
+
 global $a_acltypes;
 $a_acltypes = array();
 $a_acltypes["host_starts_with"] = array('name' => 'Host starts with:', 'casesensitive' => true,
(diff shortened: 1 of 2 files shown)


net/pfSense-pkg-haproxy-devel/files/usr/local/pkg/haproxy/haproxy.inc
```diff
@@ -41,6 +41,9 @@ $a_frontendmode['tcp'] = array('name' => "tcp", 'shortname' => "tcp");
 $a_frontendmode['health'] = array('name' => "health", 'shortname' => "health");
 */
 
+global $openssl111;
+$openssl111 = (OPENSSL_VERSION_NUMBER >= 269488128) ? true : false;
```
jim-p, line 45: The ternary operator here is redundant, if you assign the value of the test itself it has the same net result:
```php
$openssl111 = (OPENSSL_VERSION_NUMBER >= 269488128)
```

jim-p (approved):

net/pfSense-pkg-haproxy-devel/files/usr/local/pkg/haproxy/haproxy.inc
vktg: Fixed

PiBa-NL:
Thanks for fixing. The patch looks good to me, "PiBa approves :smiley:" for what that is worth.. A little weird that openssl changed their syntax..

PR #1007, opened 2020-12-19
pfSense-pkg-acme: Update web fileds for nic.ru dns validation method

Alexilmarranen:
Update fields for nic.ru validation method to match current dns api


jim-p (changes requested):

Alexilmarranen:
@jim-p thanks for the hint and sorry - I'm newbie in pfSense and FreeBSDPorts

jim-p (approved):

PR #1018, opened 2020-12-31
FRR 7.5

bmhughes:
Update the FRR port to 7.5 and update pfSense-pkg-frr to use 7.5 new features and other changes and fixes.

- FRR reload configuration when removing a list item
- Update FRR port to 7.5
- Add IP type to FRR access and prefix lists    
    - When used in a route-map the type needs to be known to select the correct `match` statement to use
    - Enhance validation so the correct address types can only be present
    - Add migrations
    - Add checkbox for `any` matches to unify behaviour between IPv4 and IPv6
- Add force service restart button to FRR main page
- FRR Improve logging
- Sort route-maps on save
- Add maxsockbuf tunable during install
- Add FRR 7.5 new features: 
    - BGP
      - Add force option to `maximum-prefix`
      - Add `maximum-prefix-out` option
      - Add neighbor shutdown message
      - Add neighbor automatic shutdown
      - Add global neighbor shutdown
  - BFD
    - Add profile support


jim-p:
The FRR binary port changes can't be in the same PR, they would be handled separately.

bmhughes:
Split into: https://github.com/pfsense/FreeBSD-ports/pull/1020 and https://github.com/pfsense/FreeBSD-ports/pull/1021

PR #1020, opened 2021-01-04
Update FRR port to 7.5

bmhughes:
- Update FRR port to 7.5


rbgarga:
I've merged new ports quarterly branch which brought FRR 7.5 to the tree.  Thanks!

PR #1021, opened 2021-01-04
pfSense-pkg-frr v1.0.0

bmhughes:
- FRR reload configuration when removing a list item
- Add IP type to FRR access and prefix lists    
    - When used in a route-map the type needs to be known to select the correct `match` statement to use
    - Enhance validation so the correct address types can only be present
    - Add migrations
    - Add checkbox for `any` matches to unify behaviour between IPv4 and IPv6
- Add force service restart button to FRR main page
- FRR Improve logging
- Sort route-maps on save
- Add maxsockbuf tunable during install
- Add FRR 7.5 new features: 
    - BGP
      - Add force option to `maximum-prefix`
      - Add `maximum-prefix-out` option
      - Add neighbor shutdown message
      - Add neighbor automatic shutdown
      - Add global neighbor shutdown
  - BFD
    - Add profile support

### DIFF
--- a/net/pfSense-pkg-frr/files/usr/local/pkg/frr.inc
+++ b/net/pfSense-pkg-frr/files/usr/local/pkg/frr.inc
@@ -26,6 +26,8 @@ require_once("service-utils.inc");
 require_once("util.inc");
 
 define('PKG_FRR_CONFIG_BASE', '/var/etc/frr');
+define('PKG_FRR_LIST_NONE_VALUE', array('name' => 'None', 'value' => 'none'));
+
 /* Since we need to embed this in strings & HERE Docs, copy to a var. Can't embed constants. */
 global $frr_config_base;
 $frr_config_base = PKG_FRR_CONFIG_BASE;
@@ -57,7 +59,7 @@ require_once("frr/inc/frr_ospf6.inc");
 /* Build a list of only CARP VIPs, formatted for use in a selection list */
 function frr_get_carp_list() {
 	$list = array();
-	$list[] = array("name" => "none", "value" => "none");
+	$list[] = PKG_FRR_LIST_NONE_VALUE;
 	$carplist = get_configured_vip_list("all", VIP_CARP);
 	foreach ($carplist as $vid => $vaddr) {
 		$vip = get_configured_vip($vid);
@@ -205,7 +207,7 @@ function frr_get_route_destinations() {
 	global $config;
 	$list = array();
 
-	$list[] = array("name" => "None", "value" => "none");
+	$list[] = PKG_FRR_LIST_NONE_VALUE;
 
 	/* Add non-disabled gateways to the list */
 	$gateways = return_gateways_array();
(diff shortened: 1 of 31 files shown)


net/pfSense-pkg-frr/files/usr/local/pkg/frr.inc
```diff
@@ -617,11 +649,12 @@ function frr_generate_config_vtysh() {
 		return;
 	}
 	if (empty($frr_conf['enable'])) {
-		/* FRR master switch is off. */
+		log_error("FRR vtysh: FRR master switch is off.");
```
jim-p, line 652: Why log this as an error? It's not a fatal problem, someone could be making configuration changes with the service disabled, and this would spam the log and not be helpful.

net/pfSense-pkg-frr/files/usr/local/pkg/frr.inc
```diff
@@ -674,10 +707,12 @@ function frr_generate_config_integrated() {
 		return;
 	}
 	if (empty($frr_conf['enable'])) {
-		/* FRR master switch is off. */
+		log_error("FRR Integrated: FRR master switch is off.");
```
jim-p, line 710: Same as last comment, seems unnecessary to log this.

net/pfSense-pkg-frr/files/usr/local/pkg/frr.inc
```diff
@@ -674,10 +707,12 @@ function frr_generate_config_integrated() {
...
 		return;
 	}
 
+	log_error("FRR Integrated: Rebuild configuration.");
```
jim-p, line 714: Same as above... Maybe wrap these in a debug setting that someone could opt into? Otherwise it seems way too chatty.

jim-p (changes requested):
Mostly looks OK, a couple minor things need addressed (see inline comments)

net/pfSense-pkg-frr/files/usr/local/pkg/frr.inc
bmhughes: I couldn't see another way to add some informational logging to the syslog to make it easily visible, but as you say it's become rather chatty. I've changed it to use function wrapping `syslog()` and added a setting to allow the user to set the package log level as you suggested, only two settings for now.

jim-p (approved):

theonemcdonald:
CARP status IP failover doesn't seem to be working on the latest code here. My config is an upgrade from the 0.6.whatever branch in the stable repo.

bmhughes:
> 
> 
> CARP status IP failover doesn't seem to be working on the latest code here. My config is an upgrade from the 0.6.whatever branch in the stable repo.

Doesn't surprise me with all the changes that have been made, can you verify your configuration and provide as much detail as possible in a redmine issue please.

theonemcdonald:
I uninstalled FRR v1.0.0, neutered my xml config by hand to remove any lingering FRR config, restored the backup, and reinstalled FRR. so this should be pretty bog vanilla, and the problem still persists.

Also, all of my passive interfaces on stub networks are not propagating to other neighbors. When I add an interface, I see something like this:

O  x.y.y.z/32 [110/0] is directly connected, lo0 inactive, weight 1, 00:03:35
C>* x.y.y.z/32 [0/1] is directly connected, lo0, 00:28:06

And this route is not propagating to other neighbors.

bmhughes:
Ok so at least it's likely not to be from a bad migration. 

But that's still not a lot of go off, I'm going to need a sanitised configuration at the very least and ideally a copy of the generated FRR conf file to debug your use case. Can you verify the generated file is what you'd expect or are there errors/missing statement? As it wouldn't be the first time I've ran across a bug in FRR. 

It's difficult to track this in a closed MR so if you could open a redmine issue with as much detail about your configuration as possible I can take a look at it.

PR #1036, opened 2021-02-08
pfBlockerNG IDN lists support. Implements #11295

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/11295
- [X] Ready for review

Adds IDN domains support for:
- DNSBL Whitelist
- TLD Exclusion List
- TLD Blacklist/Whitelist
- DNSBL Custom_List
- Python no AAAA List
- IPv4 Custom_List (domain/AS mode)

### DIFF
--- a/net/pfSense-pkg-pfBlockerNG-devel/files/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/net/pfSense-pkg-pfBlockerNG-devel/files/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -555,7 +555,7 @@ function pfb_strtolower($line) {
 
 // Function to decode alias custom entry box.
 // Default (False, True): Return as string with comments
-function pfbng_text_area_decode($text, $mode=FALSE, $type=TRUE) {
+function pfbng_text_area_decode($text, $mode=FALSE, $type=TRUE, $idn=FALSE) {
 
 	if ($mode) {
 		$custom = array();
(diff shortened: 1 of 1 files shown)


rbgarga:
@BBcan177 can you check this one?

net/pfSense-pkg-pfBlockerNG-devel/files/usr/local/pkg/pfblockerng/pfblockerng.inc
```diff
@@ -565,7 +565,24 @@ function pfbng_text_area_decode($text, $mode=FALSE, $type=TRUE) {
 	if (!empty($customlist)) {
 		foreach ($customlist as $line) {
 			if (substr(trim($line), 0, 1) != '#' && !empty($line)) {
-
+				if ($idn) {
+					if (strpos($line, '#') !== FALSE) {
+						$tmpline = preg_split('/(?=#)/', $line);
+						if (substr($tmpline[0], 0, 1) == '.') {
+							// idn_to_ascii() returns empty string if it starts with '.' 
+							$tmpline[0] = '.' . idn_to_ascii(ltrim($tmpline[0], '.'));
```
BBcan177, line 573: Need to validate that the idn_to_ascii() function doesn't return an empty line.

That condition would cause the whitelist to have a line with only the  ".", and that would probably whitelist all entries in DNSBL :)

net/pfSense-pkg-pfBlockerNG-devel/files/usr/local/pkg/pfblockerng/pfblockerng.inc
```diff
@@ -565,7 +565,24 @@ function pfbng_text_area_decode($text, $mode=FALSE, $type=TRUE) {
...
+						} else {
+							$tmpline[0] = idn_to_ascii($tmpline[0]);
+						}
+						$line = implode(' ', $tmpline);
+					} else {
+						if (substr($line, 0, 1) == '.') {
+							$line = '.' . idn_to_ascii(ltrim($line, '.'));
```
BBcan177, line 580: Same as above

net/pfSense-pkg-pfBlockerNG-devel/files/usr/local/pkg/pfblockerng/pfblockerng.inc
vktg: I see,
Fixed

net/pfSense-pkg-pfBlockerNG-devel/files/usr/local/pkg/pfblockerng/pfblockerng.inc
vktg: Fixed

BBcan177:
In the existing code, I have used the idn_to_ascii() function, however, I first validate the line with ctype_print() to avoid excessive calls to convert a line that doesn't need conversion.


```
if (!ctype_print($line)) {

    // Convert encodings to UTF-8
    $line = mb_convert_encoding($line, 'UTF-8',
            mb_detect_encoding($line, 'UTF-8, ASCII, ISO-8859-1'));

    $log = "\n  IDN converted: [ {$line} ]\t";
    $line = idn_to_ascii($line);
    if (!empty($line)) {
        pfb_logger("{$log} [ {$line} ]", 1);
    }
    ...
}
```

I also used mb_convert_encoding() to ensure the line in in UTF-8. In my testing this was needed. What do you think?

Also you didn't add validation (empty check) to the condition where there is no dot prefixed (wildcard entry).

To simplify this code, I would rather create a new function for this conversion, and use it in all instances of these conversions.

vktg:
> In the existing code, I have used the idn_to_ascii() function, however, I first validate the line with ctype_print() to avoid excessive calls to convert a line that doesn't need conversion.
That's nice, I added the `ctype_print()` check.

> I also used mb_convert_encoding() to ensure the line is in UTF-8. In my testing this was needed. What do you think?
I've never seen any issues with `idn_to_ascii()` without `mb_convert_encoding()`,
I've added it if you think it's necessary, but can you explain?

> Also you didn't add validation (empty check) to the condition where there is no dot prefixed (wildcard entry).
Fixed

> To simplify this code, I would rather create a new function for this conversion, and use it in all instances of these conversions.
I only changed `pfbng_test_area_decode()`, but not your loop, because it is too complex and easy to break.

rbgarga:
@BBcan177 please take a look if it looks OK now

PR #1043, opened 2021-02-16
Add support in acme for kdeenkhoorn's cPanel DNS plugin

bruce2409:
Updated the acme package to add [kdeenkhoorn's cPanel DNS plugin](https://github.com/kdeenkhoorn/dns_cpaneldns) as well as making changes to acme.inc to add the required env vars.


jim-p:
We do not accept direct submissions of new acme.sh providers. Submit this upstream to https://github.com/acmesh-official/acme.sh and once it is accepted there, then it can be added into this package naturally during the normal update cycle.

PR #1045, opened 2021-02-24
pfBlockerNG - Fix for clog

BBcan177:
* Fix for clog in pfSense 2.5


net/pfSense-pkg-pfBlockerNG/files/usr/local/www/pfblockerng/pfblockerng_alerts.php
```diff
@@ -427,7 +427,12 @@
 // Define common variables and arrays for report tables
 $fields_array	= $pfb_local = $pfb_localsub = $dnsbl_int = $local_hosts = array();
 
-$pfblines	= exec("/usr/local/sbin/clog {$filter_logfile} | {$pfb['grep']} -c ^");
+if (substr(trim(file_get_contents('/etc/version')), 0, 3) > '2.5') {
```
jim-p, line 430: Now that 21.02 and 2.5 are out, I don't think we'll be picking back any more updates to the 2.4.5 branch.
Also this check may fail as-is on Plus 21.02 anyhow.

Consider using another flag for testing this if you keep it, such as the presence of a file or function only found on 2.5.0. Also the `clog` binary will not be present on 2.5.0 systems.

net/pfSense-pkg-pfBlockerNG/files/usr/local/www/pfblockerng/pfblockerng_alerts.php
```diff
@@ -427,7 +427,12 @@
...
+	$pfblines = exec("{$pfb['grep']} -c ^ {$filter_logfile}");
```
jim-p, line 431: Consider using `dump_log()` instead of running grep or cat directly. This kind of command won't see inside rotated log files, so if a rotate happened recently it wouldn't find much of anything.

https://github.com/pfsense/pfsense/blob/master/src/etc/inc/syslog.inc#L581

net/pfSense-pkg-pfBlockerNG/files/usr/local/www/pfblockerng/pfblockerng_alerts.php
```diff
@@ -647,7 +652,11 @@ function conv_log_filter_lite($logfile, $nentries, $tail, $pfbdenycnt, $pfbpermi
 
 	if (file_exists($logfile)) {
 		// Collect filter.log entries
-		exec("/usr/local/sbin/clog {$logfile} | {$pfb['grep']} -v '\"CLOG\"\|\"\033\"' | {$pfb['grep']} 'filterlog:' | /usr/bin/tail -r -n {$tail}", $logarr);
+		if (substr(trim(file_get_contents('/etc/version')), 0, 3) > '2.5') {
+			exec("{$pfb['grep']} -v '\"CLOG\"\|\"\033\"' {$logfile} | {$pfb['grep']} 'filterlog:' | /usr/bin/tail -r -n {$tail}", $logarr);
```
jim-p, line 656: See above comments about `dump_log()` but also should you choose to keep going this way, you can remove the initial grep -v that removes `CLOG` and the other related value -- they cannot be present on 2.5.0 so filtering them is no longer useful.

jim-p (changes requested):

net/pfSense-pkg-pfBlockerNG/files/usr/local/www/pfblockerng/pfblockerng_alerts.php
BBcan177: The code if pfBlockerNG is so far behind that of pfBlockerNG-devel, that I am focusing all efforts on the devel branch, and want to make that the next pfBlockerNG release. I am not currently ready for that, but its in the not to distant future.

This commit is just a band-aid to get the Alerts Tab to function. I would really not want to spend a ton of time to make it perfect, if possible.

jim-p: OK, just so long as you aware that it isn't going to function well and can make users of the package aware of the limitation until it's fully addressed.

BBcan177: Yes Thanks!

jim-p (approved):

PR #1046, opened 2021-02-25
Snort ena(4) netmap support. Implements #11533

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/11533
- [X] Ready for review

add ena(4) to the list of INLINE mode (netmap) supported cards (pfSense 2.5/21.02)
see https://github.com/pfsense/FreeBSD-src/blob/devel-12/sys/dev/ena/ena.c


bmeeks8:
I am okay with this update. Thank you Viktor!

PR #1052, opened 2021-03-08
Fix OpenSSL OCSP command

csutcliff:
The OpenSSL OCSP command requires headers in the '\<key>=\<value>' format rather than '\<key> \<value>'.


csutcliff:
I've just seen this has been fixed properly in the -devel package, closing

PR #1064, opened 2021-05-14
pfSense-pkg-WireGuard v0.1.1

theonemcdonald:


vajonam:
@theonemcdonald, wireguard-kmod also needs to get pushed to 0503 (fixes a some reported few KPs)

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
```diff
@@ -0,0 +1,338 @@
...
+	}
+
+	$ret_html = "<i class=\"fa {$fa_icon} {$ret_class}\" style=\"vertical-align: middle;\" title=\"{$ret_title}\"></i>";
+
+	return $ret_html;
+
+}
+
+// Formats a given time difference in a human-friendly way
+function wg_human_time_diff($start_time, $end_time = null, $from_epoch = false) {
```
jim-p, line 133: Any particular reason you are doing this manually rather than using PHP's DateTime objects/functions which do this kind of thing natively?

Additionally working with epoch timestamps will fail on 32-bit ARM platforms in certain cases, though since these are all current timestamps and not future dates that's not a huge concern now.

I know it's not exactly the same, but look at what we do for certificates, for example at https://github.com/pfsense/pfsense/blob/master/src/etc/inc/certs.inc#L705 (for formatting) and https://github.com/pfsense/pfsense/blob/master/src/etc/inc/certs.inc#L2008 (for difference between times), and the related functions nearby.

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
```diff
@@ -0,0 +1,338 @@
...
+
+	if (isset($target_device)) {
+	
+		$s_icon = (isset($target_device['enabled']) && $target_device['enabled'] == 'yes') ? $icon_enabled : $icon_disabled;
+
+		$s_title = gettext($title);
+
+		$s_usepost = $usepost ? 'usepost' : '';
+
+		return "<a class=\"fa {$s_icon}\" title=\"{$s_title}\" href=\"{$href}\" {$s_usepost}></a>";
```
jim-p, line 216: Probably best to use `htmlspecialchars()` on the individual components here coming from variables. I checked the places calling it and most appear to be safe but that doesn't mean they'll always be safe. This kind of thing is how an XSS can easily slip in over time.

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
```diff
@@ -0,0 +1,338 @@
...
+
+		$tunnel = $wgg['tunnels'][$tun_idx];
+
+		$addresses = $tunnel['addresses']['row'];
+
+		if (!is_wg_tunnel_assigned($tunnel['name'])) {
+			
+			if (is_array($addresses) && count($addresses) > 0) {
+
+				$href = "vpn_wg_tunnels_edit.php?tun={$tunnel['name']}";
```
jim-p, line 239: This should get `htmlspecialchars()` here or in each place where it's used below.

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
```diff
@@ -0,0 +1,338 @@
...
+
+				$extras = count($addresses) - 1;
+
+				if ($extras > 0) {
+
+					$data_html = "<table><thead><th>Address</th><th style='padding-left: 10px;'>Description</th></thead><tbody>";
+		
+					foreach ($addresses as $address) {
+		
+						$data_html .= "<tr><td>{$address['address']}/{$address['mask']}</td><td style='padding-left: 10px;'>{$address['descr']}</td></tr>";
```
jim-p, line 249: The contents of these variables should be encoded with `htmlspecialchars()` before printing.

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
```diff
@@ -0,0 +1,338 @@
...
+		
+					}
+		
+					$data_html .= '</tbody></table>';
+		
+					$hint = "{$addresses[0]['address']} (+{$extras})";
```
jim-p, line 255: The contents of this string should be encoded with `htmlspecialchars()` before printing, either here or in the line below.

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
```diff
@@ -0,0 +1,338 @@
...
+		
+					$ret_html = "<a href=\"{$href}\" title=\"Interface Addresses\" data-toggle=\"popover\" data-trigger=\"hover focus\" data-content=\"{$data_html}\" data-html=\"true\">{$hint}</a>";
+		
+					return $ret_html;
+		
+				} else {
+		
+					$ret_html = "<a href=\"{$href}\" title=\"{$addresses[0]['descr']}\">{$addresses[0]['address']}/{$addresses[0]['mask']}</a>";
```
jim-p, line 263: The contents of the variables should be encoded with `htmlspecialchars()` before printing if they haven't before reaching this point.

Though you can do it earlier to avoid having to call it on each output, it's safest to do it as close as possible to the output.

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
```diff
@@ -0,0 +1,338 @@
...
+
+				return '(none)';
+
+			}
+
+		} else {
+
+			$wg_pfsense_if = wg_get_pfsense_interface_info($tunnel['name']);
+
+			return "<i class=\"fa fa-sitemap\" style=\"vertical-align: middle;\"></i><a style=\"padding-left: 3px\" href=\"../../interfaces.php?if={$wg_pfsense_if['name']}\">{$wg_pfsense_if['descr']} ({$wg_pfsense_if['name']})</a>";
```
jim-p, line 279: The contents of these variables should be encoded with `htmlspecialchars()` before printing.

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
```diff
@@ -0,0 +1,338 @@
...
+
+			$extras = count($allowedips) - 1;
+
+			if ($extras > 0) {
+
+				$data_html = "<table><thead><th>Subnet</th><th style='padding-left: 10px;'>Description</th></thead><tbody>";
+	
+				foreach ($allowedips as $address) {
+	
+					$data_html .= "<tr><td>{$address['address']}/{$address['mask']}</td><td style='padding-left: 10px;'>{$address['descr']}</td></tr>";
```
jim-p, line 308: The contents of these variables should be encoded with `htmlspecialchars()` before printing.

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
```diff
@@ -0,0 +1,338 @@
...
+	
+				}
+	
+				$data_html .= '</tbody></table>';
+	
+				$hint = "{$allowedips[0]['address']} (+{$extras})";
+	
+				$ret_html = "<a href=\"{$href}\" title=\"Allowed IPs\" data-toggle=\"popover\" data-trigger=\"hover focus\" data-content=\"{$data_html}\" data-html=\"true\">{$hint}</a>";
```
jim-p, line 316: The contents of these variables should be encoded with `htmlspecialchars()` before printing.

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
```diff
@@ -0,0 +1,338 @@
...
+	
+				return $ret_html;
+	
+			} else {
+	
+				$ret_html = "<a href=\"{$href}\" title=\"{$allowedips[0]['descr']}\">{$allowedips[0]['address']}/{$allowedips[0]['mask']}</a>";
```
jim-p, line 322: The contents of these variables should be encoded with `htmlspecialchars()` before printing.

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_install.inc
```diff
@@ -53,72 +295,284 @@ function wg_deinstall() {
...
 
-		update_status("done.\n");
+	$tunnels_to_destroy = wg_get_real_ifs();
 
-	} else {
+	// Tear down interfaces
+	foreach ($tunnels_to_destroy as $tunnel) {
 
-		update_status("  Removing WireGuard configuration settings... ");
+		pfSense_interface_destroy($tunnel);
```
jim-p, line 353: Some extra whitespace at the end of this line that needs removed

net/pfSense-pkg-WireGuard/files/usr/local/www/wg/status_wireguard.php
```diff
@@ -58,35 +58,125 @@
...
+				<th><?=gettext('TX')?></th>
+			</thead>
+<?php
+	foreach ($a_devices as $device_name => $device):
+?>
+			<tbody>	
+				<tr class="tunnel-entry">
+					<td>
+						<?=wg_interface_status_icon($device['status'])?>
+						<a href="vpn_wg_tunnels_edit.php?tun=<?=$device_name?>"><?=htmlspecialchars($device_name)?>
```
jim-p, line 97: `$device_name` should be encoded in the link as well as when printed.

net/pfSense-pkg-WireGuard/files/usr/local/www/wg/vpn_wg_peers.php
```diff
@@ -96,32 +105,25 @@
 						<th><?=gettext("Description")?></th>
 						<th><?=gettext("Public key")?></th>
 						<th><?=gettext("Tunnel")?></th>
-						<th><?=gettext("Peer Address")?></th>
 						<th><?=gettext("Allowed IPs")?></th>
-						<th><?=gettext("Endpoint").' : '.gettext("Port")?></th>
+						<th><?=wg_format_endpoint(true)?></th>
```
jim-p, line 109: This needs `htmlspecialchars()` here as well.

net/pfSense-pkg-WireGuard/files/usr/local/www/wg/vpn_wg_peers.php
```diff
@@ -147,14 +149,19 @@
 </form>
 
 <script type="text/javascript">
-//<![CDATA[
+//<![CDATA[,
```
jim-p, line 152: Is that `,` intentional and necessary here?

net/pfSense-pkg-WireGuard/files/usr/local/www/wg/vpn_wg_tunnels.php
```diff
@@ -159,9 +147,8 @@
 									<tr class="peerbg_color">
 										<th><?=gettext("Description")?></th>
 										<th><?=gettext("Public key")?></th>
-										<th><?=gettext("Peer Address")?></th>
 										<th><?=gettext("Allowed IPs")?></th>
-										<th><?=gettext("Endpoint").' : '.gettext("Port")?></th>
+										<th><?=wg_format_endpoint(true)?></th>
```
jim-p, line 151: This also needs `htmlspecialchars()`

net/pfSense-pkg-WireGuard/files/usr/local/www/wg/vpn_wg_tunnels_edit.php
```diff
@@ -293,7 +289,7 @@
 			'fa-trash'
 		))->addClass('btn-warning btn-sm');
 	
-		$section->add($group);
+		$section->add($group);
```
jim-p, line 292: Extra whitespace at the end of this line, needs removed.

net/pfSense-pkg-WireGuard/files/usr/local/www/wg/vpn_wg_tunnels_edit.php
```diff
@@ -307,39 +303,41 @@
...
-	$ifdescr = get_configured_interface_with_descr(true);
-	
-	$ifname = $iflist[$pconfig['name']];
-	$iffriendly = $ifdescr[$ifname];
+	$wg_pfsense_if = wg_get_pfsense_interface_info($pconfig['name']);
 
 	$section->addInput(new Form_StaticText(
 		'Assignment',
-		"<a href='../../interfaces_assign.php'>{$iffriendly} ({$ifname})</a>"
+		"<i class='fa fa-sitemap' style='vertical-align: middle;'></i><a style='padding-left: 3px' href='../../interfaces_assign.php'>{$wg_pfsense_if['descr']} ({$wg_pfsense_if['name']})</a>"
```
jim-p, line 310: Variables need `htmlspecialchars()`

net/pfSense-pkg-WireGuard/files/usr/local/www/wg/vpn_wg_tunnels_edit.php
```diff
@@ -307,39 +303,41 @@
...
 	));
 
 	$section->addInput(new Form_StaticText(
 		'Interface',
-		"<a href='../../interfaces.php?if={$ifname}'>Interface Configuration</a>"
+		"<i class='fa fa-ethernet' style='vertical-align: middle;'></i><a style='padding-left: 3px' href='../../interfaces.php?if={$wg_pfsense_if['name']}'>Interface Configuration</a>"
```
jim-p, line 315: Variables need `htmlspecialchars()`

net/pfSense-pkg-WireGuard/files/usr/local/www/wg/vpn_wg_tunnels_edit.php
```diff
@@ -307,39 +303,41 @@
...
 	));
 
 	$section->addInput(new Form_StaticText(
 		'Firewall Rules',
-		"<a href='../../firewall_rules.php?if={$ifname}'>Firewall Configuration</a>"
+		"<i class='fa fa-shield-alt' style='vertical-align: middle;'></i><a style='padding-left: 3px' href='../../firewall_rules.php?if={$wg_pfsense_if['name']}'>Firewall Configuration</a>"
```
jim-p, line 320: Variables need `htmlspecialchars()`

net/pfSense-pkg-WireGuard/files/usr/local/www/wg/vpn_wg_tunnels_edit.php
```diff
@@ -351,25 +349,24 @@
...
 	<div id="mainarea" class="table-responsive panel-body">
 		<table id="peertable" class="table table-hover table-striped table-condensed" style="overflow-x: 'visible'">
 			<thead>
 				<tr>
 					<th><?=gettext("Description")?></th>
 					<th><?=gettext("Public key")?></th>
-					<th><?=gettext("Peer Address")?></th>
 					<th><?=gettext("Allowed IPs")?></th>
-					<th><?=gettext("Endpoint").' : '.gettext("Port")?></th>
+					<th><?=wg_format_endpoint(true)?></th>
```
jim-p, line 369: This also needs `htmlspecialchars()`

jim-p (changes requested):

theonemcdonald:
I believe I have addressed all of Jim's points.

Thanks

net/pfSense-pkg-WireGuard/Makefile
```diff
@@ -29,10 +29,10 @@ do-install:
 	${MKDIR} ${STAGEDIR}${PREFIX}/pkg/wireguard
 
 	${MKDIR} ${STAGEDIR}${PREFIX}/www/shortcuts
-	${MKDIR} ${STAGEDIR}${PREFIX}/www/wg
+	${MKDIR} ${STAGEDIR}${PREFIX}/www/wg/js
 
 	${MKDIR} ${STAGEDIR}${DATADIR}
-
+
```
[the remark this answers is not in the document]
rbgarga: This extra blank line with a TAB should be removed

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_globals.inc
```diff
@@ -118,6 +133,34 @@ function array_unset_value(&$array, $parents) {
 
 }
 
+/*
+ * This is a generalized implementation of init_config_arr
```
rbgarga, line 137: Out of curiosity, is it not possible to use `init_config_arr()`?

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
```diff
@@ -0,0 +1,389 @@
+<?php
+/*
+ * wg_globals.inc
```
rbgarga, line 3: Wrong filename

net/wireguard-kmod/Makefile
```diff
@@ -1,5 +1,5 @@
 PORTNAME=	wireguard-kmod
-PORTVERSION=	0.0.20210502
+PORTVERSION=	0.0.20210503
```
rbgarga, line 2: I already merged origin/main from upstream, please revert any change to this file

net/wireguard-kmod/distinfo
```diff
@@ -1,3 +1,2 @@
-TIMESTAMP = 1620052580
-SHA256 (wireguard-freebsd-0.0.20210502.tar.xz) = cd59a0a711308083e40ac5e7e26abc63c8b1f4615a9d66f73f2398388b8e7b9b
-SIZE (wireguard-freebsd-0.0.20210502.tar.xz) = 50464
+SHA256 (wireguard-freebsd-0.0.20210503.tar.xz) = 96b2bceadc05dac2f026e203d1c0d25f114491afc8fbfc8d6f71c3de7f4a22f1
+SIZE (wireguard-freebsd-0.0.20210503.tar.xz) = 50548
```
rbgarga, line 2: I already merged origin/main from upstream, please revert any change to this file

rbgarga (changes requested):

net/wireguard-kmod/distinfo
theonemcdonald: Done

net/wireguard-kmod/Makefile
theonemcdonald: Done

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
theonemcdonald: Done

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_globals.inc
theonemcdonald: I touch the wireguard XML config by way of $wgg['config'], $wgg['tunnels'], and $wgg['peers'] which these reference $config in wg_globals.inc. Instead of initializing when necessary through $config, generalizing init_config_arr() allows me to perform the same action but on an arbitrary target (i.e. one of the $wgg targets above).

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
```diff
@@ -123,50 +132,75 @@ function wg_handshake_status_icon($latest_handshake = 0, $fa_icon = "fa-handshak
...
 
-		$num_of_units = floor($diff_time / $unit);
+		$num_of_units = floor($diff_time_seconds / $unit);
 
 		$plural = ($num_of_units > 1) ? 's' : '';
+
+		$tense = ($diff_time->invert) ? 'ago' : 'from now';
 		
-		return "{$num_of_units} {$text}{$plural}";
+		return "{$num_of_units} {$text}{$plural} {$tense}";
```
jim-p: Not terribly fond of this part. We have a similar function already, `convert_seconds_to_dhms()` https://github.com/pfsense/pfsense/blob/master/src/etc/inc/pfsense-utils.inc#L1575 seems like that would be a better fit, or you could leverage the PHP `DateInterval::format` method at https://www.php.net/manual/en/dateinterval.format.php which can output closer to what you are doing manually without reinventing the wheel.

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_service.inc
```diff
@@ -95,7 +93,7 @@ function wg_service_stop() {
 function wg_is_service_running() {
 	global $wgg;
 
-	if (!($h_lock = @fopen($wgg['pid_path'],'r')) || !file_exists($wgg['pid_path'])) {
+	if (!($h_lock = @fopen($wgg['pid_path'], 'a+')) || !file_exists($wgg['pid_path'])) {
```
jim-p, line 96: Why the mode switch here? Using `a+` will cause the file to be created if it doesn't exist, which doesn't seem necessary if it's only testing that the service is running.

net/pfSense-pkg-WireGuard/files/usr/local/www/wg/status_wireguard.php
```diff
@@ -1,245 +1,246 @@
...
+			
+			if (count($device['peers']) > 0):
+?>
+					<td colspan="9">
+						<table class="table table-hover table-condensed">
+							<thead>
+								<th><?=gettext('Peer')?></th>
+								<th><?=gettext('Latest Handshake')?></th>
+								<th><?=gettext('Public Key')?></th>
+								<th><?=wg_format_endpoint(true)?></th>
```
jim-p, line 122: This one still needs `htmlspecialchars()`

net/pfSense-pkg-WireGuard/files/usr/local/www/wg/vpn_wg_tunnels_edit.php
```diff
@@ -289,7 +289,7 @@
 			'fa-trash'
 		))->addClass('btn-warning btn-sm');
 	
-		$section->add($group);		
+		$section->add($group);
```
jim-p, line 292: There is still some trailing whitespace on this line

jim-p (changes requested):

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_service.inc
theonemcdonald: good catch, not necessary.

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
theonemcdonald: What I'm doing here is more closely aligning to the handshake format displayed in various WireGuard clients. From what I've seen, this is likely as good as it can get in order to get this particular output style. I'm using DateTime to derive the number of seconds and then some basic string manipulation to get the desired final output. For what it's worth, this is not too far off from how `wg(8)` generates the latest handshake time as well.

vajonam: https://git.zx2c4.com/wireguard-tools/tree/src/show.c#n129

net/pfSense-pkg-WireGuard/files/usr/local/www/wg/vpn_wg_tunnels_edit.php
theonemcdonald: Fixed.

net/pfSense-pkg-WireGuard/files/usr/local/pkg/wireguard/wg_guiconfig.inc
jim-p: I get why they do it that way in C, but in PHP it makes less sense, as the DateTime library has methods for handling exactly that type of output and formatting. There are better examples on the php.net page I linked above which do similar things without hardcoding the various time values.

vajonam: Since contributed the original code here I will comment, I took a look at  those examples. There all use very similar code to get exactly what we want to do. This is another example https://www.php.net/manual/en/dateinterval.format.php#102271 that uses almost identical logic

theonemcdonald: I guess the question here is really, do we want latest handshake times to be displayed in the "WireGuard" way (i.e. how WireGuard clients display it) or in a more  "pfSense" way (i.e. how these types of times are displayed in other parts of the UI). I'm obviously more inclined to the former. I'm not immediately seeing how the `DateInterval::format` method can be used as a replacement for implementing the former as this method doesn't return the total seconds and will always require maths to convert and add-up the various denominations.

theonemcdonald: @jim-p I'll look at the example you linked more closely in a few hours. A cursory glance looks promising, because what we are effectively doing is only showing the 'largest' denomination and that example is pretty easy to adjust to produce that. Thanks.

theonemcdonald: Alright, spent some quality time on this tonight. I think we are square on this now.

net/pfSense-pkg-WireGuard/files/usr/local/www/wg/status_wireguard.php
```diff
@@ -1,129 +1,240 @@
...
```
jim-p, line 116: This one lost the `htmlspecialchars()` in one of your last commits

jim-p (changes requested):
Just one last little thing and I think it's good for now. Thanks!

net/pfSense-pkg-WireGuard/files/usr/local/www/wg/status_wireguard.php
theonemcdonald: fixed

rbgarga:
Manually merged!  Thanks!!!

PR #1083, opened 2021-07-17
Fixes #7039 - SSL checks with intermediate CAs

dractyl:
Includes the root CA in the certificate file when specifying an intermediate cert, assuming the whole chain exists in the main certificate store. If specifying a root cert directly, this should work the same as haproxy_write_certificate_crt based on the code. Tested with intermediate cert.


jim-p:
Redmine: https://redmine.pfsense.org/issues/7039

jim-p (changes requested):

dractyl:
Done.

jim-p (approved):

dneuhaeuser:
unfortunately this change never found its way to the current packages, neither haproxy-stable 0.63_2 nor haproxy-devel 0.63_2 !?

PR #1115, opened 2021-10-16
security/pfSense-pkg-acme: Add support for preferred chain

dekimsey:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/11163
- [x] Ready for review

### DIFF
--- a/security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc
+++ b/security/pfSense-pkg-acme/files/usr/local/pkg/acme/acme.inc
@@ -1590,6 +1590,7 @@ function & get_certificate($name) {
 			$acmesh->dnssleep = $certificate['dnssleep'];
 			$acmesh->challengealias = $certificate['challengealias'];
 			$acmesh->ocspstaple = $certificate['ocspstaple'];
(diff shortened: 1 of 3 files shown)


vktg:
I think it's better to use dropdown-menu (`Form_Select()`) for the `preferredchain` option
There are only two Preferred Chains entries:

|   Name                        |  Default |
|-------------------------------|----------|
|   (STAGING) Pretend Pear X1   |   No     |
|   (STAGING) Bogus Broccoli X2 |   Yes    |

dekimsey:
I initially considered that route and decided against it in my implementation. The chains would be provider specific, it would require the maintainers to keep the list of the intermediates up to date (maintenance burden of producing timely updates) and it would tie the app tightly to Let's Encrypt. Since ACME is a protocol any CA could leverage. I didn't think it made for a good fit. By leaving the field as a text string and exposing acme.sh's interface as is, I felt it did the job without incurring additional complications.

jim-p:
Yes, that should cover the concerns I had so far. Also, make sure the chain selection drop-down has a "Default" option that is selected by default and does not add the option to the command line.

PR #1119, opened 2021-11-05
Add support for bi-directional flow in softflowd

vpiserchia:
This patch allows to configure bi-directional flows when IPFIX version is selected. 

Associated Redmine ticket: https://redmine.pfsense.org/issues/12507


jim-p (changes requested):

vpiserchia:
Thanks for pointing it out.

jim-p (approved):

PR #1138, opened 2022-01-21
pfBlockerNG: move python mode file to pkg folder. Fixes #12706

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/12706
- [X] Ready for review


vktg:
@BBcan177 please check

BBcan177:
I would also suggest adding code to pfblockerng dnsbl.php to move the files once Python mode is enabled. 

Did you test this on a reboot with Ramdisks enabled? 

This should be thoroughly tested in multiple scenarios to ensure its moving the files early on. If these files are not in the /var/unbound folder they will cause Unbound to not load.

vktg:
Finally tested in all scenarios and added some fixes

PR #1140, opened 2022-01-24
Add igc(4) to the list of Snort INLINE mode (iflib/netmap) supported cards. Feature #12719

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/12719
- [X] Ready for review


bmeeks8:
I have no issue with this update -- Approved.  Thanks @vktg!

PR #1142, opened 2022-01-29
freeRADIUS conditional modules

ccope-netgate:
Fix for https://redmine.pfsense.org/issues/12742


jim-p (approved):
Looks OK, similar fix to the one used on https://redmine.pfsense.org/issues/10197

Syntax ref: https://github.com/FreeRADIUS/freeradius-server/blob/master/doc/antora/modules/raddb/pages/mods-available/index.adoc#conditional-modules

PR #1144, opened 2022-02-16
The NiX Spam feed has moved from www.dnsbl.manitu.net to www.nixspam.net

dennypage:
Attempts to access the feed file via the old URL return a 301 (Moved Permanently), however the redirect goes to the base of the new domain (HTML text) rather than to the feed file. The result is that the download succeeds, but the downloaded
file does not contain the list of IP addresses.


BBcan177:
Thanks for the PR Denny!

Can you add a "**past_urls**" JSON entry for the old URL? This will be used in future for an Automated Feed management feature. As a note, there is also a new "**discontinued**" JSON entry for when a Feed has been discontinued.

ie:
https://github.com/pfsense/FreeBSD-ports/blob/devel/net/pfSense-pkg-pfBlockerNG-devel/files/usr/local/www/pfblockerng/pfblockerng_feeds.json#L93

Thanks

dennypage:
"past_urls" section added.

Hope things are well with you Anthony!

BBcan177:
All is well... Thank you for the patch!

PR #1147, opened 2022-03-04
Acme cpanel patch

brownowski:
Fixed an issue having uppercase letters in name field for cPanel_Hostname


jim-p:
I just fixed this myself since it also needed a version bump, it will be available once the packages build.

PR #1148, opened 2022-03-20
Fix NMap timeout when started from GUI

PhilZ-cwm6:
Redmine: https://redmine.pfsense.org/issues/12963

NMap package cannot actually run from gui because of nginx timeout
This patch adds the following features:
- run nmap in background and output results to /root/nmap.result filr
- do not start a new nmap session from gui until last one is completed
- add a results tab to get the last log output from nmap
- show the time of last nmap scan completion (start time is displayed by actual command output)


jim-p:
Looking at https://nmap.org they seem to prefer `Nmap` or `nmap`, I don't see anywhere they capitalize it as `NMap`. We should standardize around their preferred format, so using `Nmap` is best. There are existing references in both styles in the package but this PR adds a lot more, so it's a good time to correct it and make them consistent.

jim-p (changes requested):
Looks OK now, though it needs a version bump:

It would also be very helpful if you could squash the commits into a single entry: https://stackoverflow.com/a/5668050

PhilZ-cwm6:
> Looks OK now, though it needs a version bump:
> 
> 
> 
> It would also be very helpful if you could squash the commits into a single entry: https://stackoverflow.com/a/5668050

Thank you for your directions
I squashed all commits in a single one + PORTREVISION increase

Is it ok as it is or should I push a new pull request from the squashed branch ?

security/pfSense-pkg-nmap/Makefile
```diff
@@ -34,6 +34,8 @@ do-install:
 		${STAGEDIR}${PREFIX}/pkg
 	${INSTALL_DATA} ${FILESDIR}/etc/inc/priv/nmap.priv.inc \
 		${STAGEDIR}/etc/inc/priv
+    ${INSTALL_DATA} ${FILESDIR}${PREFIX}/www/nmap_view_results.php \
+        ${STAGEDIR}${PREFIX}/www
```
jim-p, line 38: These lines have leading spaces but must be tabs, otherwise make fails.

jim-p (changes requested):
I left a note about one change that needs made or else the port fails to compile, otherwise it looks OK.

> Is it ok as it is or should I push a new pull request from the squashed branch ?

What you have here is fine, when you did the squash+push it replaced the old commits with the new single squashed commit and it's fine from the same PR. After fixing the Makefile issue I noted if you squash again then it'll be ready to go.

Thanks!

security/pfSense-pkg-nmap/Makefile
PhilZ-cwm6: Sorry, I forgot nano was set to spaces instead of tabs, should have used ee
I usually work in VS, but the repo clone fails because of japanese/prn under windows:
https://github.com/freebsd/freebsd-ports/pull/122

PR #1154, opened 2022-03-28
Nmap: allow more scan and log control

PhilZ-cwm6:
nmap command:
- replace deprecated scan options (-sP)
- use nmap built-in -oN summarized output format for results log
- allow custom scan options
- add support for IPsec and localhost interfaces
- for -d option to no more allow deprecated options

GUI:
- use same tab for all operations (harmonize gui with Packet Capture)
- allow stop of running nmap processes started by gui
- display error messages and command errors in a results error field
- allow refresh results without stopping scan
- default append to results log
- properly summarized nmap log using -oN option
- properly display the full command in output file using -oN option
- hide detailed info in spannable tool tips


PhilZ-cwm6:
Further discussion in https://redmine.pfsense.org/issues/12963

security/pfSense-pkg-nmap/files/usr/local/www/nmap_scan.php
```diff
@@ -0,0 +1,468 @@
...
+$scan_types = array(
+	'syn' => gettext('SYN'),
+	'connect' => gettext('TCP connect()'),
+	'icmp' => gettext('Ping'),
+	'udp' => gettext('UDP'),
+	'arp' => gettext('ARP (directly connected networks only!)')
+);
+
+function nmap_get_running_process(){
+	$processcheck = (trim(shell_exec("/bin/ps axw -O pid= | /usr/bin/grep '/usr/local/bin/nmap' | /usr/bin/grep '{$fn}' | /usr/bin/egrep -v '(pflog|grep)'")));
```
jim-p, line 62: Seems like this could be simplified a bit. Three greps chained when it could probably be reduced to one, such as `/usr/bin/grep '/usr/local/bin/[n]map.*{$fn}'`

security/pfSense-pkg-nmap/files/usr/local/www/nmap_scan.php
```diff
@@ -0,0 +1,468 @@
...
+	return $processcheck;
+}
+
+// only accept alphanumeric, spaces (040) and '.,-_' chars, and do not allow '-o? ' for output files
+function valid_input($str) {
+	$valid_str = (preg_match('/^[a-z0-9\040\.,\-_]+$/i', $str) === 1) && (preg_match('/-o.\040/', $str) === 0);
```
jim-p, line 68: If any text run through this could be an IPv6 address then you also need to allow `:`.

security/pfSense-pkg-nmap/files/usr/local/www/nmap_scan.php
```diff
@@ -0,0 +1,468 @@
...
+						break;
+					case 'udp':
+						$nmap_options .= " -sU";
+						break;
+					case 'arp':
+						$nmap_options .= " -sn -PR";
+						break;
+				}
+			} elseif (valid_input($custom_scantype)) {
+					$nmap_options .= " " . $custom_scantype;
```
jim-p, line 132: I know you're severely limiting the input allowed here but it still seems like a bad idea to allow arbitrary parameters to be passed.

If someone knows that much about nmap parameters they aren't likely to be using the GUI for this. Possible, but unlikely.

security/pfSense-pkg-nmap/files/usr/local/www/nmap_scan.php
```diff
@@ -0,0 +1,468 @@
...
+			// prevent using source stylesheets for xls output (not really needed)
+			$nmap_options .= " -oN {$fp}{$fn} --append-output --no-stylesheet";
+
+			if (!empty($interface)) {
+				$nmap_options .= " -e " . get_real_interface($interface);
+			}
+
+			if (!empty($custom_targetspecs)) {
+				if (valid_input($custom_targetspecs)) {
+					$nmap_options .= " " . $custom_targetspecs;
```
jim-p, line 166: Same note as above about arbitrary parameters.

jim-p (changes requested):
A few suggestions for things to change.

If you want to keep the custom options GUI, you could add a second privilege entry which grants access to the custom options. Look at how the OpenVPN advanced options are handled in a similar manner.

That way an admin could give someone basic access to the nmap package without also granting them the ability to enter custom parameters. It must be checked on the frontend (to hide the GUI options) *and* on the backend to throw an input error if a user without the privilege tries to submit the options.

security/pfSense-pkg-nmap/files/usr/local/www/nmap_scan.php
PhilZ-cwm6: > I know you're severely limiting the input allowed here but it still seems like a bad idea to allow arbitrary parameters to be passed.
> 
> If someone knows that much about nmap parameters they aren't likely to be using the GUI for this. Possible, but unlikely.

> Same note as above about arbitrary parameters.


Thank you a lot for the detailed input.
Like mentioned in the redmine comment, I am aware that there is a potential risk if nmap ever changes the options during an update, or in case of nmap vulnerabilities. Adding changes for managing user permissions on the custom options seems overkill in fact for just nmap. As you say, any admin user could just drop to ssh/shell in that case.

If you agree, I'd go with simply removing the unsafe custom options fields. I could add a few additional selectable common options and maybe one or two more scan options. The hosts field could also be adjusted to allow comma or space separated IP/hosts.

Also, I'd like your input about this part:
https://github.com/pfsense/FreeBSD-ports/blob/654af55a3208ea8c9aacf8f0a2618ab52f74206f/security/pfSense-pkg-nmap/files/usr/local/www/nmap_scan.php#L86
```
	if ($_POST['startbtn'] != "") {
		$action = gettext("Start");

		// check for input errors
		if (empty($hostname)) {
```

Currently, in "Packet Capture" module, any button like "Clear" or "View Results" will check for input errors and do not allow processing.
I went with the option to only check for fields integrity when doing a scan. The fields are still processed and thus preserved after clicking View, Clear or Refresh buttons. There are no additional risks, especially if custom options are removed. Since they are never used except for preserving the form print, I found it odd that click "clear" buttons fails if invalid entries are entered.

Also, currently Packet Capture stops if page is refreshed. If the changes made here are fine, I can quickly modify Packet Capture GUI to:
- properly run in background even if user refreshes the page
- implement the Refresh results button while a scan is running
- remove check forms integrity when user clicks other buttons except Scan

jim-p: The packet capture page doesn't stop on refresh, it runs until the capture stops -- when it hits an upper limit on packets. If you start a long capture, say for lots of packets with a filter that doesn't match a lot, then it will keep running for a while. You'll still get a stop button even if you navigate away and come back.

PhilZ-cwm6: > The packet capture page doesn't stop on refresh, it runs until the capture stops -- when it hits an upper limit on packets. If you start a long capture, say for lots of packets with a filter that doesn't match a lot, then it will keep running for a while. You'll still get a stop button even if you navigate away and come back.

Ah, yes, you're right, it was because I tested on WAN and it reached the 100 hits in a few seconds

I can remove the custom fields and submit a new commit ?

PhilZ-cwm6: It is safer to remove the custom fields, if ever nmap changes again the -o option for example as they previously did
The below commits look fine to me and I tested all possibilities I could figure

jim-p (approved):

PR #1157, opened 2022-04-02
Update manifest.inc

luckman212:
as per @jim-p 's [redmine comment](https://redmine.pfsense.org/issues/12834#note-12), patch [`fd30b9eda98900eb5f1a0c37d600ef4deaf582ad`](https://github.com/pfsense/FreeBSD-ports/blob/1a4f1fdbd14484e4ea4630fe4cd16ac777a32f5a/sysutils/pfSense-pkg-System_Patches/files/usr/local/pkg/patches/manifest.inc#L68) (captive portal workaround) no longer applies to 2.7 / 22.05 since there's a new method being used


jim-p:
It's still relevant at this moment, we have not yet committed the new work. It's just going to be solved differently on there so we haven't committed the workaround because we'd only have to revert it later. Being dev snapshots, they are not expected to be stable at all times.

PR #1159, opened 2022-04-15
pfSense-pkg-cellular - add support for Quectel EC25 and symbolic link for NMEA port

kkondyl:
pfSense-pkg-cellular is updated to support Quectel EC25 and create a symbolic link for NMEA (GPS) port for Quectel EC25 and Simcom modems.


kkondyl:
https://redmine.pfsense.org/issues/13063

net/pfSense-pkg-cellular/files/etc/devd/cellular.conf
```diff
@@ -45,3 +51,18 @@ notify 1000 {
         action "%%PYTHON_CMD%% /usr/local/bin/cellular_dev.py -r";
 };
 
+# Quectel - EC25 LTE modem
+# ********
+attach 1000 {
+        match "vendor"          "0x2c7c";
+        match "product"         "0x0125";
+        action "/usr/local/bin/python3.8 /usr/local/bin/cellular_dev.py -a -d $device-name -m 2c7c0125";
```
jim-p, line 59: This should be using `%%PYTHON_CMD%%` and not the path to python directly.

net/pfSense-pkg-cellular/files/etc/devd/cellular.conf
```diff
@@ -45,3 +51,18 @@ notify 1000 {
...
+};
+
+notify 1000 {
+        match "vendor"          "0x2c7c";
+        match "product"         "0x0125";
+        match "type"            "DETACH";
+        match "port"            "[0-9]";
+        action "/usr/local/bin/python3.8 /usr/local/bin/cellular_dev.py -r";
```
jim-p, line 67: This should be using `%%PYTHON_CMD%%` and not the path to python directly.

jim-p (changes requested):

jim-p:

net/pfSense-pkg-cellular/files/etc/devd/cellular.conf
kkondyl: thanks for the comment. Updated

net/pfSense-pkg-cellular/files/etc/devd/cellular.conf
kkondyl: thanks for the comment. Updated

jim-p (approved):

PR #1177, opened 2022-07-03
ACME: implement IONOS DNS API Fix for API Version 1.0.1

afe1520860:
IONOS released a new API Version, the integration in ACME broke with that.
A Fix is already in the ACME-Repo available: https://github.com/acmesh-official/acme.sh/pull/4102

Could someone please take care and import this to the freebsd-port?


jim-p:
We will pick this up during the next upstream merge, not from a PR. Thanks!

PR #1186, opened 2022-09-07
pfblockerNG-devel: Fix spelling of 'compatible' in documentation and comments

apurvis:
cc @BBcan177


BBcan177:
Hello, you also included changes to another port in this PR ( textproc/pdfgrep/Makefile). Best to remove that change. Thanks!

PR #1187, opened 2022-09-10
pfSense-pkg-Avahi: Fix spelling of "bonjour" in pkg description

cvandesande:


jim-p:
I fixed this typo manually instead of using a merge, thanks!

PR #1188, opened 2022-09-12
Fix: Bug #13485: Interfaces are not listed correctly in pfBlockerNG-devel when selecting the cURL interface

chamilton-ccn:
Fix for [bug #13485](https://redmine.pfsense.org/issues/13485)


marcos-ng:
Change looks good to me. This can be merged before https://redmine.pfsense.org/issues/13156#note-20 to avoid bumping the version twice.

linnemannr:
Merged, pending sync

jim-p:
Merged manually

PR #1203, opened 2022-12-08
pfBlockerNG_devel_3.1.0_7

BBcan177:
**Changelog**:

* Additions/Changes to SafeSearch domains
* DNSBL SafeSearch in Unbound Python mode, Pixabay and DuckDuckGo use Local-zone/local-data (Unbound mode) due to DNSSEC limitations in the python implementation. 
* Add more cURL Error messages to download failure logging
* Changed ASN code from PHP cURL to a shell function.
* Remove any DNSBL Original files that are unassociated
* Fix an issue when DNSBL Category is enabled but no other DNSBL Groups are defined
* IP/DNBSL Pre/Post script functionality was added awhile back, but I forgot to edit the pkg-plist and Makefile for the pre-defined AWS Scripts.
* Improve the Alerts/Reports Tab Filter functionality
* Some Alert/Report Settings options were not shown when DNSBL is disabled.
* Fix Log Tab issue for Safari Browser
* Fix for https://redmine.pfsense.org/issues/13596
* Create the IP Reputation Tab when MaxMind is not enabled
* MaxMind GeoIP (when enabled with key) is downloaded each Friday instead of once per month
* Remove some discontinued Feeds
* Under the hood changes and Hardening


jim-p (approved):
I've been reviewing these changes with @BBcan177 as he's been doing development, seems to be in good enough shape to merge at this point.

PR #1206, opened 2022-12-15
pfBlockerNG-devel v3.1.0 9

BBcan177:
Changelog:

* Fix issue causing Resolver issues in certain cases
* Fix issue with Header Field Auto-Sort
* Fix issue with 'Whois' Format

* Fix issue saving URLs in the IPv4/v6/DNSBL Tab. 
Each save would repeat the htmlspecialchars() encoding resulting in duplicate ";amp" tags being added. 
Please review any IP Aliaes/ DNSBL Groups that have been saved since the last update and manually remove any of these erroneous tags. 
I assume that these will still download ok, so its more cosmetic, and these Feeds will show in the Feeds Tab - Unknown User defined feeds at the bottom of the page.

### DIFF
--- a/net/pfSense-pkg-pfBlockerNG-devel/Makefile
+++ b/net/pfSense-pkg-pfBlockerNG-devel/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-pfBlockerNG-devel
 PORTVERSION=	3.1.0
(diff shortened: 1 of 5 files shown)


hypnosis4u2nv:
Widget clears IP and DNSBL counters everyday instead of weekly.

PR #1211, opened 2023-01-02
Feature #13821 (Redmine); Added pfSense-pkg-DNSleaktest source to sysutils;

z3d6380:
Added pfSense-pkg-DNSleaktest source to sysutils; portlint passing;

'make install' is also working and allowing me to see the package added to the pfSense GUI, under the "Diagnostics" category. The package itself is working once accessed through GUI, and DNS leak test on outgoing interfaces works.

Redmine issue submitted: https://redmine.pfsense.org/issues/13821

### DIFF
--- a/sysutils/pfSense-pkg-DNSleaktest/LICENSE
+++ b/sysutils/pfSense-pkg-DNSleaktest/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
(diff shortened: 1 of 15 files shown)


jim-p:
This is not a useful test compared to testing from a client behind the firewall where it matters more.

Furthermore, the way it's implemented is insecure and introduces remote command injection vectors because it isn't validating or sanitizing/escaping its inputs.

Even if those issues were corrected, however, this isn't something we're open to considering for inclusion at this time.

PR #1212, opened 2023-01-04
pfSense-pkg-suricata-6.0.8_5 - Improve uninstall code based on Redmine 13827 and add one new feature

bmeeks8:
### pfSense-pkg-6.0.8_5
This update improves the code flow during package uninstallation based on items discussed in [Redmine Issue #13827](https://redmine.pfsense.org/issues/13827). One new feature is also added allowing use of IPS Policy Mode to automatically set rule actions when using either Inline IPS Mode or Block On DROPs Only Mode.

**New Features:**
1. Added the ability to utilize the built-in IPS Policy "action" metadata to automatically set IPS Policy rule actions to the value specified in the policy metadata (alert or drop) when using Legacy Blocking Mode with the "Block on DROPs Only" option enabled. Formerly IPS Policy "action" metadata was only available for use when using Inline IPS Mode, but that option can also be effective in Legacy Mode Blocking when "Block on DROPs Only" is enabled.

**Bug Fixes:**
1. Improved the code flow in the package uninstall function based on information gleaned from [Redmine Issue #13827](https://redmine.pfsense.org/issues/13827).

### DIFF
--- a/security/pfSense-pkg-suricata/Makefile
+++ b/security/pfSense-pkg-suricata/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-suricata
 PORTVERSION=	6.0.8
-PORTREVISION=	4
+PORTREVISION=	5
 CATEGORIES=	security
 MASTER_SITES=	# empty
 DISTFILES=	# empty
--- a/security/pfSense-pkg-suricata/files/usr/local/pkg/suricata/suricata_uninstall.php
+++ b/security/pfSense-pkg-suricata/files/usr/local/pkg/suricata/suricata_uninstall.php
@@ -32,8 +32,6 @@ $suricatalogdir = SURICATALOGDIR;
 $sidmodspath = SURICATA_SID_MODS_PATH;
 $iprep_path = SURICATA_IPREP_PATH;
 $rcdir = RCFILEPREFIX;
-$suricata_rules_upd_log = SURICATA_RULES_UPD_LOGFILE;
-$suri_pf_table = SURICATA_PF_TABLE;
 
 syslog(LOG_NOTICE, gettext("[Suricata] Suricata package uninstall in progress..."));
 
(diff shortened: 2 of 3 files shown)


marcos-ng:
Would you elaborate on the new feature? It sounds like rules within categories have an alert or drop action, and before they were all treated as alerts when using Legacy blocking mode. Is this correct? If using the SID Management functionality to disable/drop categories/rules as needed, would this feature affect that?

jim-p (approved):

bmeeks8:
> Would you elaborate on the new feature? It sounds like rules within categories have an alert or drop action, and before they were all treated as alerts when using Legacy blocking mode. Is this correct? If using the SID Management functionality to disable/drop categories/rules as needed, would this feature affect that?

Suricata has two modes for blocking traffic. Legacy Mode Blocking is the original method. It uses a custom output plugin compiled into the Suricata binary (only for pfSense). That plugin receives a copy of every alert generated by a rule. The default behavior of that plugin, when blocking is enabled, is to insert the IP address information from the packet that triggered the alert into the _snort2c_ table in pf. There is a hidden built-in pfSense firewall rule that blocks traffic from all IP addresses contained in the _snort2c_ table. The only exception there is if the IP address in question is contained in a configured Pass List. In that case, the IP is not added to the _snort2c_ table for blocking.

The other, newer blocking mode for Suricata is Inline IPS Mode. That operating mode does NOT require the custom blocking module. It uses already built-in features in the Suricata binary to drop individual network packets as they traverse a netmap-enabled interface. This is how blocking is handled with Inline IPS Mode. There are two benefits of this mode. First, there can be no leakage of traffic as every packet must traverse the netmap pipe and can thus be inspected and dropped if necessary. The second benefit of Inline IPS Mode is that now you can use the different rule actions (ALERT, DROP, REJECT, or PASS) to take different actions for packets depending on what rule triggered. So some rules may simply alert on traffic but not block it, while other rules may drop (same as block) traffic. Legacy Mode operation does not normally allow this level of granularity where some rules block while others alert.

Suricata is different from Snort in how the internal inspection engine works. That difference enables an optional operating mode for the custom output plugin when utilizing Legacy Blocking Mode. This third option sort of mimics the individual rule actions available with Inline IPS Mode. This option is called "Block on DROPs Only", and it can be enabled as an option when Legacy Mode Blocking is used. With this option enabled, the custom output plugin will only insert an IP address into the snort2c table for blocking when the associated rule action is DROP. If the triggering rule action is ALERT, then only an alert is logged and the IP address is NOT added to the _snort2c_ table for blocking.

So now let me describe the IPS Policy Mode option. Snort VRT adds IPS policy metadata to all of their rules. This metadata does not exist for any other rules (not in Emerging Threats nor any other vendor rules). This policy metadata includes two pieces of information for rules that have it. The first piece assigns the rule to an IPS Policy. Those are "Connectivity", "Balanced", "Security", or "Max Detect". The fourth "Max Detect" category is for testing purposes only as it will generate tons of false positives in most cases. A given rule may belong to more than one IPS Policy. The second piece of information encoded with the policy metadata is a suggested rule action for each assigned policy. These can also differ. For example, the same rule be assigned to both the "Connectivity" and "Security" policies. But for the "Connectivity" policy the suggested action might be ALERT while for the "Security" policy the same rule might have a suggested action of "DROP".

So this IPS Policy Mode setting uses the embedded IPS Policy metadata to automatically configure the action for impacted rules (either ALERT or DROP). This is done outside of SID MGMT itself. This IPS Policy Mode setting has always existed for Inline IPS Mode operation. All the change in this PR does is allow the IPS Policy Mode metadata to be used with the "Block on DROPs Only" mode of Legacy Blocking Mode.

PR #1248, opened 2023-03-25
security/pfSense-pkg-acme: update formatting and descriptions

heitbaum:
Formatting updates for ACME
- fix indentation in acme.inc
- Use correct ACME capitalisation
- update spacing between functions / code blocks
- standarise on “, “ between variables
- Fix v2 text
- Fix help text

### DIFF
--- a/security/pfSense-pkg-acme/files/usr/local/pkg/acme.xml
+++ b/security/pfSense-pkg-acme/files/usr/local/pkg/acme.xml
@@ -28,10 +28,10 @@
 	]]>
 	</copyright>
 	<name>acme</name>
-	<title>Acme</title>
+	<title>ACME</title>
 	<include_file>/usr/local/pkg/acme/acme.inc</include_file>
 	<menu>
-		<name>Acme Certificates</name>
+		<name>ACME Certificates</name>
 		<tooltiptext></tooltiptext>
 		<section>Services</section>
 		<url>/acme/acme_certificates.php</url>
--- a/security/pfSense-pkg-acme/files/usr/local/www/acme/acme_accountkeys_edit.php
+++ b/security/pfSense-pkg-acme/files/usr/local/www/acme/acme_accountkeys_edit.php
@@ -66,8 +66,8 @@ if (!is_numeric($id))
 
 global $simplefields;
 $simplefields = array(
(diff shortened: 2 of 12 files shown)


security/pfSense-pkg-acme/files/usr/local/pkg/acme.xml
```diff
@@ -28,10 +28,10 @@
 	]]>
 	</copyright>
 	<name>acme</name>
-	<title>Acme</title>
+	<title>ACME</title>
 	<include_file>/usr/local/pkg/acme/acme.inc</include_file>
 	<menu>
-		<name>Acme Certificates</name>
+		<name>ACME Certificates</name>
```
jim-p, line 34: Changing the menu name is fairly disruptive. Even just changing the case, it won't match in the code that adds/removes menu entries and some users will end up with two menu entries depending on what they do.

security/pfSense-pkg-acme/files/usr/local/www/acme/acme_accountkeys_edit.php
```diff
@@ -204,15 +204,14 @@ function customdrawcell_actions($object, $item, $itemvalue, $editable, $itemname
...
-	'Let\'s Encrypt ACMEv1 servers no longer allow new registrations, and in June 2021 they will be completely disabled.%1$s%1$s', '<br/>');
+	'Use testing servers until certificate validation works, then switch to production.%1$s%1$s', '<br/>');
 
 $section->addInput(new \Form_Input(
 	'email',
 	'E-Mail Address',
 	'text',
 	$pconfig['email']
-))->setHelp('The e-mail address to register for this key. This is used by Let\'s Encrypt to send automated certificate expiration notices.');
+))->setHelp('The e-mail address to register for this key. This is used to send automated certificate expiration notices.');
```
jim-p, line 214: This should at least retain a reference to it being the CA that sends these _not_ the firewall. It's easy for users to be confused otherwise.

e.g. "The ACME Server/CA may send messages to this address, such as automated notices for expiring certificates."

jim-p (changes requested):

jim-p:
Closing for lack of response, and far too many conflicts with current code.

PR #1250, opened 2023-04-18
Update the Finnish list source url

peace2000:
I made some changes in the Finnish list repo. From now on, please use this url as the source file for the list.


BBcan177:
Thanks for the patch. Seems ok to me.

PR #1255, opened 2023-04-26
pfSense-pkg-bind  Add option to keep DDNS Changes of master zones dur…

styletronix:
[Redmine issue #14314](https://redmine.pfsense.org/issues/14314)

Added option to BIND DNS Zone GUI:
  Merge DDNS entries during zone update.
  Using named-checkzone prior to replacing zone.db on config change.
  Increment serial if zone changed. Based on actual zone serial.

Changes to bind.inc:
Add logic to detect zone config changes and update zone.db only if required.

Added ability to merge all zone records from existing zone.db bevor overwriting with new zone config.
Includes detection of deleted and added records of custom zone data and manual entries.
Current zone data is parsed with named-compilezone. Fallback to overwrite zone.db if current zone data can not be validated.

Added validation of Zone.db bevor overwriting the existing zone with named-checkzone.
If validation fails, a notice is generated and bind uses the old zone.db until the new zone is valid.

Changed service restart behaviour to use "reconfigure" instead. Includes fallback to restart service if reconfigure fails.


styletronix:
Sorry for the chaos. I am struggling a bit with github

PR #1256, opened 2023-05-02
dns/pfSense-pkg-bind: fixes and improvements. Implement #14314, Fix #11563, Fix #10693

styletronix:
[Redmine issue #14314](https://redmine.pfsense.org/issues/14314)
[Redmine issue #11563](https://redmine.pfsense.org/issues/11563)
[Redmine issue #10693](https://redmine.pfsense.org/issues/10693)

new GUI options
- Merge DDNS entries during zone update.
- Using named-checkzone prior to replacing zone.db on config change.
- Increment serial if zone changed. Based on actual zone serial.

Changes to bind.inc:
Added logic to detect zone config changes and update zone.db only if required.

Added ability to merge all zone records from existing zone.db bevor overwriting with new zone config.
Includes detection of deleted and added records of custom zone data and manual entries.
Current zone data is parsed with named-compilezone. Fallback to overwrite zone.db if current zone data can not be validated.

Added validation of Zone.db bevor overwriting the existing zone with named-checkzone.
If validation fails, a notice is generated and bind uses the old zone.db until the new zone is valid.

Changed service restart behaviour to use "reconfigure" instead. Includes fallback to restart service if reconfigure fails.

Escaping TXT and SPF zone records


marcos-ng:
Please rebase and squash commits to one per feature or redmine issue.

PR #1257, opened 2023-05-03
pfBlockerNG-devel: fixed a typo. Fixes #14189

juliokele:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/14189
- [x] Ready for review


BBcan177 (changes requested):
Thanks for finding the typo. 

I don't think we need to modify the sync php page code. If we did, it would be " != 'disabled' " tho.

BBcan177 (approved):

juliokele:
I have made a mess here, please do not approve, pfblockerng_sync.php will be deleted

PR #1261, opened 2023-05-09
LCDproc - Add NUT UPS Information. Feature: #14321

georou:
Hi,

As requested in redmine. I'd like to please get this reviewed. It's branched off my PR #1125 - it looks busy until that is merged. Jim indicated that it looks good to merge. Hopefully this is an ok path to take - branching from code that isn't merged yet(please let me know if there's a more suited path).

Functions mainly for review are L100 and L1258 if that helps.

So far under testing(months) it works well locally. I need to triple check again the remote NUT monitoring part and different size LCD screens - for intended behaviours.

I've got some questions if I could please get advice on:

- L30, is that a sane way to determine if the NUT package is installed?
- L1280, are inline if conditions ok to use?

Thank you.

Example how it should look:
LCDproc_NUT_UPS

### DIFF
--- a/sysutils/pfSense-pkg-LCDproc/files/usr/local/pkg/lcdproc_client.php
+++ b/sysutils/pfSense-pkg-LCDproc/files/usr/local/pkg/lcdproc_client.php
@@ -1277,25 +1277,25 @@ function lcdproc_loop_status($lcd) {
 									break;
 							}
 						} else {
-							$runtime = ($ups_status['_hms'] == null) ? "VOLTS: {$ups_status['battery.voltage']}/{$ups_status['battery.voltage.low']}" : "{$ups_status['_hms']}";
+							$runtime = ($ups_status['_hms'] == null) ? "VOLTS: {$ups_status['battery.voltage']}/{$ups_status['battery.voltage.low']}" : "HMS: {$ups_status['_hms']}";
 							switch($lcdpanel_height) {
 								case 1:
-									$lcd_cmds[] = "widget_set $name summary_wdgt 1 1 $lcdpanel_width 2 h 2 \"{$ups_status['_summary']} HMS:{$runtime} BATT:{$ups_status['battery.charge']}% LOAD:{$ups_status['ups.load']}%\"";
+									$lcd_cmds[] = "widget_set $name summary_wdgt 1 1 $lcdpanel_width 2 h 2 \"{$ups_status['_summary']} {$runtime} BATT:{$ups_status['battery.charge']}% LOAD:{$ups_status['ups.load']}%\"";
 									break;
 								case 2:
 									$lcd_cmds[] = "widget_set $name summary_wdgt 1 1 $lcdpanel_width 2 h 3 \"{$ups_status['_summary']}\"";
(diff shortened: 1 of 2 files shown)


dennypage:
Rather than checking if NUT is installed, it might be better to check if NUT is enabled. The relevant code from nut.inc:

```
        $nut_config = config_get_path('installedpackages/nut/config/0', []);
        if (!empty($nut_config) && isset($nut_config['type'])) {
                $type = $nut_config['type'];
        } else {
                $type = 'disabled';
        }
```

As long as type is not 'disabled' you should be good to go.

georou:
Hi Denny, Hmm, the issue would then be if a user uninstalled NUT while it was enabled. It still shows as enabled in config.xml - then it would hit my now undefined function.

Thinking about something similar to your suggestion. If I use the file_exists() at the start within lcdproc_get_nut_ups_status(), I could then remove the function_exists() and only use file_exists as the condition if NUT is installed. But I'm not sure if it's ok - style wise/syntactically to have require in a function and not at the start:

```
function lcdproc_get_nut_ups_status() {
if (file_exists("/usr/local/pkg/nut/nut.inc")) {
        require_once("/usr/local/pkg/nut/nut.inc");
        $status = nut_ups_status();
        return $status;
} else {
        return false;
}
```

elvisimprsntr:
since someone is actively adding screens for NUT and APCUPS packages, can they add NTP screens?

https://github.com/elvisimprsntr/pfsense-lcdproc

georou:
Yeah - seems there's a sudden spurt of PRs now..

Jim's version that is live now isn't fully fleshed out for the NUT part - but I dare say it will be soon. Jim didn't using anything from my PR, so I'm guessing they're going a different direction?

I'm waiting for the dust to settle before I commit time to doing a PR for NTP. What might be worth doing if you don't want to wait a long time, is open a redmine ticket for a feature request. Then either Jim or I will get to it.

elvisimprsntr:
Thanks.

I was going to pick up the last CF display off evilBay for $25 and figure out how to send it to @jim-p, but someone already snatch it up.   From the commit comments, it seems like he has one of the CF displays with LEDs and buttons.  Maybe it was him. 🤣

georou:
Sounds like it was. Pretty sure I also saw a comment somewhere he bought one and was waiting for postage.

jim-p:
I did take a different approach since I wanted to make things consistent (or close) for both APCUPSD and NUT. I prefer using APCUPSD but both work on the UPS models I have. Since this PR isn't relevant anymore, I'll go ahead and close it out.

The method I used has all the same info this has just in a condensed format, I posted a snapshot of it over on the forum: https://forum.netgate.com/topic/181790/lcdproc-package-version-0-11-4

At the moment things are in good shape in the package, at least all the changes I've wanted to make are done. I have two displays now (Adafruit LCD Backpack and a Crystalfontz 835), and I've tested every screen/menu/etc. I don't foresee any more large changes in the near future so it's a good time to add new things against the most recent code.

We could add NTP, but it would depend on the design. I don't think showing all the info for every peer is viable/desirable, even a stock default config will have 5+ peers and cycling through that isn't really worth showing IMO, but taking a very summarized view like the dashboard widget may be OK (current time, and basic info on the active peer only).

If you do make a PR for an NTP screen, be sure to open a Redmine issue and link it both ways (Redmine links to the PR, PR links to the Redmine).

PR #1272, opened 2023-07-06
Fix Secure Cookie config when UI flag selected

alfredopironti:
Replace unsupported "rspirep" directive with modern "http-response replace-header" one.

Fixes bug https://redmine.pfsense.org/issues/13343


jim-p:

alfredopironti:
Thanks for all the guidance, @jim-p! Clearly my first contribution in this space. I've now increased PORTREVISION (and caught a missing character escape). Please let me know if I can contribute any further.

PR #1275, opened 2023-07-25
New package: FTP Reverse Proxy with proftpd_mod_proxy (ticket #16986)

cmassiot:
I am new to pfSense development and I am pretty sure this doesn't meet the standards, but I post it in the hopes that somebody else finds it useful.


cmassiot:
Reverse HTTP is already possible with the haproxy package, but there is no package to balance FTP connections. The proposed package uses proftpd_mod_proxy to balance FTP connections based on the FTP login.
Implements #16986

PR #1277, opened 2023-07-31
Add LCDPROC NTP Screen

elvisimprsntr:
Requested updates from @jim-p 

Original work by @georou

https://redmine.pfsense.org/issues/14625


jim-p (changes requested):
Testing this out a couple things seem less than ideal:

1. The screen has no label so it's not obvious what it is. At least printing "NTP: " before the IP address would be helpful
2. The "(stratum n)" wastes a ton of space on something I'm not sure is significant and it obscures the IP address more often than not. "Stratum" could be abbreviated to "St" maybe or omitted entirely. It would be even worse for an IPv6 address
3. There is quite a bit of dead space after the Delay/Offset/Jitter labels, though I'm not sure anything can be done for that. The numbers might be able to fit on a single line going horizontally on two lines instead of three. Like we do for the system stats or UPS info. That would free up a whole line to print the Title/IP address and stratum info separately.

elvisimprsntr:
1,2. Problem is there is not enough characters to display everything. Displaying the sync source on a single line is consistent with the web interface.  Scrolling is enabled on the row to see the entire value.

Screenshot 2023-07-31 at 10 18 00 PM

3. Problem is the delay, offset, and jitter are in msec.  So if someone is not using a GPS or PPS, the values can have a large number of significant digits.  Placing each on a single row allows for a variable length value.  

If you have a better idea, feel free to make changes as you see fit.

PR #1280, opened 2023-08-05
Update to LCDPROC NTP Screen

elvisimprsntr:
Update to LCDPROC NTP Screen
* Add time zone
* Add local PPS stability pps

https://redmine.pfsense.org/issues/14653


elvisimprsntr:
More tweaks coming to get PPS vs GPS, VS PPS+GPS working

PR #1281, opened 2023-08-05
Update LCDPROC NTP Screen

elvisimprsntr:
@Jim-p @georou

Update LCDPROC NTP Screen
* Add time zone
* Improved selection between GPS and PPS
* Add stability parameter for PPS source

https://redmine.pfsense.org/issues/14653


jim-p:
I merged it manually so it would be in one single commit (and cleaned up some whitespace issues).

Thanks!

elvisimprsntr:
Thanks.  I'll eventually figure out how to make it a single commit.  Just takes so long to fetch the entire FreeBSD-ports to my Mac.  Or figure out how to fork a portion of the FreeBSD-ports

PR #1282, opened 2023-08-06
Update haproxy_global.php to fix 14445

yobyot:
In pfSense CE 2.7, PHP has been updated to 8.2.6.  This version of PHP is stricter in using variables.

This fixes the issue described in 14445 (https://redmine.pfsense.org/issues/14445) by explicitly defining an array before assigning a value to an element of that array.


net/pfSense-pkg-haproxy/files/usr/local/www/haproxy/haproxy_global.php
```diff
@@ -135,6 +135,7 @@
 					unset($haproxycfg['config'][0]['enable']);
 				}
 			} else {
+				$haproxycfg['config'][0] = array();
```
jim-p, line 138: That would help if just the one level is missing but a better fix is to not add this line, but to use the new access functions to read the configuration variable below. I'll post a follow-up comment on that line.

net/pfSense-pkg-haproxy/files/usr/local/www/haproxy/haproxy_global.php
```diff
@@ -135,6 +135,7 @@
...
 				$haproxycfg['config'][0]['enable'] = 'off';
```
jim-p, line 139: Using our new access functions, this line would become:

```php
array_set_path($haproxycfg, 'config/0/enable', 'off');
```

That way is safe no matter what parts may be missing from the configuration data.

jim-p (changes requested):
There is a better way to reach the goal here, of which you may not be aware. We have recently changed to using new array and config access functions that let us work around these PHP problems in an abstract and safer way.

We have a new development document for them here: https://docs.netgate.com/pfsense/en/latest/development/php-config-arrays.html

If you would like to update or resubmit the PR with that change, feel free, but if you are not able to or do not have the time, I can make the change manually.

Additionally, the Makefile needs to have the PORTREVISION or PORTVERSION increased to trigger a new build/update.

Thanks!

yobyot:
_Thanks!_, @jim-p for the quick review.

As a total php (and pfSense) newbie, I'd like to attempt making the requested changes. But I need some time to work through your comments. This is a learning process for me.

If I can't grok it all, I'll let you know here. Otherwise, please forgive in advance any additional effort on your part to correct my feeble attempts and, maybe, ultimately, having to do it yourself. :-)

jim-p:
Go for it! If you get stuck, let us know.

Given that it's a small change it's no trouble at all if we have to make a fix manually, so no worries either way.

jim-p (changes requested):
We do intend to fully convert everything in base and packages over to the new config access functions over time, but it's an ongoing process and there is only so much developer time to go around.

If you look at recent commits by us, there is a steady stream of packages being converted (including FRR in the last week or so), it just takes time and the larger/more complex packages take even more time.

Thanks!

yobyot:
@jim-p: I actually thought about using `array_del_path` instead of `array_set_path` but went with the latter because I thought it better illustrated the logic. IOW, cosmetics. Gets you every time. :-) 

Also, I was unsure of exactly what the logic was doing. It seemed to me, just reading the code cold, that the `unset` in the current `if` was nullifying the true case -- IOW, backwards logic.

Anyway, I hope this commit better conforms to the project's standards overall.

Thanks.

yobyot:
@jim-p, _thank you!_

Appreciate the chance to work with the pfSense maintainers.

PR #1292, opened 2023-09-10
security/pfSense-pkg-acme: improve Cloudflare documentation

Vynce:
Many guides on setting up ACME certs with Cloudflare in pfSense show filling out all five authentication fields. This is not required for acme.sh to work correctly and potentially exposes Cloudflare credentials with broad access though the pfSense UI and configuration backups.

There are several ways that acme.sh can authenticate to Cloudflare, from least to most permissive:

1. Token with `Zone.DNS:Edit` permission and `Zone ID`. This only works with certs that cover a single zone.
2. Token with `Zone.Zone:Read` and `Zone.DNS:Edit` permission and `Account ID`. This works with certs that cover multiple zones. acme.sh uses the Account ID to look up all Zone IDs.
3. Global API key and email address. This is not recommended since it provides complete access to the entire Cloudflare account.

References:
https://github.com/acmesh-official/acme.sh/wiki/dnsapi#dns_cf
https://github.com/acmesh-official/acme.sh/issues/2398
https://github.com/acmesh-official/acme.sh/commit/c25947d5447ac03ac0fae56b338e8821d49d61ac

### DIFF
--- a/security/pfSense-pkg-acme/Makefile
+++ b/security/pfSense-pkg-acme/Makefile
@@ -2,6 +2,7 @@
 
 PORTNAME=	pfSense-pkg-acme
 PORTVERSION=	0.7.5
+PORTREVISION=	1
 CATEGORIES=	security
 MASTER_SITES=	# empty
 DISTFILES=	# empty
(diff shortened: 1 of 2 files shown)


andrebrait:
I'll take a look at this later and see if I can align wording with what you used here for this (un)related PR: https://github.com/pfsense/pfsense/pull/4688

jim-p:
The current package version improves on this a bit but doesn't take the same approach as this PR, to reduce the amount of text in that section.

If that is not sufficient, please submit a new PR against the current code base.

PR #1293, opened 2023-09-16
net/frr7: support multiple OSPF instances in raw config

edvalley:
I would like to propose this changes (some refactoring included) to support multiple ospf instances in pfSense when using raw configuration mode. I can confirm I've successfully tested it (pfSense CE 2.7.0) with three ospf instances (not ospf6) without having any issue, even if running only one instance while using the pfSense web interface. Thanks...

### DIFF
--- a/net/frr7/files/frr.in
+++ b/net/frr7/files/frr.in
@@ -30,136 +30,209 @@
 #
 # This RC script was adapted from the net/quagga port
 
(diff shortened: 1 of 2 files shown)


edvalley:
With the last two commits, we are one step closer to one day supporting multiple instances in GUI

PR #1294, opened 2023-09-19
Package: pfSense-pkg-sfpnfo

solariz:
### pfSense-pkg-sfpnfo

Package sfpnfo to display easy accsessible information about SFP Modules present in the System in Menu > Status > SFP Info


solariz:
Redmine:
https://redmine.pfsense.org/issues/14793

jim-p:
This is not needed. SFP information is already printed on Status > Interfaces. If more detail is needed the additional fields could be added there rather than making a whole separate package for it.

solariz:
Seeing it different. The present view which exists in interfaces is cluttered and bad.

solariz:
@jim-p Hey, I still see it as a valuable addon. Sometimes the interface info does not display SFP Module information at all. e.g. when running a LAG or Bridge. We run about 50 Firewalls als CPE where I would like to have it as a extra option to display those information for easy troubleshooting. This is why I added it as a package and not to the core, so everyone can decide if it is useful or not. Please have a look into it and let me know if anything need to change.

PR #1295, opened 2023-09-19
Move from nut-2.8.0 to nut-devel

dennypage:
Note that this PR depends upon the FreeBSD "nut-devel" package rather than the "nut" package.

Please see Redmine issue [14795](https://redmine.pfsense.org/issues/14795) for details behind this change.

### DIFF
--- a/sysutils/pfSense-pkg-nut/Makefile
+++ b/sysutils/pfSense-pkg-nut/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-nut
 PORTVERSION=	2.8.0
-PORTREVISION=	2
+PORTREVISION=	5
 CATEGORIES=	sysutils
 MASTER_SITES=	# empty
 DISTFILES=	# empty
@@ -13,7 +13,7 @@ COMMENT=	Network UPS Tools
 
 LICENSE=	APACHE20
 
-RUN_DEPENDS=	nut>=0:sysutils/nut
+RUN_DEPENDS=	nut>=0:sysutils/nut-devel
 
 NO_BUILD=	yes
 NO_MTREE=	yes
--- a/sysutils/pfSense-pkg-nut/Makefile
+++ b/sysutils/pfSense-pkg-nut/Makefile
@@ -1,8 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-nut
-PORTVERSION=	2.8.0
-PORTREVISION=	5
+PORTVERSION=	2.8.1
 CATEGORIES=	sysutils
 MASTER_SITES=	# empty
 DISTFILES=	# empty
--- a/sysutils/pfSense-pkg-nut/files/usr/local/pkg/nut.xml
+++ b/sysutils/pfSense-pkg-nut/files/usr/local/pkg/nut.xml
@@ -28,7 +28,7 @@
 	]]>
 	</copyright>
 	<name>nut</name>
-	<version>2.8.0_4</version>
+	<version>2.8.0_5</version>
 	<title>Services: UPS</title>
 	<savetext>Change</savetext>
 	<include_file>/usr/local/pkg/nut/nut.inc</include_file>
--- a/sysutils/pfSense-pkg-nut/files/usr/local/pkg/nut.xml
+++ b/sysutils/pfSense-pkg-nut/files/usr/local/pkg/nut.xml
@@ -28,7 +28,7 @@
 	]]>
 	</copyright>
 	<name>nut</name>
-	<version>2.8.0_5</version>
+	<version>%%PKGVERSION%%</version>
 	<title>Services: UPS</title>
 	<savetext>Change</savetext>
 	<include_file>/usr/local/pkg/nut/nut.inc</include_file>
--- a/sysutils/pfSense-pkg-nut/files/usr/local/pkg/nut/nut.inc
+++ b/sysutils/pfSense-pkg-nut/files/usr/local/pkg/nut/nut.inc
@@ -54,8 +54,8 @@ function nut_check_var_db() {
 	if (!is_dir($d)) {
 		mkdir($d, 0750, true);
 		safe_mkdir($d, 0750);
-		chown($d, "uucp");
-		chgrp($d, "uucp");
+		chown($d, "nut");
+		chgrp($d, "nut");
 	}
 }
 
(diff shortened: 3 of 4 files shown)


sysutils/pfSense-pkg-nut/Makefile
```diff
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-nut
 PORTVERSION=	2.8.0
-PORTREVISION=	2
+PORTREVISION=	5
```
jim-p, line 5: This seems like way too large a change for just a portrevision bump. It warrants a larger increase to give the user a better idea about the impact of the change. Though it doesn't look like they've settled on a new version upstream yet, seems like `2.9.0` is most likely. There really isn't a compelling reason to tie this version to the NUT version. The NUT version is shown in the package info separately these days.

jim-p (changes requested):

dennypage:
NUT doesn't really follow semantic versioning. Numbers yes, meaning no. The next release is planned to be 2.8.1, followed by 2.8.2 and 2.8.3. You can see planned release info [here](https://github.com/networkupstools/nut/blob/master/NEWS.adoc). As you can see from the 2.8.1 milestone, it's almost all bug fixes and doc fixups. In my opinion (and other downstreams), 2.8.1 should have been released at least a year ago due to the sheer magnitude of serious defects, but I don't run the project. The NUT team have continued working on 2.8.1, but have kept adding more things along the way. The milestone has sat around 90% for a long time, and there is no target date. It's what we used to call a "run-on" release. They suggest that people build from the master branch to get bug fixes.

Coming back to our problem, have a look at the planned release notes for 2.8.1. A port revision change (2.8.0_5) still seems okay to me because for all intents and purpose, we're principally fixing bugs. There are very few actual features in 2.8.1. The big updates for documentation and Windows don't apply to us.

That said, I understand your concern. Perhaps we should call it 2.8.1. That would mean that when the NUT team eventually releases 2.8.1, ours would be a port revision change (2.8.1_x) to switch the dependency from nut-devel to nut-2.8.1. At that point, the change in the underlying NUT would be very negligible as NUT releases are generally done by a tag of master rather than developing the release in a branch.

The last option I see would be to completely walk away from the NUT version for the pfSense package. But if we are to walk away from aligning with NUT versions, we would need to walk far enough away so as to prevent confusion with users over the meaning of the version of the pfSense package and meaning of the underlying NUT version. Something like 10.0.0. I'm not really fond of this as it tends to create a lot of confusion with with the users who generally expect the numbers to align. Lot's of downstream cost.

The more I think about it, the more I like 2.8.1. This seems to be the most honest representation of versions. Thoughts?

sysutils/pfSense-pkg-nut/Makefile
dennypage: I responded in the wrong box. My bad. Please see below.

jim-p: 2.8.1 would be fine.

The main problem with trying to keep them aligned is that it forces you to abide by their versions even if you make substantial changes to the package. Ideally the versions should be separate, but I get wanting to tie them together. In the past before we showed the dependencies in the package list it wasn't obvious what was being included, but that hasn't been the case in several years now. Users are shown quite visibly which underlying version would be installed.

dennypage: Thanks Jim. I've changed the version to 2.8.1.

sysutils/pfSense-pkg-nut/files/usr/local/pkg/nut.xml
```diff
@@ -28,7 +28,7 @@
 	]]>
 	</copyright>
 	<name>nut</name>
-	<version>2.8.0_4</version>
+	<version>2.8.0_5</version>
```
jim-p, line 31: This version didn't follow the other change. Though honestly the tag doesn't need to be in this file any longer, the version is handled entirely in the Makefile variables now.

There is still a version tag in `files/usr/local/share/<pkg name>/info.xml` but it should be using a macro like `<version>%%PKGVERSION%%</version>` so you don't ever have to update it by hand.

jim-p (changes requested):

sysutils/pfSense-pkg-nut/files/usr/local/pkg/nut.xml
dennypage: My bad. Thanks for the catch. I'll change it to %%PKGVERSION%%.

dennypage: Done

jim-p:
I had to revert this merge as the `nut-devel` port failed to build. The main issue seems to be that in the `nut-devel` `Makefile` it's still called `nut` and not `nut-devel` but `portlint -C` also found quite a few other problems in the `Makefile` that the maintainer may need to correct as well.

The exact error from the build system was:
> `Warning: sysutils/pfSense-pkg-nut dependency on sysutils/nut-devel has wrong PKGNAME of 'nut' but should be 'nut-devel'; Is the dependency missing a @FLAVOR?`

dennypage:
I was able find one other pfSense package that does use a -devel version which is haproxy. The Makefile for the dependent, haproxy-devel, has the PKGNAMESUFFIX approach as nut-devel, so I assume that part should work as expected.

I definitely made an error in not changing both parts of the RUN_DEPENDS variable to nut-devel. I will make that change.

dennypage:
I made the change, but when I push it git wants to create a new PR with just that one change in it. @jim-p, would you prefer me to redo the entire PR or just a follow-on PR? Thanks.

jim-p:
At this point it would be nice to have a new PR with the changes in a single commit, that way if we end up picking it to other branches it's much easier.

dennypage:
Replacement PR created.

dennypage:
Jim, was there an issue with the replacement PR?

jim-p:
Yes, sorry, I meant to comment on here Friday afternoon. The `nut-devel` port itself failed to compile that time so I had to revert the change again. Unfortunately the build system didn't print the whole error message so it may take some more investigation. The copy of `nut-devel` on our `devel` branch is behind the one on `main`, so it's possible whatever is wrong is already fixed upstream.

dennypage:
Okay, thanks. Please let me know what you find.

dennypage:
I did a quick check of diffs for sysutils/nut-devel between the freebsd/freebsd-ports repo and the pfsense/FreeBSD-ports repo. The only differences I saw were related to the GitHub tag and commit date, so I assume that whatever is preventing the build is outside of that subdirectory.

FWIW, the current build is available in the FreeBSD 14 repo: http://pkg.freebsd.org/FreeBSD:14:amd64/latest/All/nut-devel-2023.09.02.pkg.

dennypage:
I have found the issue and understand why pfSense-pkg-nut is not building with nut-devel. A change is required for the nut-devel packing list, which I will submit upstream shortly.

PR #1296, opened 2023-09-21
Move from nut-2.8.0 to nut-devel

dennypage:
Replacement for PR #1295

### DIFF
--- a/sysutils/pfSense-pkg-nut/Makefile
+++ b/sysutils/pfSense-pkg-nut/Makefile
@@ -1,8 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-nut
-PORTVERSION=	2.8.0
-PORTREVISION=	2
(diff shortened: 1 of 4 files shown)


marcos-ng:
Unfortunately the build failed again (unclear why at the moment), hence the changes have been reverted.

dennypage:
Are there differences between the NG build environment and the standard FreeBSD environments?

Is it possible for someone outside of NG to replicate the NG build environment?

dennypage:
I have found the issue and understand why pfSense-pkg-nut is not building with nut-devel. A change is required for the nut-devel packing list, which I will submit upstream shortly.

No changes will be required in the pfSense-pkg-nut PR. Is it possible to re-open the PR?

marcos-ng:
This one's already merged, so a new one will be needed (make sure to include `Implement #14795` in the title/commit message).

PR #1297, opened 2023-09-24
Package: pfSense-pkg-ipfs-go

jackmurray:
*Not ready for merging yet, just gauging interest*

This package adds a pfSense package for the `ipfs-go` daemon. Much like a caching squid proxy, I wanted to run this on a shared network device that everything else on the network can reach.

Outstanding work:
- Add status page and more config options
- Does anything need to be done to build/publish the `ipfs-go` port that this package depends on?
- Is there a way to get a CI build of this package to ensure it works when built in a proper environment that's not my test FreeBSD VM?


jim-p:
This isn't the kind of service that anyone should be running on a secure firewall. IPFS doesn't need to run on your edge or be publicly exposed, and nothing should be using the firewall as a file server.

PR #1298, opened 2023-09-24
security/pfSense-pkg-acme: support deploy hooks. Implement #11827

Inperpetuammemoriam:
The deploy hooks provided by the upstream project have been included and the UI now supports the configuration of actions that call these hooks using the provided environment.

A corresponding feature request can be found [here](https://redmine.pfsense.org/issues/11827).


jim-p:
I'm not entirely opposed to including them but I'm still hesitant to do so as their behavior could be unpredictable and conflict with the base system in various ways. Having to vet all the scripts to reduce foot shooting would be taking on a lot of technical debt.

If you can redo the PR **without** including the deploy scripts, we can look that over a lot easier. Any code from ACME gets pulled in a different way and we don't take that code via PRs because we need to make sure the files are unmodified from upstream.

Inperpetuammemoriam:
I understand your concerns on the technical debt that might come with the inclusion of the deploy hooks.

On the other side, I can clearly see a benefit for those users using their pfSense to renew the certificates of nearby appliances that do not provide an ACME capability. Up to now, those users probably either fiddled the hooks manually into their systems or even wrote custom deploy scripts (This is what I did.). Having a turnkey solution at hand would save those users a lot of time and would enable others that lack the capability to implement a solution on their own to profit from the ease of use as well.

I will create a new PR without the scripts on the weekend.

PR #1302, opened 2023-10-04
pfBlockerNG: Support AdBlock Whitelists with wildcards

andrebrait:
Creating this draft PR for early commenting

So far, I've tested it for Unbound mode. Likely needs some small adjustments for Python mode. I'll be testing it soon.

I need some special help with testing it and with the log message formatting. I'm not very happy with how it looks right now, but I whipped this together somewhat quickly, so beauty wasn't my priority.

---

Add support for AdBlock-style exclusions/whitelisting for pfBlockerNG.

While it already supported AdBlock-style blacklists (minus wildcards), whitelists weren't a thing. This PR adds support for them (including wildcards, since there's not much complexity in doing that) through filtering out the domain lists, exactly like what it does with user-defined whitelists, but expanding that to use Regular Expressions, allowing it to support wildcard entries.

Missing yet:
* Testing this on Python Mode

Next changes I plan on making, assuming this PR turns out ok:
* Wildcard support for Blacklists (Python mode only)


andrebrait:
I caught an issue with the Python mode yesterday. Fixing it later today.

andrebrait:
Closed this one as I'll be sending a more complete one instead

PR #1303, opened 2023-10-06
pfBlockerNG: extend support for AdBlock-style lists

andrebrait:
This supersedes #1302 and closes [#14838](https://redmine.pfsense.org/issues/14838). 

Extend support for AdBlock-style lists:

* Unbound mode:
  * Support whitelist entries, including wildcards
* Python mode:
  * Support whitelist entries, including wildcards
  * Expand blacklist entry support to also handle wildcards

In Python mode, most processing was moved to Python, with the surrounding code merely assembling lists that get consumed by it.

User-defined, TOP1M and Whitelist entries from AdBlock-style lists get all joined in a single file (PHP) that gets loaded and parsed later (Python).

Whenever possible, simple domain matches are kept as-is. The presence of any wildcards triggers the conversion of that specific entry into a Regular Expression in the Python format.

In Unbound mode, everything is kept the same, except that whitelisting is now done in 3 distinct steps (User-defined and TOP1M using fixed string matching with ggrep, and Whitelist entries from AdBlock-style lists using extended regular expressions with ggrep), in a postprocessing step separate from the deduplication, etc.

In both cases, the `# White` count in the logs now refers to how many whitelist entries were found, rather than how many domains were removed from the Blacklist.

This still needs testing and input on where to log things in a more adequate fashion.


andrebrait:
Idea: move the `.whitelist` files under the `Permit files` category in the `Logs` tab.

andrebrait:
Superdesed by #1343

PR #1308, opened 2023-10-09
Update suricata.inc - fix suricata_Getdirsize tab char

graham-collinson:
Assuming this was intended to be the tab character.  PHP 8 changed strpos so that Integer is no longer cast to a character by strpos.

### DIFF
--- a/security/pfSense-pkg-suricata/Makefile
+++ b/security/pfSense-pkg-suricata/Makefile
@@ -2,7 +2,7 @@
 
 PORTNAME=	pfSense-pkg-suricata
 PORTVERSION=	7.0.0
-PORTREVISION=	1
+PORTREVISION=	2
 CATEGORIES=	security
 MASTER_SITES=	# empty
 DISTFILES=	# empty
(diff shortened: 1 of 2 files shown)


graham-collinson:
redmine issue https://redmine.pfsense.org/issues/14855

bmeeks8:
I am good with this fix, but it needs an additional update to the version in the Makefile so that the package will rebuild and show as an available update. Please increment the PORTREVISION parameter in the Makefile to be "2" (it is currently "1" in the DEVEL snapshot branch package). That will make the current Snapshot branch Suricata GUI package version be 7.0.0_2.

graham-collinson:
@bmeeks8 Thanks for the pointer.  That ok now?

bmeeks8:
> @bmeeks8 Thanks for the pointer. That ok now?

Perfect! Now we wait on the Netgate developer team to review and merge. I can drop them an email to hopefully speed it along. They are a bit preoccupied I think with the upcoming 23.09 release.

As the package maintainer, they usually ask me to weigh in with my thoughts on community pull request submissions, so that's why I proactively jumped in.

PR #1310, opened 2023-10-11
Move from nut-2.8.0 to nut-devel. Implement #14795

dennypage:
Implements Redmine issue [14795](https://redmine.pfsense.org/issues/14795). Replacement for PRs #1295 and #1296.

**Note that this PR is dependent on fixes to the FreeBSD sysutils/nut-devel port that were committed [upstream](https://cgit.freebsd.org/ports/log/sysutils/nut-devel) at 2023-10-11 04:28:25 +0000. This PR will not build without those fixes.**

### DIFF
--- a/sysutils/pfSense-pkg-nut/Makefile
+++ b/sysutils/pfSense-pkg-nut/Makefile
@@ -1,8 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-nut
-PORTVERSION=	2.8.0
-PORTREVISION=	2
+PORTVERSION=	2.8.1
 CATEGORIES=	sysutils
 MASTER_SITES=	# empty
 DISTFILES=	# empty
@@ -13,7 +12,7 @@ COMMENT=	Network UPS Tools
 
 LICENSE=	APACHE20
 
-RUN_DEPENDS=	nut>=0:sysutils/nut
+RUN_DEPENDS=	nut-devel>=0:sysutils/nut-devel
 
 NO_BUILD=	yes
 NO_MTREE=	yes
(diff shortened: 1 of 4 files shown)


marcos-ng:
Thank you; commit is here:
https://github.com/pfsense/FreeBSD-ports/commit/e55ac518e1e2a4359dbf3b0e5e36aa235bfe1f13

dennypage:
Will this package update be pushed for 23.05? Or is it being held for 23.09?

marcos-ng:
There are other commits that would need to be picked back to 23.05.1 to allow the package to build in all platforms. For now anyway, availability will start with 23.09.

dennypage:
Marcos, do you know when the new package will be published?

marcos-ng:
Ideally there would be some user feedback on the redmine before it's merged from the devel branch into the release branch.

dennypage:
I don't think anyone can get the package until it's published.

marcos-ng:
It had been available in the devel branch. Now that 23.09 is out however, the devel branch is currently private; it should become public again soon, at which point it can be tested. If that goes well, I'll merge it into the 23.09 branch.

dennypage:
@jim-p, @marcos-ng, I really have quite a few people waiting on this update.

It's pretty unclear that the people we would want to test the package are going to be on development branches. I think we should just publish it. I will handle the support. I have time now to do this now, but I won't have time two weeks from now. Please consider this.

If you are completely set against publishing the update, can you please give me a copy of your package builds? This way I can at least make it available to the people who would want to test it, which would be a lot more beneficial than me distributing my builds.

marcos-ng:
Sure, see here: https://redmine.pfsense.org/issues/14795#note-10

PR #1324, opened 2023-11-17
Updates acme dns inwx script to current version

Ritzelprimpf:
Hey guys,

dns_inwx.sh script was a bit outdated and was missing a change regarding case sensitive grep.

Script change can be viewed here:

https://github.com/acmesh-official/acme.sh/commit/9143cd1485a4a0220897124fec78eb2ec45f81ec

Before the change, updating letsencrypt certs using INWX API failed.

Thank you for your great work.

Best,
Martin


Ritzelprimpf:
OK, cool. But your version seems 3 years old and the bug currently exists.

https://github.com/pfsense/FreeBSD-ports/blob/devel/security/pfSense-pkg-acme/files/usr/local/pkg/acme/dnsapi/dns_inwx.sh

Maybe I misunderstood something. However, I am happy if the bug gets fixed automatically.

PR #1325, opened 2023-11-17
suricata-7.0.2_4 - Use pfctl_table_add_addrs() and fix memory leaks and compiler warnings.

bmeeks8:
### suricata-7.0.2_4
This update to the Suricata custom blocking plugin used for Legacy Blocking Mode on pfSense includes some general code cleanup in the form of formatting fixes and other cosmetic issues. It also fixes two memory leaks and replaces a legacy _ioctl()_ call with the corresponding _pfctl_table_add_addrs()_ function from the new _libpfctl_ wrapper library for _pfctl_ functionality.

### DIFF
--- a/security/suricata/Makefile
+++ b/security/suricata/Makefile
@@ -1,6 +1,6 @@
 PORTNAME=	suricata
 DISTVERSION=	7.0.2
-PORTREVISION=   3
+PORTREVISION=   4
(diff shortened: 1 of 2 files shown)


bmeeks8:
This request was manually merged by kprovost, so closing it manually since GitHub will not auto close it.

PR #1329, opened 2023-11-18
Fix port version number

dennypage:
Discussion in Redmine [14795](https://redmine.pfsense.org/issues/14795).

If this will work, I'd like to fix the version. If it will not work please let me know and I will close the PR.

Thanks

### DIFF
--- a/sysutils/pfSense-pkg-nut/Makefile
+++ b/sysutils/pfSense-pkg-nut/Makefile
@@ -1,7 +1,8 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-nut
-PORTVERSION=	2.8.2
+PORTVERSION=	2.8.1
+PORTREVISION= 1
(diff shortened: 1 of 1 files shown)


marcos-ng:
That wouldn't work because updates are only detected when the version is higher. If a newer release has a version that is lower, then PORTEPOCH is needed. However, the issue then becomes that PORTEPOCH needs to be maintained. The documentation here explains it in more detail: https://docs.freebsd.org/en/books/porters-handbook/makefiles/index.html#makefile-portrevision

I'm not aware of any functional differences with other projects by keeping things as is, so perhaps it's not worth the effort (at least until a new version is released).

dennypage:
Yea, that's what I thought. Looks like we're stuck for a year or two until NUT catches up. For the future, please put up a PR for review instead of doing a straight commit. Thanks.

PR #1330, opened 2023-12-04
acme.sh nsupdate with challengealias is failing

seyfidin:
fixed wrong filename for key and server file when using nsupdate with challengealias

https://redmine.pfsense.org/issues/15061


jim-p:
Added manually in ACME pkg v0.8

PR #1331, opened 2023-12-07
net/pfSense-pkg-pfBlockerNG-devel: add option to disable reverse DNS lookups. Fix #15072

opoplawski:
I think there is still an issue with the default setting, and we need to restart the filterlog service when this setting is changed.

Fixes https://redmine.pfsense.org/issues/15072


JulioQc:
https://www.reddit.com/r/pfBlockerNG/comments/blmw1m/disable_reverse_dns_lookup/

PR #1339, opened 2023-12-21
pfSense-pkg-arpwatch: Add Slack messaging

MichaelLegatt:
This PR enhances the arpwatch package, so that Slack messages, if so configured, are sent during notifications, in addition to the existing email, telegram, and/or pushover notifications. 

Thanks!


marcos-ng:
IMO a better approach is to adapt the code to use notify_all_remote() instead.

PR #1342, opened 2024-01-15
Add specific log messages when MaxMind updates fail

DoubleStrike:
This adds specific messages to show that the files either existed or that the API key is empty.

Without these messages, the action just silently fails without logging anything to the window/log.


BBcan177:
I don't know if this would be of benefit. If the user enters the Account and Key, I don't see a benefit to log it each time cron runs?

DoubleStrike:
> I don't know if this would be of benefit. If the user enters the Account and Key, I don't see a benefit to log it each time cron runs?

I've had issues on multiple installs where it would not update the local file because it thought it already existed. By adding this log message I was able to see why the update wasn't working and trigger it manually via the PHP DC command. A simple one-liner to the log to show why the update didn't happen seems rather useful for a very small trade-off of log size increase.

PR #1356, opened 2024-03-23
security/tailscale: sync with upstream

bdwyertech:
Update Tailscale from 1.56.1 to 1.62.0

source:  https://www.freshports.org/security/tailscale/


andrebrait:
Ps.: after a few months, upstream is now at 1.68.2 (for arm64) and 1.70.0 (for amd64)

marcos-ng:
This is already in devel branches (currently 1.76.1).

PR #1361, opened 2024-03-30
Updated VPN pfblockerng feed (dead feed)

dylfen:
Changed from a dead feed to a maintained feed

The current Ejrv_VPNv doesn’t exist.
Found a replacement at https://github.com/X4BNet/lists_vpn

### DIFF
--- a/net/pfSense-pkg-pfBlockerNG/files/usr/local/www/pfblockerng/pfblockerng_feeds.json
+++ b/net/pfSense-pkg-pfBlockerNG/files/usr/local/www/pfblockerng/pfblockerng_feeds.json
@@ -748,16 +748,16 @@
 			]
 		},
 		"VPN_4": {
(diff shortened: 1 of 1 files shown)


net/pfSense-pkg-pfBlockerNG/files/usr/local/www/pfblockerng/pfblockerng_feeds.json
```diff
@@ -748,16 +748,16 @@
 			]
 		},
 		"VPN_4": {
-			"info":		"A list of common VPN providers",
-			"description":	"List of common VPN providers.",
+			"info":		"A list of IP's used by VPN providers",
```
andrebrait, line 751: nit: I guess people use it, but the "correct" plural would be "IPs".
Even better would be "IP addresses"

PR #1364, opened 2024-04-02
Update ntopng package.

dennypage:
Implements #15375

Redmine issue [15375](https://redmine.pfsense.org/issues/15375).

- Update for PHP 8.
- Support newly required Account ID for MaxMind GeoLite2 downloads.
- Run ntopng without root privileges.
- Add option to disable setting promiscuous mode on monitored interfaces.
- Allow selection of a specific interface for the ntopng server to listen on.
- Add support for the ntopng server listening on IPv6 addresses.
- Move ntopng configuration options from the command line into a configuration file. This allows for credentials, such as those required by dump-flows, to be secured.
- Support the addition of configuration lines outside those directly supported by the pfSense ntop package UI. This allows for customization beyond the capabilities of the package, such as importing or exporting flows to another ntopng instance.
- Use the demonize capability of the redis server.
- Record in redis that the admin password has already been set. Without this, ntopng forces the change of the admin password on first access.
- Provide an option to delete ntopng data (traffic and alerts) without destroying the ntopng configuration (preferences and customizations).
- Clarify the use of the local networks setting, and remove unnecessary options. Selected/monitored networks are already considered as local by ntopng. The --local-networks option is only used for adding networks that are not already known.
- Support setting names for additional local networks.


dennypage:
One additional note: This updated version of the pfSense package has been tested with the currently distributed ntopng package (ntopng-5.6.d20230531_1,1.pkg), as well as with the current 6.0 version in FreeBSD ports (topng-6.0.d20240307,1.pkg).

net/pfSense-pkg-ntopng/files/usr/local/pkg/ntopng.inc
```diff
@@ -217,112 +325,208 @@ function ntopng_redis_started() {
...
+	log_error(gettext("[ntopng] All traffic data, alerts, preferences and customization deleted."));
+
+	/* Resync settings and restart service if enabled */
+	unset($_POST['Delete']);
+	ntopng_sync_package();
+	log_error(gettext("[ntopng] Resynced ntopng settings."));
+}
+
 function ntopng_validate_input($post, &$input_errors) {
+	/* Validate password */
```
jim-p, line 529: As long as you're validating the password, add in something like this as well:

```php
$input_errors = [];
//...
if (function_exists('validate_password')) {
	$input_errors = array_merge($input_errors, validate_password('redis', $post['redis_password']));
}
```

That will reject password values considered invalid by the base system.

jim-p (changes requested):

net/pfSense-pkg-ntopng/files/usr/local/pkg/ntopng.inc
dennypage: I don't see validate_password() used in any packages, nor was I able to find it in pfSense itself. Am I missing something?

FWIW, I am very reluctant to introduce potentially new password requirements. This could result in a previously set admin password suddenly being unacceptable, which would prevent other settings from being changed. I think it's better to leave this alone. If ntopng had its own externalized validator I might feel differently, but it doesn't.

jim-p: It would only be preventing values nobody should be using in the first place.

https://redmine.pfsense.org/issues/15266

https://github.com/pfsense/pfsense/blob/master/src/etc/inc/auth.inc#L652

At the moment on CE it would just stop someone from using `redis` for the password, and on Plus it would also prevent using `pfsense`.

It makes it consistent with the base system password behavior on Plus 24.03 and later / CE 2.8.0 and later. The function doesn't exist on Plus <= 23.09.1 / CE <=2.7.2 so it's a no-op there.

dennypage: Okay.

What is the signature of validate_password()?

dennypage:
Requested changes pushed.

PR #1368, opened 2024-04-10
Return to the ga version of NUT (2.8.2)

dennypage:
Redmine issue [15393](https://redmine.pfsense.org/issues/15393).

- Move from the development version (nut-devel) back to the release version (nut>=2.8.2).
- Complete the update for PHP 8 and use config_get_path() for all config access.

Note that nut-2.8.2 is already present in pfSense / FreeBSD-ports.


dennypage:
@jim-p, when you have a moment, could you take a look at this please? Thank you kindly.

PR #1369, opened 2024-04-19
security/pfSense-pkg-suricata: Add section for change all actions of rules

julioliraup:
For environments with more than 1k rules, greater agility is required to change rule actions


bmeeks8:
I do not favor the approach taken in this code. The change is made to the master rules list instead of being made on a "per Suricata interface" basis. The way this pull request is written, the rule actions will change for all Suricata interfaces. That is likely not always desirable. Rule actions should be modified on a per-interface basis.

PR #1373, opened 2024-05-14
Fix for rpz zone located in several views at the same time

rantal:
This is a fix for bug with wrong generation of **response-policy** parameter in **View** sections of named.conf when for single rpz zone selected multiple views


marcos-ng:
Merged in 46eb4ba9f306081cbec19de06dcccc38c3030e39.

PR #1379, opened 2024-07-25
Redmine #14299: pfBlockerNG does not honor the cURL source interface setting for DNSBL lists.

chamilton-ccn:
This pull request addresses [Redmine #14299](https://redmine.pfsense.org/issues/14299) (previous pull request for this issue has been closed).


chamilton-ccn:
So this is still incomplete! I thought I had it, but in its current state, this OR only fixes the issue for IPv4/IPv6 lists. Another commit is forthcoming.

chamilton-ccn:
___Now___ we should be all set! Just had to add those last two lines. Sorry about that.

chamilton-ccn:
Thank you!

PR #1381, opened 2024-09-02
pfBlockerNG-devel_v3.2.0_15

BBcan177:
* Add 'txt/x-file' to allowed downloaded file types
* Add IPinfo ASN from BGPview.io
* Fix deprecation warning in Lighttpd
* Normalize IPv6 prefix to its start address to avoid PHP errors. Ref: https://redmine.pfsense.org/issues/14256
* Update GeoIP/ASN aliases on next update/cron vs Frequency settings defined
* Reduce config saves to ACB
* Fix issue with Pre/Post Script processing, where the original download feed would be overwritten. When using a Reload, it would fail to parse the plain txt file vs the AWS JSON file that it was requiring originally.
* Update MaxMind, TOP1M and ASN once per day at random hour
* Change Spamhaus Drop/eDrop feeds
* H3X discontinue feed


MichaelGer1410:
Hello, and thanks for yout great work!

I Have Update pfSense to 24.03 in hope i can update pfblockerng to the latest version 3.2.0_15 . But i get only the 3.2.0_10. Can you explain how i can Update to the latest?

Michael

BBcan177:
The latest changes are in pfBlockerNG-devel. So you need to uninstall
pfBlockerNG first to install devel. Ensure that Keep Settings is checked in
the General Tab. Then run a Force Reload all.

MichaelGer1410:
I have Installed 3.2.0_10 devel already. I have uninstall it now. but i just have the option to install the same version. 3.2.0_15 dont show...

Bloodpack:
same here no update to pfBlockerNG-devel_v3.2.0_15

MichaelGer1410:
When i choose Branch "Current Stable Version 24.03" i just have the option to install 3.2.0_10 When i choose Branch "development snapshot" i can just choose 3.2.0_13. 
Do i something wrong?

PR #1388, opened 2024-10-16
net/pfSense-pkg-pfBlockerNG-devel: bugfixes

marcos-ng:


BBcan177:
Approved

BBcan177 (approved):

BBcan177 (approved):

BBcan177 (approved):

BBcan177:
* Some ASN Description from IPinfo have accents and other special characters that need to be encoded before saving to the config.conf file

* Add "application/x-ndjason" file mime-type for the new Spamhaus json Feed

* Fix issue with IPv6 ASN conversions

* XMLRPC sync will now run on all Force Updates|Reload even if there are no other changes made

* For IPinfo ASN Database to download, it was previously required to set the ASN Cache setting. This is not required to be set now.

* Fixed a regression in the GeoIP pages causing changes not to be saved.

marcos-ng:
This has been merged into devel branches and 24.03/2.7.2.

PR #1389, opened 2024-11-05
security/pfSense-pkg-acme: change Porkbun to use api.porkbun.com instead of porkbun.com

WereCatf:
Changes the dynDNS provider 'Porkbun' to use the domain `api.porkbun.com` instead of `porkbun.com` as accessing the API through `porkbun.com` will cease to work December 1st. Porkbun informed customers by email of this change and now it is also reflected on their API documentation page at https://api.porkbun.com/api/json/v3/documentation

Signed-off-by: Nita Vesa <nita.vesa@outlook.com>


steve-bailey1:
Please dont forget to update the API URL in /etc/inc/dyndns.class lines 967, 1002, and 1008.

PR #1391, opened 2024-11-19
Add mDNS Bridge (mdns-bridge) package

dennypage:
Implements Redmine #15853 (https://redmine.pfsense.org/issues/15853)

mDNS Bridge is intended as an alternative to Avahi reflection for sharing Multicast DNS across network interfaces.

Compared with Avahi, mDNS Bridge is expected to offer several advantages:

- mDNS Bridge is purpose built for bridging mDNS data only.
- Does not retain state across packets (no caching).
- Correctly forwards mDNS NSEC packets.
- Supports newer mDNS packet types such as OPT, SVCB and HTTPS.
- Offers interface specific inbound/outbound filter controls.
- Smaller and simpler code, offering easier inspection/validation.
- Run time memory requirements approximately 1/2 the size of Avahi.


dennypage:
Hey all, can we get this out for testing with 25.03 please? Thanks

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
```diff
@@ -0,0 +1,517 @@
...
+$pconfig['global_filter_list'] = config_get_path($path_global_filter_list, '');
+$pconfig['disable_packet_filtering'] = config_get_path($path_disable_packet_filtering, false);
+$pconfig['interfaces'] = config_get_path($path_interfaces, []);
+
+// Avahi conflict
+$avahi_enabled = config_get_path('installedpackages/avahi/config/0/enable', false) &&
+		 config_get_path('installedpackages/avahi/config/0/reflection', false);
+
+$available_interfaces = get_configured_interface_with_descr();
+unset($available_interfaces['wan']);
```
marcos-ng, line 54: Don't assume `wan` is any particular kind of interface. If there is something in particular that disqualifies an interface from being available, check for that specifically. See here for an example:
https://github.com/pfsense/FreeBSD-ports/blob/4f80f22a9de3e01f744f34ccf8dd5edfb0790947/net/pfSense-pkg-udpbroadcastrelay/files/usr/local/pkg/udpbroadcastrelay/udpbroadcastrelay.inc#L321

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
```diff
@@ -0,0 +1,517 @@
...
+	if ($pconfig['enable'] && $avahi_enabled) {
+		$input_errors[] = gettext('Avahi relfection must be disabled before enabling mDNS Bridge');
+	}
+
+	// Validate interfaces
+	if (count($pconfig['active_interfaces']) < 2) {
+		$input_errors[] = gettext('A minimum of two interfaces are required');
+	}
+
+	// Vaidate and normalize the global filter
```
marcos-ng, line 70: Typo `Vaidate`

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
```diff
@@ -0,0 +1,517 @@
...
+	function hideGlobalFilter() {
+		hideClass('sh_global_filter_none', $('#global_filter_type').prop('value') != 'none');
+		hideClass('sh_global_filter_allow', $('#global_filter_type').prop('value') != 'allow');
+		hideClass('sh_global_filter_deny', $('#global_filter_type').prop('value') != 'deny');
+		hideClass('sh_global_filter_list', $('#global_filter_type').prop('value') == 'none');
+	}
+
+<?php
+	// Show/hide based on interface filters
+	foreach ($available_interfaces as $interface => $name) {
```
marcos-ng, line 451: I think a better approach here is to store `$available_interfaces` in a JS variable, loop through it, and call the relevant functions with the interface values as a function parameter.

marcos-ng (changes requested):
### Logging
It'd be best to have dedicated logging for the package. It's fairly straight-forward to do so. You'll need to add the appropriate configuration to info.xml:
https://github.com/pfsense/FreeBSD-ports/blob/4f80f22a9de3e01f744f34ccf8dd5edfb0790947/net/pfSense-pkg-udpbroadcastrelay/files/usr/local/share/pfSense-pkg-udpbroadcastrelay/info.xml#L10

And implement the logging like so:
https://github.com/pfsense/FreeBSD-ports/commit/179fe73ff8fbb18f68d94b6400d06cdac5d25072

### Config calls
Avoid fetching and setting the config multiple times by storing it in a variable first then using the appropriate `array_`* functions. For example:
```
$package_config = config_get_path('installedpackages/mdns-bridge', []);
$ip_protocols = array_get_path($package_config, 'global_ip_protocols', '');
$ip_protocols = array_get_path($package_config, "interfaces/{$interface}/ip_protocols", '');
```
This can help with performance.

marcos-ng:
For easier review, please submit any changes as additional commits rather than force-pushing the branch. Once the PR is ready, the commits can be squashed before merging.

net/pfSense-pkg-mDNS-Bridge/Makefile
```diff
@@ -0,0 +1,46 @@
+
+PORTNAME=	pfSense-pkg-mDNS-Bridge
+PORTVERSION=	1.0
+CATEGORIES=	net
+MASTER_SITES=	# empty
+DISTFILES=	# empty
+
+MAINTAINER=	coreteam@pfsense.org
```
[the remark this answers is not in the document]
dennypage: I'm okay with that.

net/pfSense-pkg-mDNS-Bridge/files/etc/inc/priv/mdns-bridge.priv.inc
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+global $priv_list;
+
+$priv_list['page-services-mdns-bridge'] = array();
+$priv_list['page-services-mdns-bridge']['name'] = "WebCfg - Services: mDNS Bridge";
+$priv_list['page-services-mdns-bridge']['descr'] = "Access the mDNS Bridge package gui";
```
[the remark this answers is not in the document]
dennypage: I'm okay changing gui -> GUI.

I understand that entire array could be created at once, but the element by element approach is used by priv.inc content from the base pfSense system (/etc/inc/priv.defs.inc), as well as other packages, so to me it seems best to stay within that custom.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/pkg/mdns-bridge.inc
```diff
@@ -0,0 +1,258 @@
...
+		unlink_if_exists(MDNS_BRIDGE_RC_FILE);
+		unlink_if_exists(MDNS_BRIDGE_CONF_FILE);
+		return;
+	}
+	
+	// Write the config and rc files
+	mdns_bridge_write_files();
+
+	// Take no further action during platform boot
+	if (platform_booting()) {
```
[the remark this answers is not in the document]
dennypage: Is this function new? I don't see it being used in other packages.
Does it exist in the CE version?

net/pfSense-pkg-mDNS-Bridge/files/usr/local/pkg/mdns-bridge.inc
```diff
@@ -0,0 +1,258 @@
...
+		switch ($status) {
+			case 'MASTER':
+				break;
+
+			case 'BACKUP':
+			case 'INIT':
+				// If the VHID is still initializing, or in backup, don't start the service.
+				log_error("mDNS Bridge service not started due to CARP status ({$status})");
+				return;
+				break;
```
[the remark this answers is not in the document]
dennypage: Yes, that's correct. It's good coding practice to always have a break.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/pkg/mdns-bridge.inc
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	// Start the service
+	log_error("Starting service mDNS Bridge");
+	start_service(MDNS_BRIDGE_SVC_NAME);
+}
+
+// Handle CARP notifications
+function mdns_bridge_plugin_carp($pluginparams) {
+	global $config;
```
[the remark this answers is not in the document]
dennypage: Agreed. Sorry about that.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
```diff
@@ -0,0 +1,517 @@
...
+$path_global_ip_protocols = $package_path . 'global_ip_protocols';
+$path_global_filter_type = $package_path . 'global_filter_type';
+$path_global_filter_list = $package_path . 'global_filter_list';
+$path_disable_packet_filtering = $package_path . 'disable_packet_filtering';
+$path_interfaces = $package_path . 'interfaces';
+
+// Get the current configuration
+$pconfig['enable'] = config_get_path($path_enable);
+$pconfig['carp_vhid'] = config_get_path($path_carp_vhid);
+$pconfig['active_interfaces'] = explode(',', config_get_path($path_active_interfaces, 'lan'));
```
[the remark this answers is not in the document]
dennypage: Okay.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
dennypage: The assumption that "wan" is a wan interface is common convention in a lot of the code. It helps users from making mistakes.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
dennypage: Okay.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
dennypage: Given that the current code works, and has been tested, I'd prefer to stick with the current code.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
marcos-ng: That assumption comes from the earlier days of the project. It's no longer true - though there are some places where the old code hasn't been fixed yet.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/pkg/mdns-bridge.inc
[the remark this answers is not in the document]
marcos-ng: No worries, no need to apologize :)

net/pfSense-pkg-mDNS-Bridge/files/usr/local/pkg/mdns-bridge.inc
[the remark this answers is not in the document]
marcos-ng: It's new-ish, and yes.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/pkg/mdns-bridge.inc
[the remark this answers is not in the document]
marcos-ng: Depends on the person I suppose. I disagree - it's like writing `break;` twice. Keep it if you'd prefer though.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/pkg/mdns-bridge.inc
```diff
@@ -0,0 +1,258 @@
...
+	}
+	$file_array[] = '';
+
+	// Write the config file
+	if (!file_put_contents(MDNS_BRIDGE_CONF_FILE, implode("\n", $file_array))) {
+		log_error('ERROR: Could not write file ' . MDNS_BRIDGE_CONF_FILE);
+	}
+
+	// Write the rc file
+	$start = MDNS_BRIDGE_CMD . ' -s -c ' . MDNS_BRIDGE_CONF_FILE . ' -p ' . MDNS_BRIDGE_PID_FILE;
```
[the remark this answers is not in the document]
marcos-ng: I mention it because it's part of the code style guidelines in the docs. I do agree it can have readability benefits though. Keep it if you prefer.

dennypage:
> For easier review, please submit any changes as additional commits rather than force-pushing the branch.

I started down this path, but it's difficult to keep track of where things are. I'm currently doing individual commits so I can keep track of everything, but I can resubmit as a single commit at the end if that is necessary.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/pkg/mdns-bridge.inc
```diff
@@ -0,0 +1,258 @@
...
+const MDNS_BRIDGE_SVC_NAME = 'mdns-bridge';
+const MDNS_BRIDGE_CMD = '/usr/local/bin/mdns-bridge';
+const MDNS_BRIDGE_RC_FILE = '/usr/local/etc/rc.d/mdns-bridge.sh';
+const MDNS_BRIDGE_CONF_FILE = '/usr/local/etc/mdns-bridge.conf';
+const MDNS_BRIDGE_PID_FILE = '/var/run/mdns-bridge.pid';
+
+
+// Is mDNS Bridge enabled?
+function mdns_bridge_enabled() {
+	return (config_get_path('installedpackages/mdns-bridge/enable', false));
```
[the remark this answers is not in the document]
dennypage: Looking at the php7 -> php8 guide, it seems that the recommendation was to avoid config_path_enabled() because there are usages that it does not interpret correctly. The current mDNS behavior is one of these.

Couple of additional things that bother me about this... it really should be a set method either way. Having 'on' be a set method, and 'off' be a delete method makes the code harder to read and less reliable. It also requires custom code in the www file to detect that the pconfig key is false, and then call delete for the config rather than simply setting the config to match pconfig. This does not seem desirable. Lastly, it would break the beta tester installs.

FWIW, I used the config_get_path pattern when I ported my other packages to php8, and it was accepted without issue.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/pkg/mdns-bridge.inc
```diff
@@ -0,0 +1,258 @@
...
+		$list[$vid] = sprintf("VHID %s on %s (%s)", $vip['vhid'], convert_friendly_interface_to_friendly_descr($vip['interface']), $vaddr);
+	}
+	return ($list);
+}
+
+// Write the config and rc files
+function mdns_bridge_write_files() {
+	// Get the real interface names
+	$interfaces = array();
+	foreach (explode(',', config_get_path('installedpackages/mdns-bridge/active_interfaces')) as $interface) {
```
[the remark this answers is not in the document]
dennypage: Good catch, thanks.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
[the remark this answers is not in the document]
dennypage: FWIW, there are a lot of places, both pfSense proper and packages, that assume a meaning for the internal names 'lan' and 'wan'. I'm okay deleting the default of 'lan' here because it doesn't hurt anything, but it does seem to go against a strong convention.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
dennypage: I understand that there may be other WAN interfaces, but there isn't a way to detect that. There are a LOT of places that assume 'wan' should be excluded from interface selections. Unless there is a function to call such as is_interface_wan(), I don't see any reasonable way to implement. And since 'wan' will always be a wan interface, I don't see any reason to not exclude it. This is a good practice to help the 80% user from making an egregious mistake.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
dennypage: If you would like to propose a diff, I'll have a look at it.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
marcos-ng: WAN interfaces are simply those that have a gateway directly assigned to them (which is implied for those using DHCP for address assignment). If needed, you may check for that specifically. The interface assigned to 'wan' will not always be a WAN-type interface hence why it's incorrect to assume that in code. And yeah, there are a lot of places that still do this - mostly because they haven't been updated in a decade - which tend to cause bugs and generally break configuration that would otherwise be expected to work.

dennypage:
Btw, I did not implement the logging suggestion. Since mdns-bridge in normal operation does not produce log entries beyond start/stop, it didn't seem appropriate to create a log file for its exclusive use.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
marcos-ng: Here's an example for the `hideAdvanced()` function that can be adopted to the rest:
```
var available_interfaces = <?=json_encode(array_keys($available_interfaces))?>;
function hideAdvanced() {
	for (let interface of available_interfaces) {
		showadvanced = (($('#global_filter_type').prop('value') == 'none')
			&& ($('#inbound_filter_type_' + interface).prop('value') == 'none')
			&& ($('#outbound_filter_type_' + interface).prop('value') == 'none')
		);
		hideClass('sh_disable_packet_filtering', !showadvanced);
	}
}
```

marcos-ng:
It's not a requirement but I'd like to see the logging implemented because it helps to keep the logs more organized and reduces clutter in the system log. It's easy to do and will make things generally easier to update/maintain (we are considering reworking how logging is done in general).

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
dennypage: To fix 'wan' across the board, we're going to need a function that performs the check. Having each individual package and internal functions perform that check just isn't going to be practical. I'll make the 'wan' change when it comes around. I've already changed 'lan' because it doesn't do any harm.

dennypage:
The array change you proposed was good. I like it for efficiency. It's done. Thanks.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/pkg/mdns-bridge.inc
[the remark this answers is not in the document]
dennypage: Yes, mdns-bridge, like my other packages, explicitly sets the value to 'yes'.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
dennypage: I'm not a JS guy, so it'll take me a bit to think that through.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/pkg/mdns-bridge.inc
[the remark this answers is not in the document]
marcos-ng: Right. That value comes from the element's `value` which the form class sets to `yes` by default (which differs from the browser-standard 'on'). As part of the ongoing API work, we are considering reworking the `enable` code. Using the proper function will make that easier, though you may keep it as is since for now it's not strictly required.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
marcos-ng: There are two functions used for this now (that is, to check if it's a "WAN-type" interface). You can do the following:
```
$interface = 'wan';
$interface_is_wan = false;
if (interface_has_gateway($interface) || interface_has_gatewayv6($interface)) {
	$interface_is_wan = true;
}
```

net/pfSense-pkg-mDNS-Bridge/files/usr/local/pkg/mdns-bridge.inc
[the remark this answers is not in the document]
dennypage: Thanks

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
dennypage: Okay, I can work with that.

dennypage:
> It's not a requirement but I'd like to see the logging implemented because it helps keep the logs more organized and reduces clutter in the system log. It's easy to do and will make things generally easier to update/maintain (we are considering reworking how logging is done in general).

While it might cut down on the system log itself, overall it makes the logging less useful and will likely make things harder for package maintainers. In supporting the NUT package, I frequently ask people for entries in the system log that contain 'usb' or 'ups', and having to send people to search multiple log files and then stitch together the result will be painful and error prone. There is a similar situation with package restart.

Unless a subsystem or package is generating a lot of entries of its own, I don't think the logs should be separated.

net/pfSense-pkg-mDNS-Bridge/files/usr/local/www/mdns-bridge.php
dennypage: The proposed code doesn't quite match what I'm doing. The proposed code produces this:
```
	// Show/hide the advanced
	var available_interfaces = ["lan","opt2","opt3","opt4"];
	function hideAdvanced() {
		for (let interface of available_interfaces) {
			showadvanced = (($('#global_filter_type').prop('value') == 'none')
				&& ($('#inbound_filter_type_' + interface).prop('value') == 'none')
				&& ($('#outbound_filter_type_' + interface).prop('value') == 'none')
			);
			hideClass('sh_disable_packet_filtering', !showadvanced);
		}
	}

```

Which makes the decision based on the global filter and the last interface.

The existing code produces this:
```
	function hideAdvanced() {
		showadvanced = ($('#global_filter_type').prop('value') == 'none')
		 && $('#inbound_filter_type_lan').prop('value') == 'none'
 		 && $('#outbound_filter_type_lan').prop('value') == 'none'
		 && $('#inbound_filter_type_opt2').prop('value') == 'none'
		 && $('#outbound_filter_type_opt2').prop('value') == 'none'
		 && $('#inbound_filter_type_opt3').prop('value') == 'none'
		 && $('#outbound_filter_type_opt3').prop('value') == 'none'
		 && $('#inbound_filter_type_opt4').prop('value') == 'none'
		 && $('#outbound_filter_type_opt4').prop('value') == 'none'
	;

```
Which makes the decision on all of the filters combined.

I think what I would need is something like this:
```
        // Show/hide advanced
        var available_interfaces = <?=json_encode(array_keys($available_interfaces))?>;
        function hideAdvanced() {
                showadvanced = $('#global_filter_type').prop('value') == 'none';
                for (let interface of available_interfaces) {
                        if ($('#inbound_filter_type_' + interface).prop('value') != 'none' ||
                            $('#outbound_filter_type_' + interface).prop('value') != 'none') {
                                showadvanced = false;
                        }
                }
                hideClass('sh_disable_packet_filtering', !showadvanced);
        }

```

Can you look that over and tell me what you think? Thanks.

dennypage: Also, I'd appreciate any examples for the other cases where I'm conditionally defining functions such as this:
```
        foreach ($available_interfaces as $interface => $name) {
                echo "function hideInterfaceInboundFilter_{$interface}() {";
                        echo "hideClass('sh_interface_inbound_filter_none_{$interface}', $('#inbound_filter_type_{$interface}').prop('value') != 'none
');";
                        echo "hideClass('sh_interface_inbound_filter_allow_{$interface}', $('#inbound_filter_type_{$interface}').prop('value') != 'all
ow');";
                        echo "hideClass('sh_interface_inbound_filter_deny_{$interface}', $('#inbound_filter_type_{$interface}').prop('value') != 'deny
');";
                        echo "hideClass('sh_interface_inbound_filter_list_{$interface}', $('#inbound_filter_type_{$interface}').prop('value') == 'none
');";
                echo "}\n";
        }

```

and defining change functions such as this:

```
        foreach ($available_interfaces as $interface => $name) {
                echo "$('#inbound_filter_type_{$interface}').change(function() {";
                        echo "hideInterfaceInboundFilter_{$interface}();";
                        echo "hideAdvanced();";
                echo "});\n";
        }

```

I'm not a JS guy, so how to extrapolate the hideAdvanced function isn't obvious to me. Thanks.

marcos-ng: I can try adding a commit to the branch and you can revise/test as needed.

dennypage: Thank you Marcos! I'll have a look.

dennypage: I fixed a minor typo, and changed it so it only looks at selected interfaces for displaying the advanced section. Please review. Thanks.

marcos-ng: LGTM

dennypage:
> I think it's good. Squash the commits and include the redmine ID in the message - e.g. `net/pfSense-pkg-mDNS-Bridge: Add mDNS Bridge package. Implement #15853`.

Done.

> The packages will also need to be added here when this is merged: https://github.com/pfsense/pfsense/blob/master/tools/conf/pfPorts/poudriere_bulk

Do you want me to do a PR for this?

dennypage:
Btw, please don't forget about the related PR:
https://github.com/pfsense/FreeBSD-ports/pull/1392

marcos-ng:
This is now in dev. After it's been tested working there, it can be picked for 2.8/25.03.

PR #1395, opened 2024-12-06
Add CautiousConnect IPv6 feed from sixint.io

robertbeverly:
Add CautiousConnect IPv6 blocker feed:
  [https://sixint.io/products/cc_docs/about.html](url)


marcos-ng:
Added with 82c4be995de1b188deb28d77a059c2e8bf24e6f3.

PR #1397, opened 2025-01-01
Update pfblockerng_category_edit.php

krishenriksen:
Instead of breaking and having the user start all over, just replace non-word characters with underscores


BBcan177:
I tested this and it should already hold incorrect user inputs and post the Error header asking for changes to be made before allowing a Save. Please confirm or send more specifics to debug

PR #1398, opened 2025-01-07
Fix ntopng listen options

dennypage:
Implements Redmine #15976 ([https://redmine.pfsense.org/issues/15976](https://redmine.pfsense.org/issues/15976))

If both IPv4 and IPv6 server listen are enabled, the IPv4 address will not work. While the current version of ntopng accepts a list for listen address, it only uses the last element in the list. This means that both an IPv4 and IPv6 listen address cannot be provided at the same time. Hopefully this will be fixed upstream at some point in the future. Until then, we limit the listen selection to one protocol.

Additionally, bracket the address for IPv6.

### DIFF
--- a/net/pfSense-pkg-ntopng/Makefile
+++ b/net/pfSense-pkg-ntopng/Makefile
@@ -1,8 +1,7 @@
 # $FreeBSD$
 
 PORTNAME=	pfSense-pkg-ntopng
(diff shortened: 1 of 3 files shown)


dennypage:
Hey all, can we get this in 25.03 please? Thanks

PR #1406, opened 2025-01-30
Remove Talos Snort

ikifar2012:
The Talos Snort IP blocklist has been moved to a different link which now requires you accept terms of service before downloading and redirects to what seems like a unique URL making it impossible to add this as a default source as of right now the current link just fails to download repeatedly


ikifar2012:
@BBcan177 I edited the status to say discontinued instead of removing the block

PR #1407, opened 2025-02-04
net/pfSense-pkg-pfBlockerNG-devel: fix dual-stack null blocking. Fix #16094

fenichelar:
See the [2nd](https://github.com/pfsense/FreeBSD-ports/pull/1407#issuecomment-2634796462) and [3rd](https://github.com/pfsense/FreeBSD-ports/pull/1407#issuecomment-2634819227) comments of this PR and https://forum.netgate.com/topic/196164/null-blocking-servfail/10.
for more details.

### DIFF
--- a/net/pfSense-pkg-pfBlockerNG-devel/files/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/net/pfSense-pkg-pfBlockerNG-devel/files/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -193,8 +193,6 @@ def init_standard(id, env):
     pfb['rr_types'] = (RR_TYPE_A, RR_TYPE_AAAA, RR_TYPE_ANY, RR_TYPE_CNAME, RR_TYPE_DNAME, RR_TYPE_SIG, \
                        RR_TYPE_MX, RR_TYPE_NS, RR_TYPE_PTR, RR_TYPE_SRV, RR_TYPE_TXT, 64, 65)
 
-    pfb['rr_types2'] = ('A', 'AAAA')
-
     # List of HSTS preload TLDs
     pfb['hsts_tlds'] = ('android', 'app', 'bank', 'chrome', 'dev', 'foo', 'gle', 'gmail', 'google', 'hangout', \
                         'insurance', 'meet', 'new', 'page', 'play', 'search', 'youtube')
@@ -257,10 +255,6 @@ def init_standard(id, env):
                 pfb['dnsbl_ipv6'] = '::' + pfb['dnsbl_ipv4']
             else:
                 pfb['dnsbl_ipv6'] = '::'
-
-            # DNSBL IP/Log types (0 = Null Blocking logging, 1 = DNSBL Web Server logging, 2 = Null Blocking no logging)
-            pfb['dnsbl_ip'] = {'A': {'0': '0.0.0.0', '1': pfb['dnsbl_ipv4'], '2': '0.0.0.0'},
-                               'AAAA': {'0': '::', '1': pfb['dnsbl_ipv6'], '2': '::'} }
 
             # List of DNS R_CODES
             rcodeDB = {0: 'NoError', 1: 'FormErr', 2: 'ServFail', 3: 'NXDOMAIN', 4: 'NotImp', 5: 'Refused', 6: 'YXDomain',
@@ -764,7 +758,7 @@ def get_details_dnsbl(m_type, qinfo, qstate, rep, kwargs):
     if isDNSBL is not None:
 
         # If logging is disabled, do not log blocked DNSBL events (Utilize DNSBL Webserver) except for Python nullblock events
-        if pfb['python_nolog'] and not isDNSBL['b_ip'] in ('0.0.0.0', '::'):
+        if pfb['python_nolog'] and not isDNSBL['null']:
             return True
 
         # Increment dnsblgroup counter
@@ -1392,11 +1386,11 @@ def operate(id, event, qstate, qdata):
             # Determine if domain has been previously validated
             if q_name not in excludeDB:
 
-                q_type_str = qstate.qinfo.qtype_str
                 isFound = False
                 log_type = False
                 isInWhitelist = False
                 isInHsts = False
+                nullBlocking = True
                 b_type = 'Python'
                 p_type = 'Python'
                 feed = 'Unknown'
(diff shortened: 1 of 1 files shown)


fenichelar:
The key for the `dnsblDB` dictionary is just `q_name`, meaning the same dictionary entry is used for all query types. The response answer is currently saved in the dictionary. If the first query is for AAAA, the response answer saved in the dictionary is an IPv6 address; otherwise it is an IPv4 address. Subsequent queries then always get the same response answer even if the answer is invalid for the query (e.g., the answer to an A query may be "::", or the answer to an MX query may be "0.0.0.0"). These invalid answers result in SERVFAIL errors because `msg.set_return_msg(qstate)` equals `False` when an invalid answer is appended.

This PR makes the following changes to fix this issue:

1. Instead of saving the answer in the dictionary, just a boolean indicating null blocking is saved. The appropriate answer is then calculated for each query depending on the query type and null blocking boolean.
2. For queries other than A, AAAA, or ANY; no answer is included because neither an IPv4 nor IPv6 address is a valid answer for most other queries. Note that there could be other ways of doing this but this is the best way I can come up with. It may make sense to add an SOA response to the authority section to increase the negative answer caching time on the client. I have this mocked up and can add to this PR if desired.
3. For ANY queries, both A and AAAA responses are returned. Note that this change is not required to fix the issue but intended to simplify the code and better support IPv6.
4. The TTL of A and AAAA answers has been increased from 60 to 3600. Note that this change is not required to fix the issue but intended to reduce DNS load.

fenichelar:
It is worth noting the `b_type` and `p_type`, both of which are saved in the dictionary, no longer contain the query type. Because the dictionary entries are reused for all query types, it doesn't really make sense to save the query type in the dictionary as it just indicates the first query type received for that `q_name`; which I think is misleading.

fenichelar:
At least with null blocking and no logging, this has been working well in practice.

mattcheau:
I was not able to recreate the `SERVFAIL` status/blank address answer for blocked domain A queries. All blocked domain A record queries always returned `NOERROR` status/`0.0.0.0` in the response.

I was, however, able to recreate the `SERVFAIL` status/blank answer for blocked domain AAAA queries. All blocked domain AAAA record queries returned `SERVFAIL` status/blank address in the answer.

After applying this patch, both A and AAAA queries for blocked domains now return `NOERROR` status/`0.0.0.0` or `::` respectively.

DNSBL Groups are running in a combination of "Null Block (logging)" and "Null Block (no logging)".

CE 2.7.2/pfBlockerNG-devel 3.2.0_20

mattcheau:
Deleted my previous comment. Even with this patch I continue to observe inconsistent results with both A and AAAA queries, and for blocked domains listed in DNSBL Groups utilizing both "Null Block (logging)" and "Null Block (no logging)".

Flushing `unbound` cache is the only thing that temporarily restores `NOERROR` status/`0.0.0.0` or `::` answers for blocked domains.

CE 2.7.2/pfBlockerNG-devel 3.2.0_20

fenichelar:
@mattcheau

> I've identified some consistent behavior: A query for `$blocked_domain` returns `NOERROR`/0.0.0.0 and is presumably cached, and now the AAAA query for `$blocked_domain` will return `SERVFAIL`/blank address.
> 
> Flush cache.
> 
> AAAA query for `$blocked_domain` now properly returns `NOERROR`/`::` and is presumably cached, but now the A query for `$blocked_domain` will return `SERVFAIL`/blank address.
> 
> So basically a cached A causes the issue to occur with an AAAA lookup. And vice versa, a cached AAAA causes the issue to occur with an A lookup.

Yes, that is the primary bug that this PR fixes. See the [2nd](https://github.com/pfsense/FreeBSD-ports/pull/1407#issuecomment-2634796462) and [3rd](https://github.com/pfsense/FreeBSD-ports/pull/1407#issuecomment-2634819227) comments of this PR for more details.

>Deleted my previous comment. Even with this patch I continue to observe inconsistent results with both A and AAAA queries, and for blocked domains listed in DNSBL Groups utilizing both "Null Block (logging)" and "Null Block (no logging)".

It sounds like the patch didn't actually get applied. You need to apply the patch to the running file, `/var/unbound/pfb_unbound.py` and then restart unbound.

mattcheau:
> It sounds like the patch didn't actually get applied. You need to apply the patch to the running file, `/var/unbound/pfb_unbound.py` and then restart unbound.

Apologies for the confusion. You're correct. I had "patched" `/usr/local/pkg/pfblockerng/pfb_unbound.py` rather than `/var/unbound/pfb_unbound.py`. Not seeing the 3600 TTL should've called my attention to this.

But okay—with the patch now properly applied, *now* I see everyting working as expected.

Tested with both a blocked domain from a "Null Block (logging)" group, and a blocked domain from a "Null Block (no logging)" group.

A and AAAA lookups for both blocked domains now consistently return `NOERROR`/`0.0.0.0` or `::` respectively (with a TTL of 3600 😊).

mattcheau:
> It should also fix other DNS query types, like MX. Instead of returning a SERVFAIL, they should get a NOERRROR with no data.

I see that. Other RR types (HTTPS, MX, NS, SRV, etc.) returning `NOERROR`/no data.

Can't believe the number of SERVFAILs I'd been observing in the `dns_reply.log` all this time without recognizing it as an issue.

mattcheau:
Observing new behavior. Near-simultaneous A and AAAA lookups (like iOS and Windows apparently sometimes do automatically) for the same blocked domain now appear as separate entries in the `dnsbl.log`.

Confirmed via packet capture that they are, in fact, separate A and AAAA (and not */ANY) queries. First-in-time contains `+` at the end of the log line, and the second-in-time contains `-` at the end of the log line.

Because of the effect this has on DNSBL statistics, it makes me wonder if some functionality which _strictly for logging purposes_ ignores an AAAA query for a blocked domain if A is cached and vice versa  was lost.

fenichelar:
> Observing new behavior. Near-simultaneous A and AAAA lookups (like iOS and Windows apparently sometimes do automatically) for the same blocked domain now appear as separate entries in the `dnsbl.log`.
> 
> Confirmed via packet capture that they are, in fact, separate A and AAAA (and not */ANY) queries. First-in-time contains `+` at the end of the log line, and the second-in-time contains `-` at the end of the log line.
> 
> Because of the effect this has on DNSBL statistics, it makes me wonder if some functionality which _strictly for logging purposes_ ignores an AAAA query for a blocked domain if A is cached and vice versa was lost.

Yes, this is the expected behavior. The "-" means it is a duplicate query. The query type isn't considered when determining duplicates because the dictionary, which controls logging, isn't dependent on query type. See https://github.com/pfsense/FreeBSD-ports/pull/1407#issuecomment-2634819227.

marcos-ng:
Please open a redmine for this issue here:
https://redmine.pfsense.org/projects/pfsense-packages/issues?set_filter=1

Make these commits to the `net/pfSense-pkg-pfBlockerNG-devel` port instead (the non-devel package will then be synced with devel). Also squash the commits and reference the redmine in the message, e.g. `net/pfSense-pkg-pfBlockerNG-devel: fix dual-stack null blocking. Fix #00000` where `00000` is the redmine URL number.

fenichelar:
@marcos-ng done

PR #1415, opened 2025-05-15
Update mDNS-Bridge to 2.0

dennypage:
Implements Redmine #16190 (https://redmine.pfsense.org/issues/16190)

This update adds filtering of link local addresses from forwarded mDNS records.


dennypage:
Closing in favor of #1422

PR #1416, opened 2025-05-28
Bump Makefile, remove use-ipv6 block from pfblockerng.inc

nintendo424:
This should take care of the following issue: https://www.reddit.com/r/pfBlockerNG/comments/1jb5rtc/ipv6_woes_wrong_vip/
@BBcan177


BBcan177:
Can you remove the Makefile bump please

PR #1423, opened 2025-08-31
Update mDNS-Bridge to 2.2

dennypage:
Implements Redmine [16399](https://redmine.pfsense.org/issues/16399)

Update mdns-bridge to 2.2.
Add an option to enable mDNS decode warings that are silent by default.


marcos-ng:
Looks fine, though preferably the commit message would match others, e.g.
> net/pfSense-pkg-mDNS-Bridge: update to 2.2. Implement #16399.

Before this is merged, the devel branch will first need to be synced with upstream.

PR #1439, opened 2026-05-07
net/pfSense-pkg-pfBlockerNG[]: Improve granularity of domain count based on physmem

FrenchToastDev:
Instead of considering increases of the max domain count at 1000 MiB boundary between 4000 and 8000, use a 500 MiB boundary. This is particularly beneficiary (and common) for machines equipped with 8 GiB but for which `hw.physmem` is slightly under 8000 MiB (since some memory is typically reserved outside what is typically physically installed for UEFI, etc).

Without this, it has now become impossible to use a handful of the largest UT1 block list (+ a few classic malware lists) without running against the 6M limit with recent increases in these block lists. So, this commit should provide some headroom.

This was tested with a 2.8.1 pfSense CE version equipped with 8GiB (but with physmem at 8364314624, that is about 7977 MiB, 23 MiB shy of the 8000 MiB). This allows to go from a max count of 2M * 3 = 6 M to a max count of 2.25M * 3 = 6.75M, giving it much more headroom in regards to ever bigger block lists.


FrenchToastDev:
The motivation of making these increments as opposed to have a linear interpolation:
- The code is quite simple to review/test.
- Should OOM be seen with unbound with the new values, it would be clear what to change. Instead, if a linear interpolation was used, we would potentially have hundreds/thousands of potential steps for domain_max_cnt making it difficult to maintain. Here, if we observe an OOM, we know which value to modify (instead of having to wonder if the slope, or the origin of the line had to be changed).

However, I am open to use a linear interpolation (between each existing steps) should you prefer (though I would assume that I would choose a round down to 100K domains to help the proliferation of domain_max_cnt values and help diagnose things easier in the future). (I have originally implemented this before choosing the simpler approach proposed in this commit).

Thanks for your consideration.

PR #1443, opened 2026-05-27
Add standalone VPN portal with LDAP login and OTP provisioning

alexgit2k:
## Summary

This adds a standalone VPN portal to the OpenVPN Client Export package at https://firewall.example.com/vpn.php

The portal lets LDAP-authenticated users set up VPN access without requiring a pfSense administrator login. After successful LDAP authentication and VPN group validation, it verifies or provisions a local FreeRADIUS user with OTP settings, shows a 2FA QR code, and exposes OpenVPN configuration/software downloads.

## Changes

- Add `vpn.php` as the user-facing VPN portal page.
- Add `vpn_portal.inc` with portal helper logic for:
  - LDAP authentication and group checks
  - FreeRADIUS package validation and local user provisioning
  - HA backup-node warning
  - OTP secret handling
  - OpenVPN Client Export integration
  - secure download handling
- Register the new files in `pkg-plist`.
- Keep FreeRADIUS provisioning limited to the local package config and generated users file.
- Block unsupported SQL-backed FreeRADIUS authorization with a clear error.
- Use stored OpenVPN Client Export settings for generated configs/installers.
- Add configurable download buttons and user-friendly profile filenames.

## Configuration

The portal configuration is kept at the top of `vpn.php`:

- `otp_qr_display_name`: Display name used as the TOTP issuer in authenticator apps.
- `ldap_authserver_name`: Optional pfSense LDAP authentication server name; empty uses the first LDAP server.
- `radius_authserver_name`: Optional pfSense RADIUS authentication server name; empty uses the first RADIUS server.
- `primary_ha_hostname_or_ip`: Optional primary HA firewall hostname, FQDN, or IP address for backup-node warnings.
- `vpn_group_name`: LDAP group required for VPN access; empty allows every authenticated LDAP user.
- `lowercase_usernames`: Normalizes login names to lowercase before LDAP, FreeRADIUS, session, and OpenVPN handling.
- `freeradius_user_defaults`: Default FreeRADIUS fields for automatically created OTP users.
- `download_buttons`: Enables or disables individual configuration, installer, legacy, MSI, and Viscosity download buttons.

## Notes

The portal does not update SQL/MySQL/MariaDB FreeRADIUS backends. It only updates `installedpackages/freeradius/config` and regenerates the FreeRADIUS users file via `freeradius_users_resync()`.


jim-p:
We have considered this approach in the past but ultimately decided deliberately not to include the functionality because it is an unacceptable reduction in security. This reduces multiple factors of authentication to a single factor (user login), and if someone's credentials are compromised the attacker gains much more access than they would have otherwise. Additionally, using the firewall for such a multi-user web service adds significant risk on its own.

This is the sort of service that should be run on a completely separate server with additional protections, not integrated into the firewall.

PR #1446, opened 2026-05-29
net/pfSense-pkg-pfBlockerNG[]: fetch sources from andrebrait/pfBlockerNG via USE_GITHUB

andrebrait:
## Summary

Move both pfBlockerNG ports from in-tree source files to `USE_GITHUB`,
pointing at https://github.com/andrebrait/pfBlockerNG.

| Port | Channel | Branch | Tag |
|------|---------|--------|-----|
| `net/pfSense-pkg-pfBlockerNG` | stable | `main` | `v3.2.15` |
| `net/pfSense-pkg-pfBlockerNG-devel` | devel | `devel` | `v3.2.16` |

**What changes:**
- Add `USE_GITHUB`, `GH_ACCOUNT`, `GH_PROJECT`, `GH_TAGNAME`, `DIST_SUBDIR` to both Makefiles
- Set `WRKSRC` to match the upstream repository layout (see below)
- Remove `MASTER_SITES`/`DISTFILES`/`EXTRACT_ONLY`/`do-extract:` (framework handles fetching and extraction)
- Remove all source files from `files/` — only `pkg-install.in` and `pkg-deinstall.in` remain
- `do-install:` now references `${WRKSRC}` instead of `${FILESDIR}`
- `do-test:` in the devel port runs pytest from `${WRKSRC:H}` (see below)

**Repository layout and `WRKSRC`:**

The upstream repository separates production code from dev tooling:
`src/` mirrors the pfSense filesystem (`etc/`, `usr/local/`), while
`tests/`, `scripts/`, and `stubs/` remain at the repo root.

`WRKSRC` is therefore set to `${WRKDIR}/${GH_PROJECT}-${PORTVERSION}/src`
so that `${WRKSRC}${PREFIX}/...` resolves correctly for all
`INSTALL_DATA`/`INSTALL_SCRIPT` lines.

The `do-test:` target (devel port only) runs from `${WRKSRC:H}` — the
repository root one level above `src/` — where `pyproject.toml` and
the `tests/` directory live.

**Going forward:** version bumps require only a one-line `GH_TAGNAME` change, automated by the release workflow in `andrebrait/pfBlockerNG`.

## Test plan

- [ ] `make fetch` succeeds for both ports (downloads the GitHub archive)
- [ ] `make extract` extracts to the expected `WRKSRC` path
- [ ] `make install` stages all files correctly
- [ ] `make test` (devel) runs the pytest suite and passes


andrebrait:
Still pending approval from @BBcan177, of course, and the repo will likely be moved to him as the owner.

andrebrait:
I had a chat with him and I realize the way you handle pfBlockerNG is a bit different than some other packages out there, so this will likely not go forward. 

I think I'll try something more manual then, but that should still allow us to have proper CI and automated reviews.

Gotta think about it.

PR #1447, opened 2026-06-02
Create ca.conf with EAB credentials when issuing ACME certificate (Issue #16862)

TGX03:
When attempting to issue a certificate using Actalis, which requires EAB, the issuance always fails with "external account binding required". This is because pfSense currently does not create a ca.conf file when putting the account.key file in the temporary acme.sh-directory.

This patch also creates the ca.conf-file, containing the e-mail-address and the EAB-Key-ID and EAB-HMAC. After usage, there are also an ACCOUNT_URL and CA_KEY_HASH present in the file, but it seems they are not strictly necessary.

This is my attempt to fix ticket #16862.


jim-p:
Merged with some changes as a part of a larger ACME pkg v1.3 commit

PR #1452, opened 2026-07-13
Multi-Method-env-overwrite Refactor envvariables initialization in acme.inc

KonradLanz:
Initialize envvariables inside the loop to isolate variables for multiple DNS provider accounts, ensuring safe accumulation of tokens across multi-domain certificates.

https://redmine.pfsense.org/issues/16946 should also solve https://redmine.pfsense.org/issues/11614


jim-p:
Merged with some changes as a part of a larger ACME pkg v1.3 commit

PR #1453, opened 2026-07-24
Update mDNS-Bridge to 3.0

dennypage:
Implements Redmine #16970

This update adds support for the following new features of mdns-bridge 3.0.0:

* Peer specific filter lists, which allows the use of individual outbound filters based on the source interface of the mDNS packet.
* Efficient deny all filter list, which allows a single pull down selection to create a "deny all" filter, rather having to use a fake allow list.


marcos-ng:
Merged.

PR #1454, opened 2026-07-30
Add network p2mp nbma

babakkazemi:
add Network Type "Point-to-Multipoint Non-Broadcast" to frr_ospf_interfaces.xml
see https://redmine.pfsense.org/issues/16978 and https://forum.netgate.com/topic/167500/pfsense-ui-not-updating-frr-conf/10


babakkazemi:
add Network Type "Point-to-Multipoint Non-Broadcast" to frr_ospf_interfaces.xml
add 'point-to-multipoint non-broadcast' to frr_validation.inc
see https://redmine.pfsense.org/issues/16978 and https://forum.netgate.com/topic/167500/pfsense-ui-not-updating-frr-conf/10

